PR #7, opened 2019-08-22
[REVIEW] Code review for initial check-in 

zhangjianting:
Changes since last code review meeting on 08/07
1. Naming. Spelling out names until they are  too long (exceptions are well-known abbreviations,  e.g., PIP for point-in-polygon and ST for spatiotemporal)
2. Using uint32_t intsead of user-defined uint. Location changed to Location3D, Coord changed to Coord2D, Time changed to TimeStamp. They are all defined under cuSpatial namespace to avoid confusion. 
3. Comments are now in header files only. Use "see X.hpp" in C++/CUDA implementation file for consistency.  Using cuSpatial namesapce for all public APIs. 
4. Restructuring.  Head files that are specific to test code are moved to test directory to reduce # of public APIs. 
5. Separating Python test code: those using hard-coded simple data (called toy samples) remain under "cuspatial/tests" directory while those using real data are put under "cuspatial/demos" directory. 
6. Adding C++ unit test for Hausdorff distance. All APIs have python test/demo code. Quite a few are compared with independent Python packages (serial code on CPUs) 
7. Bug fixes. 

Changes on 08/23
1. namespace:  changed from cuSpatial to cuspatial
2. data structures: Location3D changed to location_3d and Coord2D  changed to coord_2d; spelling out longitude/latitude/altitude in location_3d
3. filenames: coord_trans.hpp to coordinate_transform.hpp and coord_to_traj.cu to coordinate_to_trajectory.cu
4. I was able to make lonlat_to_coord function returns std::pair<gdf_colum,gdf_colum> in coordinate_transform.cu.  However, I do not know how to expose the function (with std::pair) to python API. I will ask over Slack. 
5. Follow all the changes on README.md.
6. Add license to all files. 

Changes on 08/27
1. use nullptr instead of NULL. 
2. location_3d and coord_2d are now templated to support both float and double 
3. remove all commented stream parameters in APIs (may be added back later when additional streams are allowed) 
4. spell out and document the member variables in struct polygons (previously PolyMeta). Referring to OGC SFS for definitions of a single polygon. Additional documentation will be provided in design docs.  
5. change from _ll to _lonlat and change fn to filename for clarity; use templated location_3d/coord_2d
6. change coor_x,coor_y to just x and y.  
7. change TimeStamp to its_timestamp
8. bug fix to allow using templated functions for both float and double points/vertices. 
9. remove or comment out code for outputting to std::cout. Using CUDF_EXPECTS for error message whereas possible. 
Two exceptions: 1) some simple output message serving as heartbeat single to make sure code runs as expected. 2) first n (typically 10) input/output column values for some cuSpatial operations. They seem to be useful when running the python wrapper APIs.  

A few issues might be able to be addressed later (e.g., next release) 
1) using Table, instead of individual columns (with different or same lengths), for APIs. 
2) using try...catch... instead of CUDF_EXPECTS. 
In addition, dealing with indentation and other formatting issues need more time. 
       
Hope this can be a better base for further code review.


cpp/include/cuspatial/cuspatial.h
```diff
@@ -0,0 +1,31 @@
...
+		double alt;
+	};
+
+	struct Coord2D
+	{
+		double x;
+		double y;
+	};
+
+	struct TimeStamp
```
harrism: Can you just use cuDF timestamps? This is a very nonstandard representation of a timestamp -- typically it's an offset from the Epoch.  Do you need to define this in cuSpatial?

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @Brief compute Hausdorff distances among all pairs of a set of trajectories
+ * https://en.wikipedia.org/wiki/Hausdorff_distance
+
+ * @param[in] coord_x: x coordinates of the input trajectroies
+ * @param[in] coord_y: y coordinates of the input trajectroies
+ * @param[in] traj_cnt: numbers of vertices of the set of trajectories;
+ * also used to compute the starting offsets of the trjajectories in coord_x/coord_y arrays
+
+ * @returns a flatted (1D) vector of all-pair direted Hausdorff distances among trajectories (i,j)
```
harrism:
```suggestion
 * @returns a flattened (1D) column of all-pairs directed Hausdorff distances among trajectories (i,j)
```

How are the outputs ordered in terms of (i, j)? Document this.

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * Note that ausdorff distance is not symmetrical
```
harrism:
```suggestion
 * @note Hausdorff distance is not symmetrical
```

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+gdf_column directed_hausdorff_distance(const gdf_column& coord_x,const gdf_column& coord_y,const gdf_column& traj_cnt
```
harrism:
```suggestion
gdf_column directed_hausdorff_distance(const gdf_column& x, const gdf_column& y, const gdf_column& trajectory_count);
```

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,23 @@
...
+    		/* ,cudaStream_t stream = 0   */);
```
harrism:
```suggestion
```

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+}  // namespace cuSpatial
```
harrism:
```suggestion
}  // namespace cuspatial
```

README.md
```diff
@@ -1,76 +1,88 @@
-# RAPIDS Standard Repo Template
+# cuSpatial
+GPU-Accelerated Spatial and Trajectory Data Management and Analytics Library <br>
+**NOTE:** cuSpatial depends on [cudf](https://github.com/rapidsai/cudf) and [rmm](https://github.com/rapidsai/rmm) under [RAPDIS](https://rapids.ai/) framework<br> 
+See [here](https://nvidia-my.sharepoint.com/:p:/r/personal/jiantingz_nvidia_com/Documents/GPU4STA_V5.pptx?d=wa5b5d6d397074ea9a1600e74fd8a6345&csf=1&e=h7MdRq)
```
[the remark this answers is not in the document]
zhangjianting: This is a placeholder. I am working on it.

cpp/include/cuspatial/coord_trans.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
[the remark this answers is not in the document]
zhangjianting: Will do.

cpp/include/cuspatial/coord_trans.hpp
```diff
@@ -0,0 +1,21 @@
...
+ * @Brief transforming in_lon/in_lat (lon/lat defined in Coord2D) to out_x/out_y relative to a camera origiin
+
+ * @param[in] cam_lon: logitude of camera origin
+ * @param[in] cam_lat: latitude of camera origin
+ * @param[in] in_lon: array of longitude coordinates of input points to be transformed
+ * @param[in] in_lat: array of latitude coordinates of input points to be transformed
+ * @param[out] out_x: array of x coordiantes after transformation (in kilometers -km)
+ * @param[out] out_y: array of y coordiantes after transformation (in kilometers -km)
+ */
+void lonlat_to_coord(const gdf_scalar  & cam_lon,const gdf_scalar  & cam_lat,const gdf_column  & in_lon,const gdf_column  & in_lat,
```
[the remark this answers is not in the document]
zhangjianting: I will try std::pair for this version. I need to figure out how to access the pair in Python wrapper. If not successful, using Column Group (as we talked over slack) might be a solution.

cpp/include/cuspatial/cuspatial.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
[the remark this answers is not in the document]
zhangjianting: I guess the License header needs to be added to all files. Will do.

cpp/include/cuspatial/cuspatial.h
zhangjianting: I looked into cuDF timestamp definition and APIs - they are very complex.  Day of year and day of week etc., which are popular in OLAP type of operations, need to be computed in real time which may take quite some overhead. I also need to compute duration between two timestamps in CUDA kernel and it is unclear to me how to call cuDF APIs for this purpose. We can talk more about this and make changes in the next release.

harrism:
@zhangjianting here's an example of using a `std::pair` from Cython: https://github.com/rapidsai/cudf/blob/21378308a97462fe19159ffb0d1f35231bcc54a7/python/cudf/cudf/_lib/includes/unaryops.pxd

You need `from libcpp.pair cimport pair` and your function definition would be something like

```
cdef pair[gdf_column, gdf_column] foo(
        const gdf_column& input, # etc.
    ) except +
```

Note return gdf_columns by value, not by pointer.

cpp/include/cuspatial/traj_thrust.h
```diff
@@ -0,0 +1,92 @@
...
+	/**
+	 *@brief Thrust functor for generating an 1D bounding box from two 1D box
+	 */
+	struct TBBox_reduction : public thrust::binary_function<TBBox,TBBox,TBBox>
+	{
+		__host__ __device__
+			TBBox operator()(TBBox a, TBBox b)
+			{
+				// lower left corner
+				TimeStamp minT=(a.first<b.first)?a.first:b.first;
```
harrism: Inconsistent indentation.

cpp/include/cuspatial/traj_thrust.h
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	typedef  thrust::pair<TimeStamp, TimeStamp> TBBox;
+
+	/**
+	 *@brief Thrust functor for lifting TimeStamp to an interval (1D box)
+	 */
+	struct TBBox_transformation : public thrust::unary_function<TBBox,TBBox>
+	{
+		__host__ __device__
```
harrism: Do you *ever* need to call this from the host? If not, make it device-only.

```suggestion
		__device__
```

cpp/include/cuspatial/traj_thrust.h
```diff
@@ -0,0 +1,92 @@
...
+#include <ostream>
+#include <cuspatial/cuspatial.h>
+
+namespace cuspatial
+{
+
+	/**
+	 *@brief Thrust functor for comparing two TimeStamp variables; used in sorting based on timestamp
+	 */
+	__host__ __device__
```
harrism: Do you ever call this from the host? If not, make it device-only.

```suggestion
	__device__
```

cpp/include/cuspatial/traj_thrust.h
```diff
@@ -0,0 +1,92 @@
...
+				return TBBox(time, time);
+			}
+	};
+
+	/**
+	 *@brief Thrust functor for generating an 1D bounding box from two 1D box
+	 */
+	struct TBBox_reduction : public thrust::binary_function<TBBox,TBBox,TBBox>
+	{
+		__host__ __device__
```
harrism: Ever called from the host?

cpp/include/cuspatial/traj_thrust.h
```diff
@@ -0,0 +1,92 @@
...
+		coord_2d operator()(location_3d pt)
+		{
+		  coord_2d c;
+		  c.x = ((origin.longitude - pt.longitude) * 40000.0 *
+			   cos((origin.latitude + pt.latitude) * M_PI / 360) / 360);
+		  c.y = (origin.latitude - pt.latitude) * 40000.0 / 360;
+		  return c;
+		}
+	};
+} // namespace cuspatial
```
harrism:
```suggestion
} // namespace cuspatial

```

cpp/include/cuspatial/traj_thrust.h
```diff
@@ -0,0 +1,92 @@
...
+	{
+		location_3d origin;
+		__host__ __device__
+		coord_transformation(location_3d _origin): origin(_origin){}
+
+		__host__ __device__
+		coord_2d operator()(location_3d pt)
+		{
+		  coord_2d c;
+		  c.x = ((origin.longitude - pt.longitude) * 40000.0 *
```
harrism: Avoid magic numbers (40000.0, 360). Store these values in `constexpr` variables with clearly understandable names.

Also, if you are planning to do double precision arithmetic, use `360.0`, not `360`. And if you intend single-precision math, then use the `f` suffix on the literals.

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+#pragma once
+
+#include <cudf/cudf.h>
+
+namespace cuspatial {
+
+/**
+ * @Brief deriving trajectories from points (x/y relative to an origin), timestamps and objectids
```
harrism:
```suggestion
 * @Brief Derives trajectories from points (x/y relative to an origin), timestamps and object IDs
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * by first sorting based on id and timestamp and then group by id.
```
harrism: Not part of the @brief.
```suggestion
 *
 * Sorts by object id and timestamp and then groups by id.
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * @param[in/out] coor_y:  y coordiantes reative to a camera origin (before/after sorting)
+ * @param[in/out] oid: object (e.g., vehicle) id column (before/after sorting); upon completion, unqiue ids become trajectory ids
+ * @param[in/out] ts: timestamp column (before/after sorting)
+ * @param[out] tid: trajectory id column (see comments on oid)
+ * @param[out] len: #of points in the derived trajectories
+ * @param[out] pos: position offsets of trajectories used to index coor_x/coor_y/oid/ts after completion
+
+ * @returns the number of derived trajectory
+ */
+int coord_to_traj(gdf_column& coor_x,gdf_column& coor_y,gdf_column& oid, gdf_column& ts,
```
harrism: I believe that the three output columns have equal length. Correct? If so, return a table that contains the three output columns. Much cleaner.

```suggestion
cudf::table coord_to_traj(const gdf_column& coords_x, const gdf_column& coords_y,
                          const gdf_column& object_ids, const gdf_column& timestamps);
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ 			      gdf_column& tid, gdf_column& len,gdf_column& pos/* ,cudaStream_t stream = 0   */);
+
+
+/**
+ * @Brief computing distance(length) and speed of trajectories after their formation (e.g., from coord_to_traj)
```
harrism:
```suggestion
 * @brief Compute the distance and speed of trajectories
 * 
 * Trajectories are typically derived from coordinate data using coords_to_trajectories).
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * @param[in] len: number of points column ordered by (id,timestamp)
+ * @param[in] pos: position offsets of trajectories used to index coor_x/coor_y/oid/ts ordered by (id,timestamp)
+ * @param[out] dist: computed distances/lengths of trajectories in meters (m)
+ * @param[out] speed: computed speed of trajectories in meters per second (m/s)
+
+ * Note: we might output duration (in addtiion to distance/speed), should there is a need
+ * duration can be easiy computed on CPUs by fetching begining/ending timestamps of a trajectory in the timestamp array
+ */
+
+void traj_distspeed(const gdf_column& coor_x,const gdf_column& coor_y,const gdf_column& ts,
```
harrism:
```suggestion
cudf::table trajectory_distance_and_speed(const gdf_column& coords_x, const gdf_column& coords_y,
                                          const gdf_column& timestamps, const gdf_column& trajectory_lengths,
                                          const gdf_column& trajectory_coordinate_offsets);
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ 			    const gdf_column& len,const gdf_column& pos,gdf_column& dist,gdf_column& speed
+ 			    /* ,cudaStream_t stream = 0   */);
+
+
+
+/**
+ * @Brief computing spatial bounding boxes of trjectories
```
harrism:
```suggestion
 * @brief Compute spatial bounding boxes of trajectories
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * @param[in] coor_y: y coordiantes reative to a camera origin and ordered by (id,timestamp)
+ * @param[in] len: number of points column ordered by (id,timestamp)
+ * @param[in] pos: position offsets of trajectories used to index coor_x/coor_y/ ordered by (id,timestamp)
+ * @param[out] bbox_x1/bbox_y1/bbox_x2/bbox_y2: computed spaital bounding boxes in four columns
+
+ * Note: temporal 1D bounding box can be computed similary but it seems that there is no such a need;
+ * Similar to the dicussion in coord_to_traj, the temporal 1D bounding box can be retrieved directly
+ */
+
+void traj_sbbox(const gdf_column& coor_x,const gdf_column& coor_y,
```
harrism:
```suggestion
cudf::table trajectory_spatial_bounds(const gdf_column& coords_x, const gdf_column& coords_y,
                                      const gdf_column& trajectory_lengths, const gdf_column& trajectory_coordinate_offsets);
```

cpp/include/cuspatial/cuspatial.h
```diff
@@ -0,0 +1,31 @@
...
+		double altitude;
+	};
+
+	struct coord_2d
+	{
+		double x;
+		double y;
+	};
+
+	struct TimeStamp
```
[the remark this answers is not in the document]
harrism: its is an unfortunate acronym. Makes it look like "It's a timestamp".  What's wrong with `timestamp`? It's in the cuspatial namespace so it can't be confused with a `cudf::timestamp`, they are different types and namespaces.

cpp/include/cuspatial/shared_util.h
```diff
@@ -0,0 +1,228 @@
...
+		std::cout<<"# of features="<<ply.num_f<<std::endl;
+		std::cout<<"# of rings="<< ply.num_r<<std::endl;
+		std::cout<<"# of vertices="<< ply.num_v<<std::endl;
+		size_t len=(4+ply.num_g+ply.num_f+ply.num_r)*sizeof(int)+2*ply.num_v*sizeof(double);
+		CUDF_EXPECTS(len==sz,"expecting file size and read size are the same");
+
+		ply.p_g_len=new uint32_t[ ply.num_g];
+		ply.p_f_len=new uint32_t[ ply.num_f];
+		ply.p_r_len=new uint32_t[ ply.num_r];
+		CUDF_EXPECTS(ply.p_g_len!=NULL&&ply.p_f_len!=NULL&&ply.p_r_len!=NULL,"expecting p_{g,f,r}_len are non-zeron");
```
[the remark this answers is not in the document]
harrism: At least use `nullptr`, not `NULL`.

cpp/include/cuspatial/traj_thrust.h
harrism: But it can hurt. See my section on __host__ __device__ in the C++ code standard I shared with you. Add it if and when we need it. Not "just in case".

cpp/include/cuspatial/soa_readers.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace cuspatial {
+
+	/**
+	*@Brief read uint32_t (unsigned integer with 32 bit fixed length) data from file as column;
+    *mostly for identifiers, lengths and positions.
+    *@param[in] col_fn: file name of uint32_t column data
+    *@param[out] ids: gdf_column storing the uint32_t column
+	*/
+	void read_uint_soa(const char *col_fn, gdf_column& ids);
```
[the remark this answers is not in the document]
harrism: Usually in APIs that are processing a lot of data, parameter passing is not a bottleneck. :)

- Pass by value whenever it is inexpensive to do so. 
 - Use const reference parameters when passing an object that is too expensive to pass by value (primitive types such as your int example and simple structs are NOT expensive to pass by value). 
 - Prefer returning results rather than using "in/out" parameters.
 - If you must use an in/out parameter, pass it by a pointer rather than a non-const reference.  This is a matter of style, but it's a style we prefer in cuDF/RAPIDS because we believe it makes it more clear that this parameter's contents will be modified. A non-const reference doesn't make that clear.

cpp/tests/spatial/pip_compare.cu
```diff
@@ -0,0 +1,222 @@
...
+
+#include <tests/utilities/cudf_test_utils.cuh>
+#include <tests/utilities/cudf_test_fixtures.h>
+
+using namespace cuspatial;
+
+template <typename T>
+struct PIPCompare : public GdfTest 
+{
+    T * h_point_x=nullptr, *h_point_y=nullptr;
```
harrism: Should use cudf::column_wrapper and std::vector / device_vector rather than raw pointers. Initialization will be much simpler and safer.

cpp/tests/spatial/hausdorff_toy.cu
```diff
@@ -0,0 +1,57 @@
...
+#include <cuspatial/soa_readers.hpp> 
+#include <cuspatial/hausdorff.hpp> 
+
+#include <tests/utilities/column_wrapper.cuh>
+#include <tests/utilities/cudf_test_utils.cuh>
+#include <tests/utilities/cudf_test_fixtures.h>
+
+using namespace cuspatial;
+
+struct HausdorffToy : public GdfTest
```
harrism: Why is it called a Toy? Please don't call them toys. Users need to know that cuSpatial is for real use.

cpp/tests/spatial/hausdorff_compare.cu
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright (c) 2019, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
harrism: What's the point of renaming it from `hausdorff_test` to `hausdorff_compare`? I think `hausdorff_test was a perfectly good name.

cpp/tests/spatial/pip_test.cu
```diff
@@ -101,6 +101,24 @@ TYPED_TEST(PIPTest, piptest)
     gettimeofday(&t2, nullptr);
     float gpu_pip_time2=cuspatial::calc_time("GPU PIP time 2(including point data transfer and kernel time)......",t1,t2);
 
+    int err_cnt=0,non_zero=0;
+    for(int i=0;i<this->point_len;i++)
```
harrism: Should be using google mock tools for comparisons. See the changes I made to pip test last week.

cpp/tests/spatial/hausdorff_test.cu
```diff
@@ -32,12 +32,12 @@ using namespace cuspatial;
 
 struct is_true
```
harrism: Rename. This is an equality comparison operator.

cpp/tests/spatial/pip_compare.cu
zhangjianting: h_point_x and h_point_y are pointers that are used in cpu code, which does/can not use cudf.  They are also used to populate the data field of point_x and point_y columns, which shows an alternative approach to create gdf_columns than column_wrapper. Column_wrapper is used in pip_toy, which tests on hard coded toy data.

cpp/tests/spatial/hausdorff_toy.cu
zhangjianting: It means that that the test uses simple toy data just for demonstration purposes - some packages use the convention. Please feel free to change.

cpp/tests/spatial/hausdorff_compare.cu
zhangjianting: *_test.cu is now for test cuSpatial only, using real data.  Some many execute twice to test consistency. 
*_compare.cu now compares CPU and GPU results using the same real data. 
These are two separate tests. Some tests may not have compare code.

cpp/tests/spatial/pip_compare.cu
harrism: Tests should not show alternative approaches. They should be succinct and easily maintainable. column_wrapper makes writing tests easier and less bug prone. Writing good tests typically is much more difficult and takes longer than writing the functionality itself, which is why testing utilities are important.

In general, raw pointers should be avoided when containers can be used because raw pointers often lead to memory leks.

cpp/tests/spatial/hausdorff_compare.cu
harrism: Note that multiple tests of the same module can go in the same source file. See libcudf for suggested test organization.

cpp/tests/spatial/pip_test.cu
zhangjianting: I could not locate your code. I vaguely remember gtest does have something specifically for it but  can not recall the details.

cpp/tests/spatial/hausdorff_test.cu
zhangjianting: It might be better to use inline lambda function as the functor is so simple. There are other similar situations. Need to figure out this. Any idea?

cpp/tests/spatial/pip_compare.cu
zhangjianting: I agree. This is also the reason that I separated CPU code (used in _compare.cu only) and GPU code (used in both _test and _compare code). 
The CPU code does need access to raw pointers. I could pass gdf_columns to CPU code, but then the data need needs to be downloaded to CPU again, which is a unnecessary detour, in addition making the CPU code dependent on cudf which is unnecessary again.

cpp/tests/spatial/hausdorff_compare.cu
zhangjianting: Please feel free to merge. I am still far from a gtest expert :-)
On the other hand, having _toy,_test,_compare separate might enable build time re-configuration, which could be a desirable feature than using file level marcos.

PR #26, opened 2019-09-05
[REVIEW] Trajectory pytests. Remove toy examples from tests.

thomcom:
Confusion was already being caused by the leftovers in the `tests` directory so I removed the leftover `toy` files.

First pass of trajectory tests - `trajectory_distance_and_speed` doesn't support `cudf::timestamp` yet, we need to work on that @harrism .


python/cuspatial/cuspatial/bindings/trajectory.pyx
```diff
@@ -1,68 +1,86 @@
...
-                                                         c_len[0],c_pos[0])
+                                                         c_length[0],c_pos[0])
 
     dist_data, dist_mask = gdf_column_to_column_mem(&c_distance_speed.first)
     speed_data, speed_mask = gdf_column_to_column_mem(&c_distance_speed.second)
     dist=Column.from_mem_views(dist_data, dist_mask)
     speed=Column.from_mem_views(speed_data, speed_mask)
 
-    return dist,speed
+    return Series(dist), Series(speed)
```
harrism: Why not make this a DataFrame?

thomcom: No disagreement! I'm working in that direction as I get better bringing the two APIs together.

PR #27, opened 2019-09-05
Update soa readers and the trajectory locust demo.

thomcom:
The soa readers now return Series.

The Trajectory/soa/locust demo has been updated to work with real data.

### DIFF
--- a/python/cuspatial/cuspatial/demos/traj_test_soa_locust.py
+++ b/python/cuspatial/cuspatial/demos/traj_test_soa_locust.py
@@ -10,7 +10,7 @@ import cuspatial.bindings.trajectory as traj
 import cuspatial.bindings.soa_readers as readers
 import cuspatial.utils.traj_utils as tools
 
-data_dir = "/home/jianting/cuspatial/data/"
+data_dir = "./"
 df = pd.read_csv(data_dir + "its_camera_2.csv")
 this_cam = df.loc[df["cameraIdString"] == "HWY_20_AND_LOCUST"]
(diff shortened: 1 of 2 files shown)


thomcom:
This simple update makes one of the data demos work, and has soa readers returning the proper data type.
I haven't written unit tests for the soa readers yet as that will require careful binary data construction inside of the tests. This demo file with the demo data from https://www.dropbox.com/s/fu5jof3q929tmhx/demodata.tar.gz?raw=1 is the python level test for soa readers for now.

python/cuspatial/cuspatial/demos/traj_test_soa_locust.py
```diff
@@ -30,17 +30,11 @@
 y, m, d, hh, mm, ss, wd, yd, ms, pid = tools.get_ts_struct(ts_0)
 
 pnt_x, pnt_y = gis.cpp_lonlat2coord(cam_lon, cam_lat, pnt_lon, pnt_lat)
-num_traj, tid, len, pos = traj.cpp_coord2traj(pnt_x, pnt_y, id, ts)
-
+num_traj, trajectories = traj.cpp_derive_trajectories(pnt_x, pnt_y, id, ts)
```
harrism: Why is this demo using the bindings directly rather than the Python API?  e.g. it should work like cuDF. Maybe I misunderstood your previous PRs, but there should be a `cuspatial.derive_trajectories` so users don't have to call into `cuspatial.bindings`.

PR #32, opened 2019-09-06
Python API

thomcom:
Here's that python API you asked for. Restructure cuspatial to better match cudf. Start renaming/removing `cpp_` dirs.


python/cuspatial/cuspatial/core/gis.py
```diff
@@ -0,0 +1,91 @@
+# COPYRIGHT 2019 NVIDIA
```
harrism: Probably need a full license block here?

thomcom: 👍

harrism:
Let's not merge PRs without reviews.

PR #39, opened 2019-09-06
[REVIEW] Add cython headers to install, python / cmake packaging cleanup

kkraus14:
Packaging cleanup

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,19 @@
+[tool.black]
+line-length = 79
+target-version = ["py36"]
+include = '\.py?$'
+exclude = '''
+/(
+    thirdparty |
+    \.eggs |
+    \.git |
+    \.hg |
+    \.mypy_cache |
+    \.tox |
+    \.venv |
+    _build |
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+    \.hg |
+    \.mypy_cache |
+    \.tox |
+    \.venv |
+    _build |
+    buck-out |
+    build |
+    dist
+)/
+'''
```
thomcom: what is a `.toml` file?

kkraus14: PEP 518: https://www.python.org/dev/peps/pep-0518/ basically future Python build system things, but it's a crapshoot as to what supports it and what doesn't. Currently just using it for `black` configuration.

kkraus14:
Merging this.

@harrism if you still have issues after adding the `cudftestutil.a` to your `LD_LIBRARY_PATH` or if you use `CUDF_ROOT` and we're not adding a sufficient hint, we can fix it in a followup.

PR #58, opened 2019-09-12
[REVIEW] Improve documentation and return a hausdorff matrix.

### LINKED ISSUE #40: [BUG] `cuspatial.directed_hausdorff_distance` does not return a DataFrame
**Describe the bug**
This function returns a Series, but should return an nxn Dataframe, where n is the `len(sizes)` of the third argument. The orientation and method for building the `DataFrame` from `Series` should be carefully built, as it can be inefficient.

thomcom:
Write an example that explains what Hausdorff distance is. Correct the return value of this function. Fixes #40


python/cuspatial/cuspatial/tests/test_hausdorff_distance.py
```diff
@@ -91,20 +89,17 @@ def test_values():
...
-            4.12310563,
-            4.0,
-            3.60555128,
-            0.0,
-            1.41421356,
-            4.47213595,
-            1.41421356,
-            0.0,
-        ]
+    print(distance)
```
kkraus14: remove print statement from test?

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -25,10 +25,41 @@ def directed_hausdorff_distance(x, y, count):
...
+    Consider a pair of lines on a grid.
+
+     |
+     o
+    -oxx-
+     |
+    o = [[0, 0], [0, 1]]
+    x = [[1, 0], [2, 0]]
+
+    o[0] is the nearer point in o to x. The distance from o[0] to the farthest
```
harrism:
```suggestion
    o[0] is the closest point in o to x. The distance from o[0] to the farthest
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -25,10 +25,41 @@ def directed_hausdorff_distance(x, y, count):
...
+    point in x = 2.
+    x[0] is the nearer point in x to o. The distance from x[0] to the farthest
```
harrism:
```suggestion
    x[0] is the closest point in x to o. The distance from x[0] to the farthest
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -25,10 +25,41 @@ def directed_hausdorff_distance(x, y, count):
     ----------
```
harrism: I can't comment on it, but I don't think the top definition of what this function does is correct (or at least it's inconsistent with the C++ function docs). 

Should be "Compute the directed Hausdorff distances between all pairs of input trajectories" (not "any groupings of polygons").

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -25,10 +25,41 @@ def directed_hausdorff_distance(x, y, count):
...
+            cudf.Series([2, 2,]),
+        )
+        print(result)
+             0         1
+        0  0.0  1.414214
+        1  2.0  0.000000
+
+    Returns
+    -------
+    DataFrame: The pairwise Hausdorff distance of each set to each other set.
```
harrism: More precise docs:

```suggestion
    A DataFrame representing a pairwise directed distance matrix with one row and one column
    per input trajectory; the value at row i, column j represents the directed Hausdorff distance
    from trajectory i to trajectory j.
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -25,10 +25,41 @@ def directed_hausdorff_distance(x, y, count):
     ----------
     {params}
 
-    returns
-    DataFrame: 'min', 'max' columns of Hausdorff distances for each polygon
+    Example
+    -------
+    Consider a pair of lines on a grid.
```
harrism: Add a link to a definition of Hausdorff distance here. Suggestion:
```suggestion
    
           The directed Hausdorff distance from one trajectory to another is the greatest of all
           the distances from a point in the first trajectory to the closest point in the second. See
           [Wikipedia entry](https://www.google.com/search?client=safari&rls=en&q=Hausdorff+distance)
           
            Consider a pair of lines on a grid.
```

harrism: Forgive indentation.

python/cuspatial/cuspatial/core/gis.py
thomcom: Hausdorff doesn't explicitly compute over "trajectories", but I'm a fan of consistency! :thumbsup:

thomcom:
Ready to go again pending CI

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -14,21 +14,59 @@
 
 def directed_hausdorff_distance(x, y, count):
     """ Compute the directed Hausdorff distances between any groupings
```
harrism:
```suggestion
    """ Compute the directed Hausdorff distances between all pairs
```

python/cuspatial/cuspatial/core/gis.py
harrism: Maybe not in general, but the input is in the form of what cuSpatial calls trajectories, is it not? I guess in general they are "coordinate lists", but I think the expectation is that they will be created using `derive_trajectories`?

PR #59, opened 2019-09-12
[REVIEW] Improve detail of point in polygon docs.

### LINKED ISSUE #51: [DOC] Point-in-polygon documentation doesn't specify the need to repeat polygon start coord as end coord
**Describe the bug**
A close enough review of the point-in-polygon example code could reveal this, but it is not at all obvious. Polygons specified as input to this function need to be closed - their starting vertex must be the same as the ending vertex.

Clarify this in the docs.

thomcom:
It is important to specify that polygons in this function need to be defined in a closed form: the first and last coordinate must always be the same. Specified in the docs now.
Fixes #51


thomcom:
rerun tests

PR #68, opened 2019-09-26
Use YYMMDD tag in nightly build of cuspatial python package

mluukkainen:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - PR #61 Point-in-polygon DataFrame output
 - PR #59 Improve detail of point in polygon docs
 - PR #64 Use YYMMDD tag in nightly build
+- PR #68 Use YYMMDD tag in nightly build of cuspatial python 
 
 ## Bug Fixes
 
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -26,6 +26,11 @@ export GIT_DESCRIBE_TAG=`git describe --abbrev=0 --tags`
 export GIT_DESCRIBE_NUMBER=`git rev-list ${GIT_DESCRIBE_TAG}..HEAD --count`
 export MINOR_VERSION=`echo $GIT_DESCRIBE_TAG | grep -o -E '([0-9]+\.[0-9]+)'`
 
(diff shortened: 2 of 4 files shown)


ci/cpu/cuspatial/build_cuspatial.sh
```diff
@@ -1,5 +1,10 @@
 set -e
 
+# If nightly build, append current YYMMDD to version
```
dillon-cullinan: This block should be in `cpu/build.sh`. I'm assuming it exists in the libcuspatial one as well, and that should be removed. We try to minimalize the `build_<package>.sh`.

mluukkainen: I see following error when running ci/cpu/build.sh
 ci/cpu/build.sh: line 59: ci/cpu/libcuspatial/build_libcuspatial.sh: No such file or directory

I believe this is because build_libcuspatial.sh is not executable and doesn't have #!/bin/bash denotation. build_libcuspatial.sh expects to be run as bash build_libcuspatial.sh. To me it looks like ci/cpu/build.sh is either broken or I'm missing a piece on how to run it correctly.

I would need help understanding what to do here. Cannot proceed without help.

mluukkainen: I think I understand. The script expect WORKSPACE variable to be set. If it's not set the script execution fails. I'll move the logic to top level script and verify by setting WORKSPACE variable

mluukkainen: Fixed the patch. Variable is set now in ci/cpu/build.sh. If someone calls directly build_libcuspatial.sh they would not get the desired result. Build has to be done via toplevel ci/cpu/build.sh

mluukkainen: Moved to ci/cpu/build.sh. Works for me locally

mluukkainen:
> Still needs to be removed here: https://github.com/rapidsai/cuspatial/blob/branch-0.10/ci/cpu/libcuspatial/build_libcuspatial.sh
> 
> Once that is done this can be merged.

Fixed

Issue #69, opened 2019-09-27
ValueError while calling point_in_polygon


Nanthini10:
**Describe the bug**
Checking to see where certain points lie in 2 polygons, results in the following error:
```ValueError: Length of new column names: 2 does not match length of previous column names: {l_old_cols}```

**Steps/Code to reproduce bug**
```
import dask_cudf
df = dask_cudf.read_csv("cuspatial_error_sample_data.csv").set_index('Unnamed: 0')
df = df.compute()
df[['LATITUDE', 'LONGITUDE']].head()
```
```
	LATITUDE	LONGITUDE
Unnamed: 0		
0	70.9333333	-8.6666667
1	70.9333333	-8.6666667
2	70.9333333	-8.6666667
3	70.9333333	-8.6666667
4	70.9333333	-8.6666667
```
```
import cuspatial

poly_x  = cudf.Series([-10.0, 5, 5, -10, -10, 0, 10, 10, 0, 0])
poly_y =  cudf.Series([-10.0, -10, 5, 5, -10, 0, 0, 10, 10, 0])

query_ids = cudf.Series()
query_ids = query_ids.append(1)
poly_ids = query_ids.append(2)

query_offsets = cudf.Series()
query_offsets = query_offsets.append(5)
poly_offsets = query_offsets.append(10)

query_x = cudf.Series([0, -8.4, 6.0, 7.])
query_y = cudf.Series([0, -8, 6.0, 100.])
bitmaps = cuspatial.point_in_polygon_bitmap(
     df['LONGITUDE'], df['LATITUDE']
    poly_ids, poly_offsets,
    poly_x, poly_y
)
```

Error StackTrace:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-25-bc1d531045e2> in <module>
     17     df['LATITUDE'], df['LONGITUDE'],
     18     poly_ids, poly_offsets,
---> 19     poly_x, poly_y
     20 )

/conda/envs/rapids/lib/python3.6/site-packages/cuspatial-0.10.0a0+142.gcbe0513-py3.6-linux-x86_64.egg/cuspatial/core/gis.py in point_in_polygon_bitmap(x_points, y_points, polygon_ids, polygon_end_indices, polygons_x, polygons_y)
    141     )._apply_support_method("astype", dtype="bool")
    142     result_bools.columns = [
--> 143         f"in_polygon_{x}" for x in list(reversed(polygon_ids))
    144     ]
[...]

PR #70, opened 2019-09-27
[REVIEW] Require width parameterization of bitmap to binary conversion

beckernick:
Summary of Changes
- Requires and makes explicit the need for a width parameter in the bitmap to binary conversion utility. This is necessary for downstream usage in `point_in_polygon_bitmap`, ensuring that the dataframe will have the same number of columns as the number of polygon IDs.
- Updates tests accordingly


beckernick:
cc @thomcom

beckernick:
This closes #69

Issue #85, opened 2019-10-11
[BUG] Build fails against cudf branch-0.11


randerzander:
```
export CUDF_HOME=/rapids/cudf
export CUSPATIAL_HOME=/rapids/cuspatial
mkdir -p /rapids/cuspatial/cpp/build
cd /rapids/cuspatial/cpp/build
cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX
make install
...
...
...

-- Generating done
-- Build files have been written to: /rapids/cuspatial/cpp/build
Scanning dependencies of target cuspatial
[  6%] Building CUDA object CMakeFiles/cuspatial.dir/src/io/soa/polygon_soa_reader.cu.o
[ 13%] Building CUDA object CMakeFiles/cuspatial.dir/src/io/soa/point_soa_reader.cu.o
[ 20%] Building CUDA object CMakeFiles/cuspatial.dir/src/io/soa/uint32_soa_reader.cu.o
[ 26%] Building CUDA object CMakeFiles/cuspatial.dir/src/io/soa/timestamp_soa_reader.cu.o
[ 33%] Building CUDA object CMakeFiles/cuspatial.dir/src/query/spatial_window_points.cu.o
[ 40%] Building CUDA object CMakeFiles/cuspatial.dir/src/spatial/point_in_polygon.cu.o
[ 46%] Building CUDA object CMakeFiles/cuspatial.dir/src/spatial/haversine.cu.o
[ 53%] Building CUDA object CMakeFiles/cuspatial.dir/src/spatial/hausdorff.cu.o
[ 60%] Building CUDA object CMakeFiles/cuspatial.dir/src/spatial/lonlat_to_coordinate.cu.o
[ 66%] Building CUDA object CMakeFiles/cuspatial.dir/src/trajectory/trajectory_spatial_bound.cu.o
[ 73%] Building CUDA object CMakeFiles/cuspatial.dir/src/trajectory/trajectory_distance_speed.cu.o
[ 80%] Building CUDA object CMakeFiles/cuspatial.dir/src/trajectory/derive_trajectories.cu.o
[ 86%] Building CUDA object CMakeFiles/cuspatial.dir/src/trajectory/subset_trajectories.cu.o
/rapids/cuspatial/cpp/src/trajectory/subset_trajectories.cu:18:10: fatal error: cudf/copying.hpp: No such file or directory
 #include <cudf/copying.hpp>
          ^~~~~~~~~~~~~~~~~~
[...]

PR #89, opened 2019-10-17
Fix cudf header include

vishalmehta1991:
Fixes a header, which breaks compilation


vishalmehta1991:
Addresses issue #85

thomcom:
Thanks!

harrism:
@vishalmehta1991 why did you close it?

thomcom:
I think we're now all set in https://github.com/rapidsai/cuspatial/pull/90, thanks for your line!

PR #94, opened 2019-10-25
[REVIEW] Fix legacy migrations so cuspatial builds again

thomcom:
Modify headers with various `legacy` paths.


thomcom:
rerun test

PR #113, opened 2020-01-10
Refactor supporting libcudf++

thomcom:
This PR adds support for changes made upstream in cudf with the libcudf++ merge.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -42,7 +42,7 @@ logger "Check GPU usage..."
 nvidia-smi
 
 logger "Activate conda env..."
-conda create --name cuspatial_dev --file conda/environments/cuspatial_dev.yml
+conda env create --name cuspatial_dev --file conda/environments/cuspatial_dev.yml
 conda activate cuspatial_dev
 # conda install "cudf=${MINOR_VERSION}.*" "cudatoolkit=$CUDA_REL" "gdal=2.4.*"
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -42,9 +42,9 @@ logger "Check GPU usage..."
 nvidia-smi
 
 logger "Activate conda env..."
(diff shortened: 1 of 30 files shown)


mike-wendt:
Looks like the GPU tests are failing as they are pulling cudf 0.11 and should be pulling 0.12

mike-wendt:
I think we can leave it the other way, just this branch is not seeing the right tag. The latest tag it sees is 0.11 hence getting the old version.

There are benefits to leaving it being tag driven, but I know there is also the need to unblock the nightlies... plus it's the weekend. Given all that if this works merge, I'll approve and we'll clean up this on Monday

ci/gpu/build.sh
```diff
@@ -42,8 +42,10 @@ logger "Check GPU usage..."
 nvidia-smi
 
 logger "Activate conda env..."
+# conda env create --name cuspatial_dev --file conda/environments/cuspatial_dev.yml
+# conda activate cuspatial_dev
 source activate gdf
-conda install "cudf=${MINOR_VERSION}.*" "cudatoolkit=$CUDA_REL" "gdal=2.4.*"
+conda install -c rapidsai-nightly -c default "cudf=0.12.*" "cudatoolkit=$CUDA_REL" "gdal=2.4.*"
```
kkraus14: These channels should already be set in the container, should also continue to be able to use the minor version variable for cudf

conda/environments/cuspatial_dev.yml
```diff
@@ -0,0 +1,10 @@
+name: cuspatial_dev
+channels:
+  - rapidsai
+  - nvidia
+  - rapidsai-nightly
+  - conda-forge
```
kkraus14: Add defaults below conda-forge

mike-wendt:
Just fixed the tagging issue, will re-test and this should pass so we can merge

mike-wendt:
rerun tests

PR #123, opened 2020-02-11
[WIP] Fixing error_utils.hpp

randerzander:
libcuspatial is no longer building after recent changed to libcudf, specifically moving `utilities/error_utils.hpp` to `cudf/utilities/error.hpp`.

After these changes, I'm still failing to build with missing `cub/util_type.cuh`, but that could be an issue with my build environment.


thomcom:
You should now need a `git submodule sync --recursive && git submodule update --init --recursive` with cuspatial.

PR #128, opened 2020-02-21
Use RMM's `DeviceBuffer` for Python allocations

jakirkham:
Instead of using RMM's `device_array`, which is planned to go away ( https://github.com/rapidsai/rmm/issues/180 ), switch to using RMM's `DeviceBuffer`. Since this already supports `__cuda_array_interface__`, it's straightforward to get a Numba array that views the allocation. From there only a little work is needed to cast it to the expected type and shape accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Improvements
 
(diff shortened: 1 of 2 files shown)


jakirkham:
rerun tests

(looks like we had an HTTP error downloading Conda packages in [this build])

PR #146, opened 2020-03-26
[REVIEW] quadtree-polygon/polyline pairing for spatial filtering

zhangjianting:
Given a quadtree structure/table for points with 5 columns (key/lev/sign/length/fpos) and a table of polygon bounding boxes with 4 columns (x1,y1,x2,y2), the code pairs up quadrants leaf nodes and polygons/polylines where quadrants intersect with their (expanded) bounding boxes. This is typically called the filtering phase which is known as the first phase in spatial joins. 
Depends on #145


harrism:
Moving to 0.15.

trxcllnt:
Files for review:
* [src/join/quadtree_poly_filtering.cu](https://github.com/zhangjianting/cuspatial/blob/fea-spatial-filtering/cpp/src/join/quadtree_poly_filtering.cu)
* [src/join/detail/traversal.cuh](https://github.com/zhangjianting/cuspatial/blob/fea-spatial-filtering/cpp/src/join/detail/traversal.cuh)
* [src/join/detail/intersection.cuh](https://github.com/zhangjianting/cuspatial/blob/fea-spatial-filtering/cpp/src/join/detail/intersection.cuh)
* [tests/spatial/polygon_bbox_test.cu](https://github.com/zhangjianting/cuspatial/blob/fea-spatial-filtering/cpp/tests/spatial/polygon_bbox_test.cu)
* [tests/spatial/polyline_bbox_test.cu](https://github.com/zhangjianting/cuspatial/blob/fea-spatial-filtering/cpp/tests/spatial/polyline_bbox_test.cu)

cpp/src/join/detail/intersection.cuh
```diff
@@ -0,0 +1,144 @@
...
+  auto nodes_and_polys = make_zip_iterator(
+    node_indices,
+    thrust::make_permutation_iterator(quadtree.column(0).begin<uint32_t>(), node_indices),  // keys
+    thrust::make_permutation_iterator(quadtree.column(1).begin<uint8_t>(), node_indices),  // levels
+    thrust::make_permutation_iterator(quadtree.column(2).begin<bool>(), node_indices),  // is_quad
+    poly_indices,
+    thrust::make_permutation_iterator(poly_bbox.column(0).begin<T>(), poly_indices),   // poly_x_min
+    thrust::make_permutation_iterator(poly_bbox.column(1).begin<T>(), poly_indices),   // poly_y_min
+    thrust::make_permutation_iterator(poly_bbox.column(2).begin<T>(), poly_indices),   // poly_x_max
+    thrust::make_permutation_iterator(poly_bbox.column(3).begin<T>(), poly_indices));  // poly_y_max
```
trxcllnt, line 100: I'm curious whether using `thrust::make_permutation_iterator()` multiple times with the same indices leads to multiple reads from global memory, or if that's cached. If so, I can refactor this to be one `permutation_iterator()` per set of indices like this:
```c++
  auto nodes_and_polys = make_zip_iterator(
    node_indices,
    thrust::make_permutation_iterator(make_zip_iterator(
      quadtree.column(0).begin<uint32_t>(), // keys
      quadtree.column(1).begin<uint8_t>(),  // levels
      quadtree.column(2).begin<bool>()      // is_quad
    ), node_indices),
    poly_indices,
    thrust::make_permutation_iterator(make_zip_iterator(
      poly_bbox.column(0).begin<T>(),   // poly_x_min
      poly_bbox.column(1).begin<T>(),   // poly_y_min
      poly_bbox.column(2).begin<T>(),   // poly_x_max
      poly_bbox.column(3).begin<T>()    // poly_y_max
    ), poly_indices)
  );
```

cpp/src/join/detail/traversal.cuh
```diff
@@ -0,0 +1,120 @@
...
+                 cudaStream_t stream)
+{
+  // compute the total number of child nodes
+  auto num_children =
+    thrust::reduce(rmm::exec_policy(stream)->on(stream), lengths, lengths + num_quads);
+
+  // exclusive scan on the number of child nodes to compute the offsets
+  rmm::device_uvector<uint32_t> parent_offsets(num_quads, stream);
+  thrust::exclusive_scan(
+    rmm::exec_policy(stream)->on(stream), lengths, lengths + num_quads, parent_offsets.begin());
```
harrism, line 57: You could combine these. No need to compute the num_children with a separate reduction if you do an inclusive scan  offset by 1 and read back the last element:

(note two `+ 1` offsets below, and also need to set the first element of the scan results to 0)

```suggestion
  // scan on the number of child nodes to compute the offsets
  rmm::device_uvector<uint32_t> parent_offsets(num_quads + 1, stream);
  thrust::inclusive_scan(
    rmm::exec_policy(stream)->on(stream), lengths, lengths + num_quads, parent_offsets.begin() + 1);
  parent_offsets.set_element_async(0, 0, stream);
  auto num_children = parent_offsets.back_element(stream); // synchronizes stream
```

I think the synchronization here is less heavy handed than `thrust::reduce` as well (and doesn't happen until you need `num_children`.

cpp/src/join/detail/traversal.cuh
```diff
@@ -0,0 +1,120 @@
+/*
```
harrism, line 1: This header contains an implementation, and it appears to only be used in one place in this PR. Will be it included from other files in future PRs? If not, could the implementation just move into the .cu file that includes it?

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+
+template <typename T>
+inline std::unique_ptr<cudf::table> join_quadtree_and_bboxes(cudf::table_view const &quadtree,
+                                                             cudf::table_view const &poly_bbox,
+                                                             T x_min,
+                                                             T x_max,
+                                                             T y_min,
+                                                             T y_max,
+                                                             T scale,
+                                                             cudf::size_type max_depth,
```
harrism, line 47: ?

```suggestion
                                                             int8_t max_depth,
```

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+                                                             rmm::mr::device_memory_resource *mr,
+                                                             cudaStream_t stream)
+{
+  auto const node_levels  = quadtree.column(1);  // uint8_t
+  auto const node_counts  = quadtree.column(3);  // uint32_t
+  auto const node_offsets = quadtree.column(4);  // uint32_t
+
+  auto num_polys = poly_bbox.num_rows();
```
harrism, line 55: Could const this too...

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+  rmm::device_uvector<uint8_t> out_types(num_pairs, stream);
+  rmm::device_uvector<uint8_t> out_levels(num_pairs, stream);
+  rmm::device_uvector<uint32_t> out_node_idxs(num_pairs, stream);
+  rmm::device_uvector<uint32_t> out_poly_idxs(num_pairs, stream);
+
+  // A zip iterator for the intermediate intersections or parent quadrants found during traversal
+  auto tmp_pairs = make_zip_iterator(
+    tmp_types.begin(), tmp_levels.begin(), tmp_node_idxs.begin(), tmp_poly_idxs.begin());
+
+  // A zip iterator for the intersecting poly and quad indicies found
```
harrism, line 85:
```suggestion
  // A zip iterator for the intersecting poly and quad indices found
```

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+  auto out_pairs = make_zip_iterator(
+    out_types.begin(), out_levels.begin(), out_node_idxs.begin(), out_poly_idxs.begin());
+
+  cudf::size_type num_results{0};
+  cudf::size_type num_parents{0};
+  cudf::size_type num_children{0};
+
+  // Traverse the quadtree starting at level 0 and descending to `max_depth` or until no more parent
+  // quadrants are found.
+  for (cudf::size_type level{0}; level < max_depth; ++level) {
```
harrism, line 95: ?

```suggestion
  for (int8_t level{0}; level < max_depth; ++level) {
```

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+    if (max_num_results > out_types.capacity()) {
+      // grow preallocated output sizes in multiples of the current capacity
+      auto new_size = out_types.capacity() * ((max_num_results / out_types.capacity()) + 1);
+      out_types.resize(new_size, stream);
+      out_levels.resize(new_size, stream);
+      out_node_idxs.resize(new_size, stream);
+      out_poly_idxs.resize(new_size, stream);
+      out_pairs = make_zip_iterator(
+        out_types.begin(), out_levels.begin(), out_node_idxs.begin(), out_poly_idxs.begin());
+    }
```
harrism, line 109: This is not done on level == 0. So can you factor level 0 out of the loop, and then have a simpler loop?

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+                                       tmp_levels,
+                                       tmp_node_idxs,
+                                       tmp_poly_idxs,
+                                       stream);
+
+    num_pairs     = std::get<0>(next_level);
+    tmp_types     = std::move(std::get<1>(next_level));
+    tmp_levels    = std::move(std::get<2>(next_level));
+    tmp_node_idxs = std::move(std::get<3>(next_level));
+    tmp_poly_idxs = std::move(std::get<4>(next_level));
```
harrism, line 178: Can't you  just use `std::tie(num_pairs, tmp_types, ...)` above so you don't have to move things?

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+    // update tmp_pairs iterator to get ready for next level iteration
+    tmp_pairs = make_zip_iterator(
```
harrism, line 180: Alternatively, just pass `next_level` directly to `make_zip_iterator`, since it takes a tuple, either way eliminates all the `move` calls.

cpp/src/join/quadtree_poly_filtering.cu
```diff
@@ -0,0 +1,292 @@
...
+                                            cudf::table_view const &poly_bbox,
+                                            double x_min,
+                                            double x_max,
+                                            double y_min,
+                                            double y_max,
+                                            double scale,
+                                            cudf::size_type max_depth,
+                                            rmm::mr::device_memory_resource *mr)
+{
+  CUSPATIAL_EXPECTS(quadtree.num_columns() == 5, "quadtree table must have 5 columns");
```
harrism, line 273: I didn't see tests for all of these expectations.

cpp/include/cuspatial/point_quadtree.hpp
```diff
@@ -50,9 +50,9 @@ namespace cuspatial {
  * columns for a complete quadtree:
  *     key - UINT32 column of quad node keys
  *   level - UINT8 column of quadtree levels
- * is_node - BOOL8 column indicating whether the node is a leaf or not
- *  length - UINT32 column for the number of child nodes (if is_node), or number of points
- *  offset - UINT32 column for the first child position (if is_node), or first point position
+ * is_quad - BOOL8 column indicating whether the node is a leaf or not
```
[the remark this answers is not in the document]
trxcllnt: The description should be changed since `false/0 indicate a leaf, true/1 indicate a quad.

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -0,0 +1,87 @@
...
+ */
+// std::unique_ptr<cudf::table> pip_refine(
+//   cudf::table_view const& pq_pair,
+//   cudf::table_view const& quadtree,
+//   cudf::table_view const& pnt,
+//   cudf::column_view const& poly_fpos,
+//   cudf::column_view const& poly_rpos,
+//   cudf::column_view const& poly_x,
+//   cudf::column_view const& poly_y,
+//   rmm::mr::device_memory_resource* mr = rmm::mr::get_default_resource());
```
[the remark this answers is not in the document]
trxcllnt: If #149 didn't exist, then yes. This definition was left in the header by accident, so I commented it out. It's in #149, and left here to avoid merge conflicts when merging updates from #146 into #149.

cpp/src/join/detail/intersection.cuh
```diff
@@ -0,0 +1,144 @@
...
+
+#include "indexing/construction/detail/utilities.cuh"
+#include "utility/z_order.cuh"
+
+namespace cuspatial {
+namespace detail {
+
+static __device__ uint8_t const leaf_indicator = 0;
+static __device__ uint8_t const quad_indicator = 1;
+static __device__ uint8_t const none_indicator = 2;
```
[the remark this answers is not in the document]
trxcllnt: I seem to recall a compiler error when I tried this before, but I'll try it again now.

cpp/src/join/detail/intersection.cuh
```diff
@@ -0,0 +1,144 @@
...
+    thrust::make_permutation_iterator(poly_bbox.column(0).begin<T>(), poly_indices),   // poly_x_min
+    thrust::make_permutation_iterator(poly_bbox.column(1).begin<T>(), poly_indices),   // poly_y_min
+    thrust::make_permutation_iterator(poly_bbox.column(2).begin<T>(), poly_indices),   // poly_x_max
+    thrust::make_permutation_iterator(poly_bbox.column(3).begin<T>(), poly_indices));  // poly_y_max
+
+  thrust::transform(rmm::exec_policy(stream)->on(stream),
+                    nodes_and_polys,
+                    nodes_and_polys + num_pairs,
+                    node_pairs,
+                    [=] __device__(auto const &node_and_poly) {
```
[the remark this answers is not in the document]
trxcllnt: This space is controlled by clang-format.

cpp/src/join/detail/traversal.cuh
trxcllnt: My two motivations for `intersection.cuh` and `traversal.cuh`:
1. Could be reused in other poly/quad traversal algorithms?
2. Avoid a 700+ line `quadtree_poly_filtering.cu` 

I'm not sure whether these can be reused, or if they're too specific, but my instinct was to factor them out by default.

cpp/src/join/detail/intersection.cuh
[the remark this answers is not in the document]
trxcllnt: Ah yes. Thrust doesn't like comparing pairs of `uint8_t` and `intersection_type`:
```
../../../../src/join/detail/intersection.cuh (51): error: no operator "==" matches these operands
            operand types are: const thrust::detail::uint8_t == cuspatial::detail::intersection_type
```

The alternative is casting all uses of `intersection_type` which seems uglier than necessary:
```c++
return thrust::get<0>(t) == static_cast<uint8_t>(intersection_type::leaf);
```

I also can't create the `is_quad` iterator as `column(2).begin<intersection_type>()`, because the output iterator that's passed in has no knowledge of the `intersection_type` enum, leading to this error:
```
_deps/thrust-src/thrust/detail/tuple.inl (320): error: a value of type "cuspatial::detail::intersection_type" cannot be assigned to an entity of type "uint8_t"
```

cpp/src/join/quadtree_poly_filtering.cu
trxcllnt: You're right, I forgot `tie()` participates in rvo.

cpp/src/join/detail/intersection.cuh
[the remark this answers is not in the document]
harrism: Alternative suggestion: use a weakly-typed `enum`, or:
```suggestion
constexpr uint8_t leaf_indicator = 0;
constexpr uint8_t quad_indicator = 1;
constexpr uint8_t none_indicator = 2;
```

cpp/src/join/detail/intersection.cuh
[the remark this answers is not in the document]
harrism: That's a bug. ;)

cpp/src/join/detail/intersection.cuh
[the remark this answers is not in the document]
trxcllnt: nvcc is giving me this error when I try to change these to `uint8_t constexpr`, so I think our `--expt-relaxed-constexpr` isn't making it into the compile flags somehow...
```
intersection.cuh (135): error: identifier "cuspatial::detail::none_indicator" is undefined in device code
```

cpp/src/join/detail/intersection.cuh
trxcllnt: I simplified the logic by doing the lookups inside the `transform` lambda directly.

cpp/src/join/detail/traversal.cuh
```diff
@@ -0,0 +1,120 @@
...
+#include <tuple>
+
+#include "indexing/construction/detail/utilities.cuh"
+#include "utility/z_order.cuh"
+
+namespace cuspatial {
+namespace detail {
+
+template <typename LengthsIter, typename OffsetsIter>
+inline std::tuple<cudf::size_type,
```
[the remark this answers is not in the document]
trxcllnt: Added a docstring that describes the arguments/return tuple.

cpp/src/join/detail/traversal.cuh
trxcllnt: Great suggestion! I updated the implementation.

cpp/src/join/quadtree_poly_filtering.cu
trxcllnt: Restructured the loop, hopefully it's a bit simpler.

cpp/src/join/quadtree_poly_filtering.cu
trxcllnt: Added a test for these :+1:

cpp/include/cuspatial/spatial_join.hpp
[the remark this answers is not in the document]
trxcllnt: See answer here: https://github.com/rapidsai/cuspatial/pull/146#discussion_r452412639

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <memory>
+
+namespace cuspatial {
+
+/**
+ * @brief Search a quadtree for polygon or polyline bounding box intersections.
+ *
+ * @note Swaps `x_min` and `x_max` if `x_min > x_max`.
+ * @note Swaps `y_min` and `y_max` if `y_min > y_max`.
+ * @note `scale` is applied to (x - x_min) and (y - y_min) to convert coordinates into a Morton code
```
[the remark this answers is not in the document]
trxcllnt: From my understanding, using a larger scale than the quadtree was created with could cause more quad nodes to intersect with with each polygon.

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * @param x_max The upper-right x-coordinate of the area of interest bounding box.
+ * @param y_min The lower-left y-coordinate of the area of interest bounding box.
+ * @param y_max The upper-right y-coordinate of the area of interest bounding box.
+ * @param scale Scale to apply to each x and y distance from x_min and y_min.
+ * @param max_depth Maximum quadtree depth.
+ *
+ * @return A cudf table with two columns:
+ * poly_offset - INT32 column of indices for each poly bbox that intersects with the quadtree.
+ * quad_offset - INT32 column of indices for each leaf quadrant intersecting with a poly bbox.
+ */
```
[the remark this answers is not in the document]
trxcllnt: Done, thanks for the reminder!

cpp/include/cuspatial/spatial_join.hpp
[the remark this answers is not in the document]
cwharris: I'm wondering if it makes sense to have a data structure that keeps that value around along side the columns. :shrug:

PR #143, opened 2020-03-26
[REVIEW] construct quadtree on large-scale point data

zhangjianting:
Implementation of an efficient bottom-up approach to constructing quadtree on large-scale point data.

PR #149, opened 2020-03-26
[REVIEW] Add quadtree-based point-in-polygon and point-to-nearest-polyline

zhangjianting:
Depends on #143, #145 and #146  
Given a quadtree constructed from a set of points, bounding boxes of polygons and quadrant-polygon paris, the code pairs up points and polygons based on point-in-polygon test. This is a significant enhancement to existing point-in-polygon test API in cuspatial by utilizing point quadtree indices, polygon bounding box approximation, level-wise BFS search for matching point quadrants and polygons before finally performing point-in-polygon test for all points in matched quadrant-polygon pairs and generate point-polygon pairs. The output is a dense representation of point-polygon pairs which requires a two-phase approach to precisely count the memory footprint for the output pairs before actually writing to output memory locations.   

In addition to the two unit tests (_small, _large), a separate test on NYC yearly taxi trip pickup locations (~170 million) and three polygon datasets (taxi zones, community districts and census tracts) with hundreds to thousands of single-/multi- ring polygons has been developed as a demo/benchmark to verify the correctness on real world data and compare the code's performance with implementations based on GDAL. The NYC test code will be released under a different category later.   

Similar to point-in-polygon test based spatial refinement, point-to-polyline distance based nearest neighbor/distance refinement can also be implemented. As a point has no more than one nearest neighbor (zero for points in quadrants when quadrants and expanded polyline bbox do not intersect), there is no need to apply the two-phase strategy, as applied to point-in-polygon test based spatial refinement, to precisely allocate output memory. One phase is sufficient in this case.


zhangjianting:
rerun tests

harrism:
Does this PR include commits from other PRs? If so, put "Depends on #XXXX" in the description.

harrism:
Moving to 0.15

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  auto d_point_y       = cudf::column_device_view::create(point_y, stream);
+  auto d_poly_offsets  = cudf::column_device_view::create(poly_offsets, stream);
+  auto d_ring_offsets  = cudf::column_device_view::create(ring_offsets, stream);
+  auto d_poly_points_x = cudf::column_device_view::create(poly_points_x, stream);
+  auto d_poly_points_y = cudf::column_device_view::create(poly_points_y, stream);
+
+  quad_pip_phase1_kernel<T><<<num_unit_pairs, threads_per_block, 0, stream>>>(d_pq_poly_idx.data(),
+                                                                              d_pq_quad_idx.data(),
+                                                                              d_quad_offset.data(),
+                                                                              d_quad_length.data(),
```
trxcllnt, line 381: Is there a better way to pass a `device_uvector` to a kernel, similar to how we can pass a `cudf::column_device_view`?

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## New Features
 
 - PR #146 quadtree-polygon pairing for spatial filtering
+- PR #149 Add point-in-polygon spatial refinement using quadtree indexing
```
harrism, line 6: Fix duplicate entries.

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
-// std::unique_ptr<cudf::table> pip_refine(
-//   cudf::table_view const& pq_pair,
-//   cudf::table_view const& quadtree,
-//   cudf::table_view const& pnt,
-//   cudf::column_view const& poly_fpos,
-//   cudf::column_view const& poly_rpos,
-//   cudf::column_view const& poly_x,
-//   cudf::column_view const& poly_y,
-//   rmm::mr::device_memory_resource* mr = rmm::mr::get_default_resource());
+ * @brief Finds points in a set of (polygon, quadrant) pairs derived from spatial filtering.
```
harrism, line 67: This brief is unclear to me. What points does it find? What does "derived from spatial filtering" mean?

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
+ * @param ring_offsets Begin indices of the first point in each ring (i.e. prefix-sum).
+ * @param poly_points_x Polygon point x-coodinates.
+ * @param poly_points_y Polygon point y-coodinates.
+ * @param mr The optional resource to use for output device memory allocations.
+ *
+ * @throw cuspatial::logic_error If the poly_quad_pairs table is malformed.
+ * @throw cuspatial::logic_error If the quadtree table is malformed.
+ * @throw cuspatial::logic_error If the number of point indices doesn't match the number of points.
+ * @throw cuspatial::logic_error If the number of rings is less than the number of polygons.
+ * @throw cuspatial::logic_error If each ring has fewer than three vertices.
```
harrism, line 84: Each ring? Or any ring?

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
+ * @throw cuspatial::logic_error If the types of point and polygon vertices are different.
+ *
+ * @returns A cudf table with two columns of point/polygon pairs for each intersection:
```
harrism, line 87: Are there two columns per intersection or one pair per intersection? Clarify.

harrism:
Seems this PR introduces two public APIs: `quadtree_point_in_polygon` and `quadtree_point_to_nearest_polyline`. If that's the case, suggest changing the PR title to say this rather than what it says now (which is unclear).

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
+  cudf::column_view const& point_x,
+  cudf::column_view const& point_y,
+  cudf::column_view const& poly_offsets,
+  cudf::column_view const& ring_offsets,
+  cudf::column_view const& poly_points_x,
+  cudf::column_view const& poly_points_y,
+  rmm::mr::device_memory_resource* mr = rmm::mr::get_default_resource());
+
+/**
+ * @brief given a vector of paired of quadrants and polylines, for each points in a quadrant,
```
harrism, line 108:
```suggestion
 * @brief given a vector of pairs of quadrants and polylines, for each point in a quadrant,
```

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
+ * find its nearest polylines and the corresponding distance between the point and the polygon
```
harrism, line 109:
```suggestion
 * find its nearest polyline and the corresponding distance between the point and the polyline.
```

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
+ *
+ * @param poly_quad_pairs Table of (quadrant, polygon) index pairs derived from spatial filtering.
```
harrism, line 111:
```suggestion
 * @param poly_quad_pairs Table of (quadrant, polyline) index pairs derived from spatial filtering.
```

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -51,44 +52,97 @@ namespace cuspatial {
...
+ * @param point_y y-coordinates of points to test
+ * @param poly_offsets Begin indices of the first ring in each polyline (i.e. prefix-sum)
+ * @param poly_points_x Polyline point x-coordinates
+ * @param poly_points_y Polyline point y-coordinates
+ * @param mr The optional resource to use for output device memory allocations.
+ *
+ * @throw cuspatial::logic_error If the poly_quad_pairs table is malformed.
+ * @throw cuspatial::logic_error If the quadtree table is malformed.
+ * @throw cuspatial::logic_error If the number of point indices doesn't match the number of points.
+ * @throw cuspatial::logic_error If each polyline has fewer than two vertices.
```
harrism, line 124: Each? or any?

cpp/include/cuspatial/spatial_join.hpp
trxcllnt: Great point. "derived from spatial filtering" means "the table returned by `quad_bbox_join()`", but I'll modify this to be more clear.

cpp/include/cuspatial/spatial_join.hpp
trxcllnt: Aye good call, this should be _any_ ring.

cpp/include/cuspatial/spatial_join.hpp
trxcllnt: One pair per intersection, but I can see how the wording can be better :pray:

cpp/include/cuspatial/spatial_join.hpp
trxcllnt: On a related note, I think the `quad_bbox_join` name could be more descriptive. Would that be good to add to this PR before it's released in 0.15?

trxcllnt:
This PR also includes the Cython/Python bindings, and adds the related methods to the api docs index.

cpp/src/join/quadtree_point_in_polygon.cu
harrism: Nope.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+                                          cudf::column_view const &point_x,
+                                          cudf::column_view const &point_y,
+                                          cudf::column_view const &poly_offsets,
+                                          cudf::column_view const &ring_offsets,
+                                          cudf::column_view const &poly_points_x,
+                                          cudf::column_view const &poly_points_y,
+                                          rmm::mr::device_memory_resource *mr,
+                                          cudaStream_t stream)
+  {
+    return compute_quadtree_point_in_polygon<T>(poly_quad_pairs.column(0),
```
harrism, line 468: Possibly an unnecessary layer? Could move body of `compute_quadtree_point_in_polygon` here?

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+    // copy idx of orginal pairs to all sub-pairs
+    thrust::inclusive_scan(rmm::exec_policy(stream)->on(stream),
+                           d_quad_offset.begin(),
+                           d_quad_offset.end(),
+                           d_quad_offset.begin(),
+                           thrust::maximum<int>());
+
+    // d_num_sums and d_num_units are no longer needed, so don't return them
+
+    return std::move(d_quad_offset);
```
harrism, line 320: Shouldn't need `std::move` on return due to Return Value Optimization

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+                           d_num_units.end(),
+                           d_num_sums.begin());
+
+    // allocate memory for sub-pairs with four components:
+    // (polygon_idx, quadrant_idx, offset, length)
+
+    rmm::device_uvector<uint32_t> d_quad_offset(num_unit_pairs, stream);
+
+    thrust::fill(
+      rmm::exec_policy(stream)->on(stream), d_quad_offset.begin(), d_quad_offset.end(), 0);
```
harrism, line 302: I think when zero-filling it's better to use `cudaMemsetAsync`, which I doubt Thrust uses.  (Faster in my tests)
 
```suggestion
    cudaMemsetAsync(d_quad_offset.data(), 0, num_unit_pairs * sizeof(uint32_t), stream);
```

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+
+  thrust::gather(rmm::exec_policy(stream)->on(stream),
+                 d_quad_offset.begin(),
+                 d_quad_offset.end(),
+                 quad_indices.begin<uint32_t>(),
+                 d_pq_quad_idx.begin());
+
+  // allocate memory to store numbers of points in polygons in all sub-pairs initialized to 0
+  rmm::device_uvector<uint32_t> d_num_hits(num_unit_pairs, stream);
+  thrust::fill(rmm::exec_policy(stream)->on(stream), d_num_hits.begin(), d_num_hits.end(), 0);
```
harrism, line 342: cudaMemsetAsync...

Actually, quad_pip_phase1_kernel seems to initialize this array (see line 136), so I don't think you need to initialize it to zero at all.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+                 d_pq_poly_idx.begin());
+
+  thrust::gather(rmm::exec_policy(stream)->on(stream),
+                 d_quad_offset.begin(),
+                 d_quad_offset.end(),
+                 quad_indices.begin<uint32_t>(),
+                 d_pq_quad_idx.begin());
+
+  // allocate memory to store numbers of points in polygons in all sub-pairs initialized to 0
+  rmm::device_uvector<uint32_t> d_num_hits(num_unit_pairs, stream);
```
harrism, line 341: Move this definition closer to its first use (quad_pip_phase1_kernel launch).

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  thrust::fill(rmm::exec_policy(stream)->on(stream), d_num_hits.begin(), d_num_hits.end(), 0);
+
+  // generate offsets of sub-paris within the orginal pairs
```
harrism, line 344:
```suggestion
  // generate offsets of sub-pairs within the orginal pairs
```

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+                                                                              *d_point_indices,
+                                                                              *d_point_x,
+                                                                              *d_point_y,
+                                                                              *d_poly_offsets,
+                                                                              *d_ring_offsets,
+                                                                              *d_poly_points_x,
+                                                                              *d_poly_points_y,
+                                                                              d_num_hits.data());
+
+  CUDA_TRY(cudaStreamSynchronize(stream));
```
harrism, line 392: I don't think this cudaStreamSync is needed.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  d_quad_offset.resize(num_valid_pair, stream);
+  d_quad_length.resize(num_valid_pair, stream);
+  d_num_hits.resize(num_valid_pair, stream);
+
+  uint32_t total_hits =
+    thrust::reduce(rmm::exec_policy(stream)->on(stream), d_num_hits.begin(), d_num_hits.end());
+
+  // prefix sum on numbers to generate offsets
+  thrust::exclusive_scan(
+    rmm::exec_policy(stream)->on(stream), d_num_hits.begin(), d_num_hits.end(), d_num_hits.begin());
```
harrism, line 420: Doing a reduction and a prefix sum on the same array is overkill, you can get the the scan and the sum in one go.  If you make `d_num_hits` one element larger than needed initially (and operate on n-1 elements everywhere above):

```suggestion
  // prefix sum on numbers to generate offsets
  thrust::exclusive_scan(
    rmm::exec_policy(stream)->on(stream), d_num_hits.begin(), d_num_hits.end(), d_num_hits.begin());
  uint32_t total_hits = d_num_hits.back_element(stream);
```
Note `back_element` calls `element`, which synchronizes the stream so `total_hits` is valid and ready when it returns.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+                                                                              *d_point_y,
+                                                                              *d_poly_offsets,
+                                                                              *d_ring_offsets,
+                                                                              *d_poly_points_x,
+                                                                              *d_poly_points_y,
+                                                                              d_num_hits.data(),
+                                                                              *d_poly_index_col,
+                                                                              *d_point_index_col);
+
+  CUDA_TRY(cudaStreamSynchronize(stream));
```
harrism, line 445: sync not needed here since this function takes a stream.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  if (poly_quad_pairs.num_rows() == 0 || quadtree.num_rows() == 0 || point_indices.size() == 0 ||
+      poly_offsets.size() == 0) {
+    std::vector<std::unique_ptr<cudf::column>> cols{};
+    cols.reserve(2);
+    cols.push_back(cudf::make_empty_column(cudf::data_type{cudf::type_id::UINT32}));
+    cols.push_back(cudf::make_empty_column(cudf::data_type{cudf::type_id::UINT32}));
+    return std::make_unique<cudf::table>(std::move(cols));
+  }
+
+  return detail::quadtree_point_in_polygon(poly_quad_pairs,
```
harrism, line 560: Probably need to sync stream 0 here before returning, rather than deeper down in the functions that take streams.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  if (threadIdx.x == 0) {
+    quad_idx     = pq_quad_id[blockIdx.x];
+    poly_idx     = pq_poly_id[blockIdx.x];
+    qpos         = block_offset[blockIdx.x];
+    num_point    = block_length[blockIdx.x];
+    first_pos    = quad_offsets.element<uint32_t>(quad_idx) + qpos;
+    num_adjusted = ((num_point - 1) / num_threads_per_warp + 1) * num_threads_per_warp;
+  }
+
+  __syncthreads();
```
harrism, line 77: This syncthreads is redundant since there is one after the next line too.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+        bool y_between_by_ay = y1 <= y && y < y0;  // is y in range [by, ay) when by < ay
+        bool y_in_bounds     = y_between_ay_by || y_between_by_ay;  // is y in range [by, ay]
+        T run                = x1 - x0;
+        T rise               = y1 - y0;
+        T rise_to_point      = y - y0;
+
+        if (y_in_bounds && x < (run / rise) * rise_to_point + x0) { in_polygon = not in_polygon; }
+      }
+    }
+  }
```
harrism, line 118: I think most of this core computation here should be in its own function for clarity / modularity.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+                                       cudf::column_device_view const poly_offsets,
+                                       cudf::column_device_view const ring_offsets,
+                                       cudf::column_device_view const poly_points_x,
+                                       cudf::column_device_view const poly_points_y,
+                                       uint32_t *num_hits)
+{
+  __shared__ uint32_t quad_idx, poly_idx, num_point, first_pos, qpos, num_adjusted;
+
+  // assume # of points/threads no more than num_threads_per_warp * max_warps_per_block (32*32)
+  __shared__ uint16_t data[max_warps_per_block];
```
harrism, line 65: `data` is not a very helpful name. I think this is `warp_num_hits` or something?

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+
+#include <vector>
+
+namespace cuspatial {
+namespace detail {
+namespace {
+
+static uint8_t const max_warps_per_block  = 32;
+static uint8_t const num_threads_per_warp = 32;
+static uint32_t const threads_per_block   = 256;
```
harrism, line 45:
```suggestion
constexpr uint8_t max_warps_per_block  = 32;
constexpr uint8_t num_threads_per_warp = 32;
constexpr uint32_t threads_per_block   = 256;
```

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+#include <thrust/iterator/transform_iterator.h>
+#include <thrust/reduce.h>
+
+#include <vector>
+
+namespace cuspatial {
+namespace detail {
+namespace {
+
+static uint8_t const max_warps_per_block  = 32;
```
harrism, line 43: If threads_per_block is 256, how can max_warps_per_block be 32? Shouldn't this be 8?

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  }
+
+  __syncthreads();
+
+  if (threadIdx.x < max_warps_per_block) {
+    uint32_t num = data[threadIdx.x];
+    for (uint32_t offset = max_warps_per_block / 2; offset > 0; offset /= 2) {
+      num += __shfl_xor_sync(0xFFFFFFFF, num, offset);
+    }
+    if (threadIdx.x == 0) { num_hits[blockIdx.x] = num; }
```
harrism, line 136: This appears to be a warp reduction. Use CUB for this. That way we automatically use hardware intrinsics for warp sums where they exist (assuming CUB is up to date).

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+    poly_idx     = pq_poly_id[blockIdx.x];
+    qpos         = block_offset[blockIdx.x];
+    num_point    = block_length[blockIdx.x];
+    mem_offset   = num_hits[blockIdx.x];
+    first_pos    = quad_offsets.element<uint32_t>(quad_idx) + qpos;
+    sums[0]      = 0;
+    num_adjusted = ((num_point - 1) / num_threads_per_warp + 1) * num_threads_per_warp;
+  }
+
+  __syncthreads();
```
harrism, line 174:
```suggestion
```

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+        bool y_between_by_ay = y1 <= y && y < y0;  // is y in range [by, ay) when by < ay?
+        bool y_in_bounds     = y_between_ay_by || y_between_by_ay;  // is y in range [by, ay]?
+        T run                = x1 - x0;
+        T rise               = y1 - y0;
+        T rise_to_point      = y - y0;
+
+        if (y_in_bounds && x < (run / rise) * rise_to_point + x0) { in_polygon = not in_polygon; }
+      }
+    }
+  }
```
harrism, line 215: Does this code duplicate the code in the first kernel? If not, it's very similar. Try to factor it out.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+  if (threadIdx.x < num_threads_per_warp) {
+    uint32_t num = temp[threadIdx.x];
+    for (uint8_t i = 1; i <= num_threads_per_warp; i *= 2) {
+      int n = __shfl_up_sync(0xFFFFFFF, num, i, num_threads_per_warp);
+      if (threadIdx.x >= i) num += n;
+    }
+    sums[threadIdx.x + 1] = num;
+  }
+
+  __syncthreads();
```
harrism, line 237: This is also very similar to phase 1 kernel Factor out as much as possible.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -0,0 +1,573 @@
...
+
+  // warp-level scan; only one warp is used
+  if (threadIdx.x < num_threads_per_warp) {
+    uint32_t num = temp[threadIdx.x];
+    for (uint8_t i = 1; i <= num_threads_per_warp; i *= 2) {
+      int n = __shfl_up_sync(0xFFFFFFF, num, i, num_threads_per_warp);
+      if (threadIdx.x >= i) num += n;
```
harrism, line 232: CUB warp reduce.

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -0,0 +1,332 @@
...
+    *cudf::column_device_view::create(point_x, stream),
+    *cudf::column_device_view::create(point_y, stream),
+    *cudf::column_device_view::create(poly_offsets, stream),
+    *cudf::column_device_view::create(poly_points_x, stream),
+    *cudf::column_device_view::create(poly_points_y, stream),
+    *cudf::mutable_column_device_view::create(*point_index_col, stream),
+    *cudf::mutable_column_device_view::create(*poly_index_col, stream),
+    *cudf::mutable_column_device_view::create(*poly_distance_col, stream));
+
+  CUDA_TRY(cudaStreamSynchronize(stream));
```
harrism, line 208: As elsewhere, probably should synchronize outside of this function, since this function takes a stream.

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -0,0 +1,332 @@
...
+#include <thrust/iterator/transform_iterator.h>
+#include <thrust/reduce.h>
+
+#include <vector>
+
+namespace cuspatial {
+namespace detail {
+namespace {
+
+static uint32_t const threads_per_block = 256;
```
harrism, line 43:
```suggestion
constexpr uint32_t threads_per_block = 256;
```

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -0,0 +1,332 @@
...
+        T r  = (d1x * d2x + d1y * d2y) / d1;
+        T d  = 1e20;
+        if (r <= 0 || r >= d1) {
+          T d3x = px - x1, d3y = py - y1;
+          T d2 = sqrt(d2x * d2x + d2y * d2y);
+          T d3 = sqrt(d3x * d3x + d3y * d3y);
+          d    = min(min(d, d2), d3);
+        } else {
+          d = sqrt((d2x * d2x + d2y * d2y) - (r * r));
+        }
```
harrism, line 113: For a function that only needs to find the nearest distance, this is doing a lot of expensive sqrts. 

If you only want to find the minimum distance, you should only need to calculate *distance squared*. Then, once you find the minimum distance squared, take the sqrt of just that distance-squared to find the minimum distance.

I bet this will be a lot faster...

cpp/tests/join/point_in_polygon_test_large.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+// #include <algorithm>
+
+/*
+ * The test uses the same quadtree structure as in pip_refine_test_small.
+ * However, the numbers of randomly generated points under all quadrants (min_size) are increased
+ * to be more than the number of threads per-block (currently fixed to 256, but can be set between
+ * 32 2048 (CUDA Compute Capacity 7.0, multiples of warp size, which is 32) The test is designed to
+ * fully test the two kernels in the refinment code, including both warp level reduce and scan, vote
+ * and popc. Thrust primitives to dvide quadrants into sub-blocks are also tested.
```
harrism, line 50:
```suggestion
 * and popc. Thrust primitives to divide quadrants into sub-blocks are also tested.
```

cpp/src/join/quadtree_point_in_polygon.cu
trxcllnt: cc: @zhangjianting

zhangjianting: > cc: @zhangjianting
max_warps_per_block suggests the maximum warps per block. For CC>=3.0, the number is 1024,

cpp/include/cuspatial/spatial_join.hpp
harrism: Probably yes.

cpp/src/join/quadtree_point_in_polygon.cu
harrism: OK, but it is only used to size shared memory. Instead could use dynamically sized shared memory based on the *actual* warps per block.

mike-wendt:
Removing @rapidsai/ops-codeowners from review as there are no CODEOWNER files in this PR, if you need our review please re-add us

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -98,16 +98,19 @@ __global__ void find_nearest_polyline_kernel(
...
         if (r <= 0 || r >= d1) {
-          T d3x = px - x1, d3y = py - y1;
-          T d2 = sqrt(d2x * d2x + d2y * d2y);
-          T d3 = sqrt(d3x * d3x + d3y * d3y);
-          d    = min(min(d, d2), d3);
+          T d3x = px - x1;
+          T d3y = py - y1;
+          T d2  = d2x * d2x + d2y * d2y;
+          T d3  = d3x * d3x + d3y * d3y;
+          d     = min(d, sqrt(min(d2, d3)));
```
harrism, line 113: You don't need the sqrt here!

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -98,16 +98,19 @@ __global__ void find_nearest_polyline_kernel(
...
-        T d1 = sqrt(d1x * d1x + d1y * d1y);
-        T r  = (d1x * d2x + d1y * d2y) / d1;
-        T d  = 1e20;
+        T d1x = x1 - x0;
+        T d1y = y1 - y0;
+        T d2x = px - x0;
+        T d2y = py - y0;
+        T d1  = sqrt(d1x * d1x + d1y * d1y);
+        T r   = (d1x * d2x + d1y * d2y) / d1;
+        T d   = 1e20;
```
harrism, line 107: Use numeric_limits. Also for the initialization of `distance` above.

```suggestion
        T d   = std::numeric_limits::max<T>();
```

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -98,16 +98,19 @@ __global__ void find_nearest_polyline_kernel(
...
-          T d2 = sqrt(d2x * d2x + d2y * d2y);
-          T d3 = sqrt(d3x * d3x + d3y * d3y);
-          d    = min(min(d, d2), d3);
+          T d3x = px - x1;
+          T d3y = py - y1;
+          T d2  = d2x * d2x + d2y * d2y;
+          T d3  = d3x * d3x + d3y * d3y;
+          d     = min(d, sqrt(min(d2, d3)));
         } else {
           d = sqrt((d2x * d2x + d2y * d2y) - (r * r));
```
harrism, line 115: You don't need the sqrt here either!

The key is to name the outer variable `distance_squared` instead of `distance`, and treat everything as a squared distance until you absolutely need the actual distance. That point is the last line of the function:

```
out_distance.element<T>(point_pos)           = sqrt(distance_squared);
```

trxcllnt: Yeah this is on my list of things to revisit, still trying to work out a better way to write the point-in-polygon logic.

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -192,191 +108,102 @@ struct dispatch_quadtree_point_in_polygon {
...
-                                  d_quad_offset.begin());
-
-    // assemble components in input/output iterators; note d_quad_offset used in both input and
-    // output
-    auto quad_block_id_iter = make_zip_iterator(d_pq_quad_idx.begin(), d_quad_offset.begin());
-    auto offset_length_iter = make_zip_iterator(d_quad_offset.begin(), d_quad_length.begin());
+    auto quad_lengths        = quadtree.column(3);
+    auto quad_offsets        = quadtree.column(4);
+    auto poly_indices        = poly_quad_pairs.column(0);
+    auto quad_indices        = poly_quad_pairs.column(1);
```
harrism, line 114: A thought, possibly for a different PR: we should probably encode the column indices somewhere (so you can access by name rather than number), or consider making a quadtree a data structure rather than a table -- I don't think anyone will pass in a table that is a quadtree from outside cuSpatial -- they will probably build it using cuSpatial, right?

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -192,191 +108,102 @@ struct dispatch_quadtree_point_in_polygon {
...
+
+    // Allocate output columns for the number of pairs that intersected
+    auto poly_idx_col  = make_fixed_width_column<uint32_t>(num_intersections, stream, mr);
+    auto point_idx_col = make_fixed_width_column<uint32_t>(num_intersections, stream, mr);
+
+    // Note: no need to resize `poly_idxs` or `point_idxs` if we set the end iterator to
+    // `idxs.begin() + num_intersections`.
+
+    // populate the polygon indices column
+    thrust::copy(rmm::exec_policy(stream)->on(stream),
```
harrism, line 192: Do you need the extra copy? Could you use `copy_remove_if` above?

trxcllnt: Oh sweet, I didn't know about `copy_remove_if`! :tada:

cpp/src/join/quadtree_point_in_polygon.cu
trxcllnt: I agree, making it a data structure would make things simpler all the way up to Python.

cpp/src/join/quadtree_point_in_polygon.cu
trxcllnt: I can't use `copy_remove_if` because the size of the output columns is determined by the number of elements removed by `remove_if`.

cpp/src/join/quadtree_point_to_nearest_polyline.cu
trxcllnt: success!

harrism:
@trxcllnt did the squared distances make a perf difference?

trxcllnt:
@harrism yes it did! 

```
old: 902 ms ± 2.97 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
new: 699 ms ± 14 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

PR #170, opened 2020-04-07
[REVIEW] Update coordinate_transform.hpp files to libcudf++

### LINKED ISSUE #161: [BUG] Update coordinate_transform.hpp to libcudf++
* Eliminate use of gdf_column in shapefile APIs.
* Remove all raw RMM_ALLOC/RMM_FREE and replace with rmm::device_buffer / device_vector.
* Add appropriate CUSPATIAL_EXPECTS calls. See #153

cwharris:
Contributes to #161


cwharris:
Hopefully #cudf/4828 fixed this build. Otherwise we'll need to add yet another `move` declaration. :man_shrugging:

cwharris:
Generic `move` pr was a no-go, so making our own `move.pxd` which imports all of `cudf`'s.

trxcllnt:
@cwharris It looks like the Python side needs an update to call `column.as_column()` in the `lonlat_to_xy_km_coordinates` function.

cwharris:
rerun tests

cpp/src/spatial/lonlat_to_coordinate.cu
```diff
@@ -0,0 +1,113 @@
...
+    std::enable_if_t<not std::is_floating_point<T>::value, pair_of_columns>
+    operator()(Args&&...)
+    {
+        CUSPATIAL_FAIL("Non-floating point operation is not supported");
+    }
+
+    template <typename T>
+    std::enable_if_t<std::is_floating_point<T>::value, pair_of_columns>
+    operator()(double origin_lon,
+               double origin_lat,
```
thomcom, line 43: Shouldn't origin_lon and origin_lat be type `T`, the same type as the column?

Also, from a performance standpoint, isn't there a substantial performance loss mixing 64 and 32 bit floats on the GPU?

cpp/src/spatial/lonlat_to_coordinate.cu
```diff
@@ -0,0 +1,113 @@
...
+                         {
+                            out_lon.element<T>(idx) = (origin_lon - in_lon.element<T>(idx))
+                                * 40000.0
+                                * cos((origin_lat + in_lon.element<T>(idx)) * M_PI / 360)
+                                / 360;
+
+                            out_lat.element<T>(idx) = (origin_lat - in_lat.element<T>(idx))
+                                * 40000.0
+                                / 360;
+                         });
```
thomcom, line 76: A question for @harrism  - isn't this the perfect situation to use a `thrust::transform` iterator, instead of a `for_each`, since the operation is identical and independent for every element of the input vectors?

cwharris: I believe using `transform` in this scenario would require two kernel invocations (one for lon, one for lat) and reading the latitude column twice (since it is used in the longitude calculation). Using `for_each` allows us to launch a single kernel and only read each input column once.

It would be nice to have a transform function which took a tuple of columns and returned a tuple of columns, accepted a tuple of Ts and returned a tuple of Ts, but that doesn't exist (until we write one).

trxcllnt: I think `thrust::zip_iterator` can only be used as an input iterator, not an output iterator. If we had something like `make_zip_output_iterator()`, we could use transform to read the two input columns and return a `thrust::tuple()` of the two output values.

zhangjianting: > I think `thrust::zip_iterator` can only be used as an input iterator, not an output iterator. If we had something like `make_zip_output_iterator()`, we could use transform to read the two input columns and return a `thrust::tuple` of the two output values.

Zip_iterator worked for me as both inputs and outputs.

cwharris: @zhangjianting are there docs or an example for this?

zhangjianting: > @zhangjianting are there docs or an example for this?

cpp/tests/spatial/coordinate_transform_test.cu

cpp/tests/spatial/coordinate_transform_test.cu
```diff
@@ -0,0 +1,48 @@
...
+#include <cuspatial/coordinate_transform.hpp>
+
+using cudf::test::fixed_width_column_wrapper;
+
+template <typename T>
+struct ShiftTest : public cudf::test::BaseFixture {};
+
+TYPED_TEST_CASE(ShiftTest, cudf::test::FloatingPointTypes);
+   
+TYPED_TEST(ShiftTest, CoordinateTest)
```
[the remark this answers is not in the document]
harrism: I also like to see tests of extreme cases, and error cases.

Extreme case: does this do what you expect when the origin of the output coordinates is at one of the poles?
Error case: does this correctly handle empty inputs? Non-floating-point inputs? Inputs with nulls?

cpp/src/spatial/lonlat_to_coordinate.cu
harrism: There's no loss from mixing. But 32-bit float operations are always faster than 64-bit. On some GeForce GPUs they are many times faster.

cpp/src/spatial/lonlat_to_coordinate.cu
cwharris: Thanks for the insight, everyone. I was thinking about this from the wrong perspective, but zip iterators as output iterators is great!

cpp/tests/spatial/coordinate_transform_test.cu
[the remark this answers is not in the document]
cwharris: Can you help me understand why having a test with a larger number of elements is beneficial? If I were implementing a kernel manually, I can see that it would verify threading was working as intended for large numbers of inputs - but I'm using thrust under the hood, so that's taken care of. I can see how a benchmark could be useful, but the actual result of the function could be ignored (since we're testing performance not correctness). I'm hesitant to add a CPU implementation because I'm not confident the GPU implementation makes logical sense as-is. it feels like adding a CPU implementation increases a false sense of correctness. I'm not sure of the exact use case for lonlat => cartesian transformation, so it's difficult to tell what's useful / correct.

cpp/src/spatial/lonlat_to_cartesian.cu
```diff
@@ -0,0 +1,149 @@
...
+                                               output_y->mutable_view().begin<T>());
+
+        auto input_zip = thrust::make_zip_iterator(input_iters);
+        auto output_zip = thrust::make_zip_iterator(output_iters);
+
+        auto to_cartesian = [=] __device__ (auto lonlat) {
+            auto lon = thrust::get<0>(lonlat);
+            auto lat = thrust::get<1>(lonlat);
+            return thrust::make_pair(
+                lon_to_x(origin_lon - lon, midpoint(lat, origin_lat)),
```
cwharris, line 89: The reason I factored this differently from legacy is that `M_PI / 360` is a misnomer. It at first appears to be an incorrectly implemented degree => radian calculation. However, it can also be understood to be taking the midpoint between lat and origin_lat. I find this to be more intelligible than the previous implementation, but I am not sure _why_ it makes sense to take the midpoint of `lat` and `origin_lat`.

cwharris:
```
x = ((origin_lon - lon) / 360) * 40000.0 * cos((origin_lat + lat) * M_PI / 360);
```
vs
```
x = ((origin_lon - lon) / 360) * 40000.0 * cos(((origin_lat + lat) / 2) * (M_PI / 180));
```

cwharris: I'm also confused why we are using `origin - position` rather than `position - origin`. To get from A to B, we use `B - A`. Isn't that the question we're trying to answer here?

_"from the perspective of the the origin (A), where is the destination (B)?"_

Instead we're asking

_"from the perspective of the destination (B), where is the origin (A)?_

which gives us the opposite.

This might make sense if it were reducing a more complex equation, such as intentionally inverting x/y because lon/lat are opposite by nature, but as longitude increases, easting increases, and as latitude increases, northing increases. Wouldn't it then make sense that as longitude increases, X also increased? (An argument could be made for 3d graphics and left/right hand rule, but we are returning 2d coordinates).

harrism: The signs do seem inverted. As for the origin of the formula, the whole thing is unclear to me. Without knowing what map projection is being used, it will be difficult to figure out. I do note that calculating the length of a secant on a circle results in a half angle, which could account for the PI/360 instead of PI/180...

I think we need @zhangjianting to point us at the original code.

Issue #155, opened 2020-03-27
[BUG] Update trajectory.hpp to libcudf++


harrism:
* Eliminate use of gdf_column
* Remove all raw `RMM_ALLOC/RMM_FREE` and replace with `rmm::device_buffer / device_vector`.
* Add appropriate `CUSPATIAL_EXPECTS` calls. See #153

harrism:
Fixed by #171 and #201

PR #171, opened 2020-04-08
[REVIEW] Update trajectory.hpp to libcudf++

trxcllnt:
Second half of #155.

Notes:
* Renamed `trajectory_spatial_bounds` -> `trajectory_bounding_boxes`
* Didn't port `subset_trajectory_id`, use `cudf::experimental::gather` instead
* Changed `derive_trajectories` to return a pair of <table, column>. The column is a prefix-sum of offsets using `thrust::exclusive_scan`. This eliminates the need to return the `trajectory_ids` and `lengths` columns.


harrism:
> Second half of #155.

Where is the first half?

harrism:
I guess the problem with moving the old APIs in a separate PR is that we can't easily compare APIs. I think what we did in libcudf was to create the header for the new APIs in the same PR that moved the old APIs to the legacy folder. This way we could review the design changes to the API with full visibility of the old API it is replacing.

zhangjianting:
> > Second half of #155.
> 
> Where is the first half?

This is also my question. Where does id/timestamp/x/y get sorted?

trxcllnt:
@harrism yes, the first half was https://github.com/rapidsai/cuspatial/pull/166
@zhangjianting yes, the unit tests are running for both modules. I factored out the common test into a utility utility that creates a sorted test data table here: https://github.com/rapidsai/cuspatial/blob/80d9bbc551ddd57c1c41527d44834c551e17972a/cpp/tests/trajectory/trajectory_utilities.cuh#L63

You're right that sorting is necessary. My observation was that it's probably best to sort before calling these three APIs to avoid duplicating work.

If we used `cudf::sort_by_key()` in `derive_trajectories()` to replace the `thrust::stable_sort_by_key()`, we'd need to return both the sorted table, and the grouped `{id, length, offset}` table, since cudf doesn't sort in-place.

So the idea is the caller can sort up-front, then use the sorted columns and derived id, length, and offset together, like this:

```c++
// sort once up-front
auto sorted = cudf::experimental::sort_by_key(
      cudf::table_view{{id, ts, x, y}}, cudf::table_view{{id, ts}});

auto id_sorted = sorted->get_column(1);
auto ts_sorted = sorted->get_column(1);
auto xs_sorted = sorted->get_column(2);
auto ys_sorted = sorted->get_column(3);


// derive trajectories
auto trajectories = cuspatial::experimental::derive_trajectories(id_sorted);
auto lengths = trajectories->get_column(1);
auto offsets = trajectories->get_column(1);

auto velocities = cuspatial::experimental::compute_velocities(
  xs_sorted, ys_sorted, ts_sorted, lengths, offsets));

auto bounding_boxes = cuspatial::experimental::compute_bounding_boxes(
  xs_sorted, ys_sorted, lengths, offsets);
```

I ported the current tests here:
* [test_compute_velocities.cu](https://github.com/rapidsai/cuspatial/blob/80d9bbc551ddd57c1c41527d44834c551e17972a/cpp/tests/trajectory/test_compute_velocities.cu)
* [test_compute_bounding_boxes.cu](https://github.com/rapidsai/cuspatial/blob/80d9bbc551ddd57c1c41527d44834c551e17972a/cpp/tests/trajectory/test_compute_bounding_boxes.cu)

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @param[in] id column of object (e.g., vehicle) ids
+ * @param[in] mr The optional resource to use for all allocations
+ *
+ * @return a sorted table with the following three int32 columns:
+ *   * trajectory id - the unique ids from the input object ids column
+ *   * trajectory length - the number of objects in the derived trajectories
+ *   * trajectory offset - the cumulative sum of start positions for each group
+ */
+std::unique_ptr<cudf::experimental::table> derive_trajectories(
```
harrism, line 42: It's strange to return a table with just the ID, length and offset. The old API took the x, y, and timestamps and sorted them by the ID and timestamp.  I don't think that the gather required for this re-sorting should be moved outside of the library.

In fact, if this function only returns the *unique* IDs, then how can you use it to reorder the x/y/timestamp data?

I think it's probably best to produce the sorted data in this function. You could return a pair of tables: one with the trajectory ID/length/offset, and one with the sorted dataset. Also, ID is not used in any of the other APIs, once the unique IDs are computed. So is it necessary to return it? 

Length is not strictly needed either; if you make the ID/offset table one larger, then you can compute the lengths on the fly. If length is not needed outside of cuSpatial that would save some precious memory and bandwidth.

So it could be possible to just return a cudf::column of offsets and the sorted data, and that would be the input to both of the other trajectory functions.

cpp/src/trajectory/derive_trajectories.cu
```diff
@@ -0,0 +1,85 @@
...
+    cudf::column_view const& id, rmm::mr::device_memory_resource* mr,
+    cudaStream_t stream) {
+  using typename cudf::column;
+  using typename cudf::table_view;
+  using namespace cudf::experimental;
+
+  std::vector<groupby::aggregation_request> reqs{};
+  reqs.reserve(2);
+
+  // append count ids aggregation
```
harrism, line 45: Could this all be something like 

```
std::vector<groupby::aggregation_request> reqs{{{id}, {make_count_aggregation()},
                                               {{id}, {make_nth_element_aggregation(0)}};
```

?

(Or with more explicit types.)

cpp/src/trajectory/derive_trajectories.cu
```diff
@@ -0,0 +1,85 @@
...
+  reqs.push_back(groupby::aggregation_request{id});
+  reqs[0].aggregations.push_back(make_count_aggregation());
+
+  // append nth_element aggregation to force method=sort
+  reqs.push_back(groupby::aggregation_request{id});
+  reqs[1].aggregations.push_back(make_nth_element_aggregation(0));
+
+  // do the needful
```
harrism, line 53: // add a helpful...

... comment, please. :)

harrism:
@trxcllnt If you want to do the approach you describe above, then remove the groupby, since it is re-doing the sort of IDs. In that  case you would need to re-implement the non-sort part of groupby,  similar to the old implementation of derive_trajectories.

Alternatively, I think you could do the sort-based groupby and then a gather to reorder the x/y and timestamp data. You wouldn’t have to call sort_by_key on them all.  The scan  should already be done inside the groupby, just need to know how to get its results out.

I think it might be a more intuitive API to  return the column of offsets and the sorted data in a table.  My comments in the review talk about how the unique IDs and lengths are optional  as well.

cpp/src/trajectory/derive_trajectories.cu
trxcllnt: That's my preferred style as well, but the `make_*_aggregation()` methods return `unique_ptr<aggregation_requestion>`, which didn't play nice with the `initializer_list` construction no matter how I tried to `std::move()` things around. Is there some magic that makes them work together?

cpp/src/trajectory/derive_trajectories.cu
trxcllnt: Just having a bit of fun with one of my genuinely favorite phrases :smile: :
> [Do the needful](https://en.wikipedia.org/wiki/Indian_English#Vocabulary): To do that which is necessary or required, with the respectful implication that the other party is trusted to understand what needs doing without being given detailed instruction

cpp/src/trajectory/derive_trajectories.cu
```diff
@@ -0,0 +1,85 @@
...
+  // Append ids output column
+  cols.push_back(std::move(result.first->release()[0]));
+  // Append lengths output column
+  cols.push_back(std::move(result.second[0].results[0]));
+  // Append offsets output column. Use `allocate_like` to
+  // ensure the `lengths` null_mask is retained if exists
+  cols.push_back(cudf::experimental::allocate_like(
+      *cols.at(1), mask_allocation_policy::RETAIN, mr));
+
+  // cumulative sum to fill offsets
```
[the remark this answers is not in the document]
trxcllnt: Looks like the [`sort_groupby_helper` does](https://github.com/rapidsai/cudf/blob/54b50931a380fa11b4a158ebf894dabd9668c024/cpp/include/cudf/detail/groupby/sort_helper.hpp#L175), but the [`helper()` accessor](https://github.com/rapidsai/cudf/blob/54b50931a380fa11b4a158ebf894dabd9668c024/cpp/include/cudf/groupby.hpp#L217) is private on the `cudf::experimental::groupby::groupby` instance.

cpp/include/cuspatial/trajectory.hpp
trxcllnt: Upon closer inspection, I think we might be able to eliminate the `derive_trajectories` groupby altogether, and instead implement `compute_velocities` and `compute_bounding_boxes` via `thrust::reduce_by_key` on the unsorted data.

cpp/src/trajectory/derive_trajectories.cu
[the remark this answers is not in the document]
devavret: Although you can get the offsets publicly using [`groupby().get_groups()`](https://github.com/rapidsai/cudf/blob/3cd4c9f0602840dddb9a0e247d5a0bcf3d7266e1/cpp/include/cudf/groupby.hpp#L181-L195). It also returns a copy of sorted keys.

ci/gpu/build.sh
```diff
@@ -75,6 +75,19 @@ cd $WORKSPACE
...
     logger "Skipping tests..."
 else
+    logger "Check GPU usage..."
+    nvidia-smi
+
+    # TODO
+    # logger "GoogleTests..."
+    # cd $WORKSPACE/cpp/build
+
+    # for gt in ${WORKSPACE}/cpp/build/gtests/* ; do
```
ajschmidt8, line 85: hey @trxcllnt. Ops guy here. is there any reason these Google Tests were commented out?

trxcllnt: @ajschmidt8 ah shit, I forgot I'd done that locally. Currently CI isn't running the C++ GTests at all, so I copied this over from the cuDF script. Then I realized the C++ tests aren't currently passing for other reasons, and turning them on would fail CI for this PR, so they're commented out.

trxcllnt: @harrism do you know what we can do to fix the C++ tests? Currently the `CubicSplitTest` is failing with `RMM_ERROR_NOT_INITIALIZED`. If we can initialize RMM in those tests, I can enable C++ testing in CI here.

cpp/include/cuspatial/trajectory.hpp
trxcllnt: Yeah it seems those extra columns were redundant, which eliminates the need to use groupby. For that reason, and the fact we can't get the groupby offsets as a device buffer, I've changed it back to a thrust reduce_by_key over the sorted input ids.

ci/gpu/build.sh
thomcom: `CubicSplineTest` should be working so I'm digging into that presently.

thomcom: I resolved the issue with RMM_ERROR_NOT_INITIALIZED in https://github.com/rapidsai/cuspatial/pull/175. Is that the only test that was failing?

trxcllnt: @thomcom thanks! Yes I think so. Once that's merged, I'll merge it in and uncomment this here.

harrism: I think the correct thing to do is to change all non-legacy tests to not use rmmAlloc or rmmFree (use `rmm::device_buffer`, `rmm::device_vector`, and if necessary `device_memory_resource` classes).

cpp/src/trajectory/compute_distance_and_speed.cu
```diff
@@ -0,0 +1,191 @@
...
+  return cudf::experimental::type_dispatcher(x.type(), dispatch_element{}, x, y,
+                                             timestamp, offset, mr, stream);
+}
+}  // namespace detail
+
+std::unique_ptr<cudf::experimental::table> compute_distance_and_speed(
+    cudf::column_view const& x, cudf::column_view const& y,
+    cudf::column_view const& timestamp, cudf::column_view const& offset,
+    rmm::mr::device_memory_resource* mr) {
+  CUSPATIAL_EXPECTS(!(x.is_empty() || y.is_empty() || offset.is_empty() ||
```
harrism, line 171: Generally in cuDF for empty inputs we return empty outputs, and don't throw an exception. This is to support things like Dask that sometimes generates empty tasks.  I am not sure, but I think we should probably follow the same process here.

cpp/tests/trajectory/legacy/test_trajectory_derive.cu
```diff
@@ -101,12 +103,12 @@ TEST_F(TrajectoryDerive, BadData)
     gdf_column_view(&bad_timestamp, 0, 0, 0, GDF_TIMESTAMP);
 
     // null pointers
-    CUDF_EXPECT_THROW_MESSAGE(cuspatial::derive_trajectories(bad_x, bad_y,
-                                                             bad_in_id,
-                                                             bad_timestamp,
-                                                             out_id,
-                                                             out_len, 
-                                                             out_offset),
+    CUSPATIAL_EXPECT_THROW_MESSAGE(cuspatial::derive_trajectories(bad_x, bad_y,
```
harrism, line 106: I think just use CUSPATIAL_EXPECT_THROW.

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,95 @@
...
+ * table with the trajectory ids, the number of objects in each trajectory,
+ * and the offset position of the first object for each trajectory in the
+ * input object ids column.
+ *
+ * @param[in] x coordinates (km) (sorted by id, timestamp)
+ * @param[in] y coordinates (km) (sorted by id, timestamp)
+ * @param[in] object_id column of object (e.g., vehicle) ids
+ * @param[in] timestamp column (sorted by id, timestamp)
+ * @param[in] mr The optional resource to use for all allocations
+ *
```
harrism, line 39: Missing @throw for the exceptions this can throw. Same for the other functions in this header.

cpp/include/cuspatial/detail/trajectory.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @param[in] x coordinates (km) (sorted by id, timestamp)
+ * @param[in] y coordinates (km) (sorted by id, timestamp)
+ * @param[in] timestamp column (sorted by id, timestamp)
+ * @param[in] end position for each trajectory's last object, used to index
+ * timestamp/x/y columns (sorted by id, timestamp)
+ * @param[in] mr The optional resource to use for all allocations
+ * @param[in] stream Optional CUDA stream on which to schedule allocations
+ *
+ * @return a sorted cudf table of distances (meters) and speeds (meters/second)
```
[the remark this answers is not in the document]
trxcllnt: Sorting in the same order as the input trajectories. @zhangjianting will have to answer why output is meters and meters/second.

cpp/include/cuspatial/detail/trajectory.hpp
```diff
@@ -0,0 +1,102 @@
...
+ * @param[in] x coordinates (km) (sorted by id, timestamp)
+ * @param[in] y coordinates (km) (sorted by id, timestamp)
+ * @param[in] object_id column of object (e.g., vehicle) ids
+ * @param[in] timestamp column (sorted by id, timestamp)
+ * @param[in] mr The optional resource to use for all allocations
+ * @param[in] stream Optional CUDA stream on which to schedule allocations
+ *
+ * @return an `std::pair<table, column>`:
+ *  1. table of (object_id, timestamp, x, y) sorted by (object_id, timestamp)
+ *  2. int32 column of end positions for each trajectory's last object
```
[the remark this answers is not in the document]
trxcllnt: The offsets are the position of the last object in each trajectory because that's what the legacy code was doing (using thrust `inclusive_scan` instead of `exclusive_scan`):

```
point_ids:    [0, 0, 0, 1, 1, 2]
traj_offsets: [2, 4, 5] // (0 is the inferred starting offset)
```

Initially I'd changed it to zero-based offsets, but as you [noted earlier](https://github.com/rapidsai/cuspatial/pull/171#discussion_r405960154), using inclusive_scan allows us to avoid returning the additional `lengths` buffer.

But now I think changing this back to zero-based offsets is the right approach, since it would allow us to replace the custom trajectory kernels with `cub::DeviceSegmentedReduce`.

cpp/src/trajectory/compute_distance_and_speed.cu
```diff
@@ -0,0 +1,191 @@
...
+    cudf::mutable_column_device_view speed) {
+  auto tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid < offset.size()) {
+    auto const end = offset.element<int32_t>(tid) - 1;
+    auto const idx = tid == 0 ? 0 : offset.element<int32_t>(tid - 1);
+    auto const time_ms = simt::std::chrono::floor<cudf::timestamp_ms::duration>(
+        timestamp.element<Timestamp>(end) - timestamp.element<Timestamp>(idx));
+
+    if ((end - idx) < 2) {
+      speed.element<Element>(tid) = -2.0;
```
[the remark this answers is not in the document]
trxcllnt: No idea, it's just what the legacy kernel did. I assume some kind of sentinel used to indicate there was only one trajectory in the group?

cpp/include/cuspatial/detail/trajectory.hpp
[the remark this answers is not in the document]
zhangjianting: > Sorting in the same order as the input trajectories. @zhangjianting will have to answer why output is meters and meters/second.

It is kind of arbitrary, both seem to be most popular metrics for distance/speed.  Changing to other metrics has low overheads and is an element-wise op, which can be done in scripts on CPUs.

cpp/src/trajectory/compute_bounding_boxes.cu
```diff
@@ -0,0 +1,157 @@
...
+#include <cudf/utilities/type_dispatcher.hpp>
+#include <cuspatial/detail/trajectory.hpp>
+
+namespace cuspatial {
+namespace experimental {
+
+namespace {
+
+template <typename Element>
+__global__ void compute_bounding_boxes_kernel(
```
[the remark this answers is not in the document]
trxcllnt: This kernel finds the min/max x/y for each route in a trajectory (which is a group of routes). Assuming it's possible for cub to use thrust tuples/zip_iterators, we could do this with `cub::DeviceSegmentedReduce` using the offsets computed by `derive_trajectories`.

Alternatively we can change the API to accept the original `object_id` column instead of the offsets from `derive_trajectories`, and use `thrust::reduce_by_key`, which is equivalent to this in cudf:

```python
points_df = cudf.DataFrame({
  'id': [0, 0, 1, 1],
  'x':  [1, 2, 3, 0],
  'y':  [2, 3, 4, 1] })

trajectory_bboxes = points_df \
     .groupby(by='id', as_index=False, level=1) \
     .agg({ 'id': 'min', 'x': ['min', 'max'], 'y': ['min', 'max'] })

print(trajectory_bboxes)

#   id   x       y    
#      min max min max
# 0  0   1   2   2   3
# 1  1   0   3   1   4

```

cpp/include/cuspatial/detail/trajectory.hpp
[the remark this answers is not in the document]
harrism: RAPIDS aims to be end-to-end accelerated. Going back to the host to change units would be expensive and unnecessary. Shouldn't we be consistent regardless?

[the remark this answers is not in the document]
harrism: I think we should 

A.) Pick a default unit of distance. 
B.) Be consistent with that unit for all distances.
C.) Base velocities / speeds on that unit.
D.) Provide unit conversion APIs (or a guide on how to do it with cuDF).

cpp/src/trajectory/compute_bounding_boxes.cu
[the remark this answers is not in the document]
harrism: Hmmm, reduce_by_key is the most intuitive. If it allows us to do away with computing offsets altogether, I'm all for that.

cpp/src/trajectory/compute_distance_and_speed.cu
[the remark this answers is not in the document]
harrism: If there is only one point, distance and speed should be zero, shouldn't they?  FWIW, this is what nulls are for. :)  But you could also set the distance to 0 and speed to +Inf.

cpp/include/cuspatial/detail/trajectory.hpp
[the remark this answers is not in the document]
trxcllnt: I agree @harrism. Speed in kilometers/millisecond (as would be returned now without the scaling) only seems useful for trips across the galaxy at relativistic speeds... :stuck_out_tongue:

cpp/src/trajectory/compute_bounding_boxes.cu
```diff
@@ -0,0 +1,157 @@
...
+    cudf::column_device_view const x,
+    // Point Y
+    cudf::column_device_view const y,
+    // Offset for each trip in the group of trajectories
+    cudf::column_device_view const offset,
+    cudf::mutable_column_device_view bbox_x1,
+    cudf::mutable_column_device_view bbox_y1,
+    cudf::mutable_column_device_view bbox_x2,
+    cudf::mutable_column_device_view bbox_y2) {
+  auto tid = blockIdx.x * blockDim.x + threadIdx.x;
```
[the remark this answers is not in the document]
trxcllnt: Reimplemented with thrust.

cpp/include/cuspatial/detail/trajectory.hpp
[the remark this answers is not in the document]
zhangjianting: > I think we should
> 
> A.) Pick a default unit of distance.
> B.) Be consistent with that unit for all distances.
> C.) Base velocities / speeds on that unit.
> D.) Provide unit conversion APIs (or a guide on how to do it with cuDF).

I agree. I guess my point was that users know their data the best and they should know how to do the unit conversion, whether it is on CPU or GPU. I'm sure cuDF supports some sort of unary UDFs through JIT for this purpose where the unit conversion can be done on GPU. 
So, what should be the default for distance (meter/foot) and distance (m/s, km/h, MPH)?

cpp/src/trajectory/compute_distance_and_speed.cu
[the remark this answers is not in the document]
trxcllnt: Yes I agree. For now distance and speed for a single point will default to 0/0, but we can change this to 0/inf if we'd rather.

cpp/include/cuspatial/detail/trajectory.hpp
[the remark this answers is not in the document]
harrism: Definitely metric system. :)

Accuracy is a consideration. If the scale of distances tends to be large (100s of kms), we wouldn't want to use a small distance unit (meters) or truncation error and catastrophic cancellation can become problems. 

In practice we may find it makes sense to provide a unit specification or scale factor to some routines. *or* to have custom column wrappers that specify the scale factor for the data in the column.

cpp/tests/trajectory/legacy/test_trajectory_derive.cu
trxcllnt: Looks like cudf still has `CUDF_EXPECT_THROW_MESSAGE` in all its tests. Is the reason that we don't want to be explicitly testing error messages (i.e. in case we need to localize them later)?

cpp/src/trajectory/compute_bounding_boxes.cu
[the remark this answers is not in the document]
trxcllnt: The `thrust::reduce_by_key` re-implementations are definitely cleaner. Removing the offsets means we have to call thrust `unique_by_key()` to determine the size of the output columns. This seems like an acceptable cost if we're trying to save the memory allocated for the offsets.

On the other hand if we'd rather not repeat work, we could precompute the offsets. Doesn't thrust's reduce_by_key have to figure out offsets from the keys internally anyway? If so, we might be able to skip that too and call `cub::DeviceSegmentedReduce` directly.

cpp/tests/trajectory/legacy/test_trajectory_derive.cu
harrism: Hmmm, maybe I'm misremembering...

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * computes the minimum bounding box to contain all routes in each trajectory.
+ *
+ * @note Assumes object_id, timestamp, x, y presorted by (object_id, timestamp).
+ *
+ * @param[in] object_id column of object (e.g., vehicle) ids
+ * @param[in] x coordinates (in kilometers)
+ * @param[in] y coordinates (in kilometers)
+ * @param[in] mr The optional resource to use for all allocations
+ *
+ * @return a cudf table of bounding boxes with four columns:
```
harrism, line 85: Are these in km? Should probably say.

cpp/src/trajectory/trajectory_distances_and_speeds.cu
```diff
@@ -59,12 +59,13 @@ struct dispatch_timestamp {
     auto policy = rmm::exec_policy(stream);
 
     // Compute output column size
-    rmm::device_vector<int32_t> unique_keys(object_id.size());
-    auto last_key_pos =
-        thrust::unique_copy(policy->on(stream), object_id.begin<int32_t>(),
-                            object_id.end<int32_t>(), unique_keys.begin());
-    auto size = thrust::distance(unique_keys.begin(), last_key_pos);
-    unique_keys.~device_vector();
+    auto size = [&]() {
```
harrism, line 62: This is identical with the code in trajectory_bounding_boxes.cu -- make a function?

cpp/include/cuspatial/detail/trajectory.hpp
```diff
@@ -47,14 +51,31 @@ std::unique_ptr<cudf::experimental::table> trajectory_distances_and_speeds(
...
+ * @brief Count the number of unique object ids
+ *
+ * @param object_id Column of object (e.g., vehicle) ids
+ * @param stream Optional CUDA stream on which to schedule allocations
+ * @return cudf::size_type The number of unique elements
+ */
+inline cudf::size_type count_unique_ids(cudf::column_view const& object_id,
+                                        cudaStream_t stream = 0) {
+  auto policy = rmm::exec_policy(stream);
+  rmm::device_vector<int32_t> unique_keys(object_id.size());
```
harrism, line 72: Now that I see this, if the purpose is only to count unique ids, then this is quite inefficient in memory overhead and bandwidth. Instead, I would do a transform_reduce with one of the input iterators offset by one. The transform just returns 1 if the values are different, 0 if they are the same. The reduction operator is just sum.

harrism: If you prefer, we can file an issue to fix this in a separate PR.

harrism:
@trxcllnt I just noticed your size computation which counts the number of trajectories. This is actually unnecessary -- derive_trajectories returns a column whose `size()` is the number of trajectories. So doesn't it make sense for the `bounding_boxes` and `distances_and_speeds()` functions to take a `num_trajectories` parameter to save computing it again?

trxcllnt:
Yes we can pass the size in. I was anticipating we might remove the offsets buffer from `derive_trajectories` and coding defensively :sweat_smile:.

trxcllnt:
Seeing this error in the Ubuntu 18.04/CUDA 10.2 container:
```
[ 65%] Linking CUDA executable ../gtests/DERIVE_TRAJECTORIES_TEST

/usr/bin/ld: /conda/envs/gdf/lib/libcudftestutil.a(column_utilities.cu.o): relocation R_X86_64_32 against symbol `_ZN6thrust8cuda_cub3cub28DeviceReduceSingleTileKernelINS1_18DeviceReducePolicyINS_5tupleIblNS_9null_typeES5_S5_S5_S5_S5_S5_S5_EEiNS0_9__find_if7functorIS6_EEE9Policy600EPS6_SC_iS9_S6_EEvT0_T1_T2_T3_T4_' can not be used when making a PIE object; recompile with -fPIC

/usr/bin/ld: /conda/envs/gdf/lib/libcudftestutil.a(table_utilities.cu.o): relocation R_X86_64_32S against symbol `_ZTVSt9basic_iosIcSt11char_traitsIcEE@@GLIBCXX_3.4' can not be used when making a PIE object; recompile with -fPIC

/usr/bin/ld: final link failed: Nonrepresentable section on output

collect2: error: ld returned 1 exit status
```

We're already setting `POSITION_INDEPENDENT_CODE ON` for each test in CMake... is this saying we need to compile `cudftestutils.a` with `-fPIC`?

thomcom:
rerun tests

ci/gpu/build.sh
```diff
@@ -22,6 +22,7 @@ export PATH=/conda/bin:/usr/local/cuda/bin:$PATH
 export PARALLEL_LEVEL=4
 export CUDA_REL=${CUDA_VERSION%.*}
 export CUDF_HOME="${WORKSPACE}/cudf"
+export CUSPATIAL_HOME="${WORKSPACE}"
```
harrism, line 25: How can CUSPATIAL_HOME be the parent directory of CUDF_HOME?

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * Groups the input object ids to determine unique trajectories. Returns a
+ * table with the trajectory ids, the number of objects in each trajectory,
+ * and the offset position of the first object for each trajectory in the
+ * input object ids column.
+ *
+ * @param object_id column of object (e.g., vehicle) ids
+ * @param x coordinates (in kilometers)
+ * @param y coordinates (in kilometers)
+ * @param timestamp column of timestamps in any resolution
+ * @param mr The optional resource to use for all allocations
```
harrism, line 38:
```suggestion
 * @param mr The optional resource to use for output device memory allocations
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * trajectory.
+ *
+ * @note Assumes object_id, timestamp, x, y presorted by (object_id, timestamp).
+ *
+ * @param num_trajectories number of trajectories (unique object ids)
+ * @param object_id column of object (e.g., vehicle) ids
+ * @param x coordinates (in kilometers)
+ * @param y coordinates (in kilometers)
+ * @param timestamp column of timestamps in any resolution
+ * @param mr The optional resource to use for all allocations
```
harrism, line 71:
```suggestion
 * @param mr The optional resource to use for output device memory allocations
```

cpp/include/cuspatial/trajectory.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * @brief Compute the spatial bounding boxes of trajectories. Groups the x, y,
+ * and timestamp columns by object id to determine unique trajectories, then
+ * computes the minimum bounding box to contain all routes in each trajectory.
+ *
+ * @note Assumes object_id, timestamp, x, y presorted by (object_id, timestamp).
+ *
+ * @param object_id column of object (e.g., vehicle) ids
+ * @param x coordinates (in kilometers)
+ * @param y coordinates (in kilometers)
+ * @param mr The optional resource to use for all allocations
```
harrism, line 99:
```suggestion
 * @param mr The optional resource to use for output device memory allocations
```

ci/gpu/build.sh
trxcllnt: I agree it's odd, it happens a few lines below:
https://github.com/rapidsai/cuspatial/blob/2a14919e70b9dd4ff90d0ea1e95d7423ec1699db/ci/gpu/build.sh#L54-L60

trxcllnt:
> @harrism Any temporary allocations should use `rmm::mr::get_default_resource()` directly

That's good to know. I'm passing the `mr` into `cudf::detail::sort_by_key()`, but the sorted result table is part of the output, and all the other references are for output column allocations too.

trxcllnt:
rerun tests

PR #174, opened 2020-04-10
[REVIEW] Update hausdorff.hpp to libcudf++ 

cwharris:
Contributes to #160


cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,178 @@
...
+
+        for (int offset = blockDim.x / 2; offset > 0; offset >>= 1)
+        {
+            if (threadIdx.x < offset)
+            {
+                dist_sqrd_shared[threadIdx.x] = max(dist_sqrd_shared[threadIdx.x],
+                                                    dist_sqrd_shared[threadIdx.x + offset]);
+            }
+
+            __syncthreads(); // in a loop? yikes!
```
zhangjianting, line 99: This is needed when you do a reduction. The # of loops is log(blockDim.x) which is no more than 10.

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,178 @@
...
+            for (int i = a_begin; i < a_end; i++)
+            {
+                T ax = xs[i];
+                T ay = ys[i];
+                T dist_sqrd = magnitude_squared(bx - ax, by - ay);
+                min_dist_sqrd = min(min_dist_sqrd, dist_sqrd);
+            }
+        }
+
+        if (min_dist_sqrd > 1e10) // this may make some of the following statements uncessary.
```
zhangjianting, line 73: I think it would be safer to change it something below 1e+20.  sqrt(1e+10) may still be a valid distance.

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,178 @@
...
+        {
+            min_dist_sqrd = -1;
+        }
+
+        __shared__ T dist_sqrd_shared[1024];
+
+        dist_sqrd_shared[threadIdx.x] = -1;
+
+        __syncthreads(); // not sure if this is necessary
```
zhangjianting, line 82: You can move this to two lines above (right after if)

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,178 @@
...
+
+        if (threadIdx.x < max_threads) // not sure why this condition exists.
```
zhangjianting, line 84: we want to make sure that only threads that are assigned to compute a distance AND produce a valid distance can contribute to the max-reduction below.

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,178 @@
...
+{
+    CUSPATIAL_EXPECTS(x.type() == y.type(), "inputs `x` and `y` must have same type.");
+    CUSPATIAL_EXPECTS(x.size() == y.size(), "inputs `x` and `y` must have same length.");
+
+    CUSPATIAL_EXPECTS(not x.has_nulls() and
+                      not y.has_nulls() and
+                      not trajector_lengths.has_nulls(),
+                      "inputs must not have nulls.");
+
+    CUSPATIAL_EXPECTS(x.size() >= trajector_lengths.size(),
```
trevorsm7, line 169: I suppose this is an easy check that can fail early, but I think it should still check if the sum of all elements in `trajectory_length` is <= `x.size()` (or maybe it should be exactly equal)? Since you're already computing `exclusive_scan` of `trajectory_length` inside `hausdorff_functor()` anyway, maybe you could change it to an `inclusive_scan` so you can grab the sum as `d_trajectory_offsets.back()`. You can use `d_trajectory_offsets.begin() + 1` for the output iterator and manually set the first element to zero so it still looks like an `inclusive_scan`.

harrism: That's a device to host copy which requires synchronization.

cpp/src/spatial/hausdorff.cu
cwharris: When would `min_dist_sqrd` be an invalid distance? Does invalid mean "higher than we can calculate" ?

cpp/src/spatial/hausdorff.cu
```diff
@@ -26,89 +26,78 @@
...
 
-        if (threadIdx.x == 0)
+        if (threadIdx.x < offset)
         {
-            results[block_idx] = (dist_sqrd[0] < 0) ? 1e10 : sqrt(dist_sqrd[0]);
+            dist_sqrd[threadIdx.x] = max(dist_sqrd[threadIdx.x],
+                                            dist_sqrd[threadIdx.x + offset]);
         }
     }
+
```
harrism, line 96: You need a __syncthreads() after the loop before you read the shared memory from a different thread, since you moved it to the top.

cpp/src/spatial/hausdorff.cu
```diff
@@ -122,15 +111,15 @@ struct hausdorff_functor
 
     template<typename T>
     std::enable_if_t<std::is_floating_point<T>::value, std::unique_ptr<cudf::column>>
-    operator()(cudf::column_view const& x,
-               cudf::column_view const& y,
-               cudf::column_view const& trajectory_lengths,
+    operator()(cudf::column_view const& xs,
```
harrism, line 114: What is the `s` for?  What is a `space`? Are these not trajectories?

cpp/src/spatial/hausdorff.cu
```diff
@@ -140,35 +129,29 @@ struct hausdorff_functor
...
-        int block_x = result->size();
-        int block_y = 1;
-
-        if (result->size() > 65535)
-        {
-            block_y = ceil((float) result->size() / 65535.0);
-            block_x = 65535;
-        }
+        // utilize one block per result (pair of spaces).
+        int max_block_size = 65535;
```
harrism, line 140: I think the max CUDA thread block size is a lot smaller than this, isn't it?  Maybe this should be named "max_num_blocks"?

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,189 @@
+/*
+ * Copyright (c) 2019-2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you mpoint_a_y not use this file except in compliance with the License.
```
trevorsm7, line 5:
```suggestion
 * you may not use this file except in compliance with the License.
```
Find and replace of `a` and `b` went haywire here :rofl:

cpp/src/spatial/hausdorff.cu
cwharris: Read is on the same thread as the write: 0.

cpp/src/spatial/hausdorff.cu
```diff
@@ -26,89 +26,78 @@
...
+                                 T* results)
 {
     auto block_idx = blockIdx.y * gridDim.x + blockIdx.x;
-    auto num_space_pairs = num_spaces * num_spaces;
+    auto num_results = num_spaces * num_spaces;
 
-    // each block processes a single pair of spaces.
-    if (block_idx < num_space_pairs)
+    // each block processes a single result / pair of spaces
+    if (block_idx >= num_results)
```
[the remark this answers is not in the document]
cwharris: This approach reduces nesting and eliminates the cognitive overhead of a potential `else` block much further down. The original implementation launches one block per pair, so this check may be out of an abundance of caution, and not necessary. Is changing the algorithm in scope for this PR? What do you mean by strided loop? I did not see a clear approach to unwrap the inner loop here, since it is dependent on the length of two separate spaces (trajectories is a misnomer. Hausdorff distance compares two spaces, finite subsets of which may include trajectories, curves, mesh vertices, point clouds, etc).

cpp/src/spatial/hausdorff.cu
cwharris: Hausdorff distance is the maximum of the minimum distance between pairs of points in two spaces. Trajectories may have been the original use case for our implementation, but are misnomers in regard to the mathematical definition of the algorithm.

cpp/src/spatial/hausdorff.cu
cwharris: I found mixed information about this when investigating, but ultimately landed on this because it appeared elsewhere in cudf. I agree that “block_size” is misleading. “max_thread_count” makes more sense.

cpp/src/spatial/hausdorff.cu
cwharris: “xs” was just plural for “x”. It gave me the option to use “x” for a single value later in the function, but I ended io kot using it. In languages like Haskell, list-operation functions often use x and xs as the head (singular) and tail (plural) arguments. I’m happy to use “x” and “y” if thats preferable.

cwharris: I personally prefer xs and ys as an idiom, though, because for functions which will use x and y as variable names, it eliminates the naming conflict.

cpp/src/spatial/hausdorff.cu
cwharris: Read is on the same thread as the write, no?

cpp/src/spatial/hausdorff.cu
cwharris: YES. I missed `a_` as `a`, but thought I found all of the errors pretty quick so didn't redo it. Lesson learned. :roll_eyes: 

Thanks!

thomcom:
I have the impression that these PRs should include the Cython updates, as well as pointing the python tests to the new code instead of the legacy code. This looks like a nice implementation, thorough and well-understood, but we don't have any certainty that it matches the previous implementation because the python tests are not being validated.

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,189 @@
...
+                            rmm::mr::device_memory_resource *mr)
+{
+    CUSPATIAL_EXPECTS(xs.type() == ys.type(), "inputs `xs` and `ys` must have same type.");
+    CUSPATIAL_EXPECTS(xs.size() == ys.size(), "inputs `xs` and `ys` must have same length.");
+
+    CUSPATIAL_EXPECTS(not xs.has_nulls() and
+                      not ys.has_nulls() and
+                      not points_per_space.has_nulls(),
+                      "inputs must not have nulls.");
+
```
trevorsm7, line 179:
```suggestion
CUSPATIAL_EXPECTS(points_per_space.size() < 65536, "points_per_space must be less than 65536");
```
Or similar, otherwise `points_per_space.size()` squared will overflow `size_type`

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,189 @@
...
+    int space_b_begin = space_b_idx == 0 ? 0 : space_offsets[space_b_idx - 1];
+    int space_b_end   =                        space_offsets[space_b_idx];
+
+    T min_dist_sqrd = 1e20;
+
+    int num_points_in_b = space_b_end - space_b_begin;
+
+    if (threadIdx.x < num_points_in_b)
+    {
+        T point_b_x = xs[space_b_begin + threadIdx.x];
```
trevorsm7, line 64: These reads from `xs`/`ys` could potentially be out of bounds if `points_per_space` were malformed. The sum of `points_per_space` (`space_offsets[num_spaces]`) should be <= `xs.size()`, but this is never asserted.

Reading `space_offsets[num_spaces]` back to host memory before launching this kernel would require synchronization where there wouldn't be otherwise, so I suppose it would be a tradeoff between performance and being able to fail gracefully. Just mentioning it rather than requesting a change. :man_shrugging:

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,189 @@
...
+            min_dist_sqrd = min(min_dist_sqrd, dist_sqrd);
+        }
+    }
+
+    if (min_dist_sqrd > 1e10)
+    {
+        min_dist_sqrd = -1;
+    }
+
+    __shared__ T dist_sqrd[1024];
```
trevorsm7, line 82: The `1024` here is suppose to be the same as `NUM_THREADS` below, right? Maybe make it a `constexpr` global and use it here as well.

cpp/src/spatial/hausdorff.cu
trevorsm7: This is because in the final iteration of the loop `offset` will be 1, so only thread 0 will write to `dist_sqrd[0]`, followed by only thread 0 reading from `dist_sqrd[0]`, right?

cwharris:
@thomcom I asked a question in https://github.com/rapidsai/cuspatial/pull/189#issuecomment-623726715 without realize it was better placed here! Sorry! :smile:

cpp/src/spatial/hausdorff.cu
cwharris: Definitely worth mentioning. I honestly had thought of it. It would be nice if we had a way of communicating errors back to the host, but since we don't have exceptions, we'd have to reduce some flags or something :man_shrugging: . Ultimately it's up to the consumer to make sure they're providing valid data. We should document the fact that this specific case will cause a cuda error (which could be helpful for the consumer when determining the cause of the error).

cpp/src/spatial/hausdorff.cu
cwharris: That's my understanding.

cpp/src/spatial/hausdorff.cu
cwharris: Eventually I'd like to flatten out these loops so the block size is not a limitation, but that's out of scope for this PR.

cpp/src/spatial/hausdorff.cu
trevorsm7: Oh, actually this should be `< 46341` since `size_type` is signed too...
Would `constexpr auto too_big = size_type(ceil(sqrt(std::numeric_limits<size_type>::max())));` be too much of an atrocity?

cwharris: It'd be `floor`, thought right?

cwharris: I'll take a closer look to see what we can do.

trevorsm7: `floor` would give you the largest safe value, so you could use it with `<=`

cpp/src/spatial/hausdorff.cu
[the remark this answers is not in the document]
harrism: But it doesn't let you do a strided loop, which is the idiomatic (and usually correct) way to implement most CUDA kernels over variable size data...

cpp/src/spatial/hausdorff.cu
harrism: I guess this is all going to be replaced by a CUB call in the future anyway.

cpp/src/spatial/hausdorff.cu
harrism: Please avoid hard-coding an integer literal...  What is the source of 65536 here?

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * 
+ * https://en.wikipedia.org/wiki/Hausdorff_distance
+ *
+ * `points_per_space` is used to compute the offset of the first point in each space.
+ * 
+ * @param[in] xs: x coordinate of points in space
+ * @param[in] ys: y coordinate of points in space
+ * @param[in] points_per_space: number of points in each space
+ *
+ * @returns A flattened matrix of all Hausdorff distances for each pair of spaces
```
harrism, line 35: The order of the outputs needs explanation (and an example in this doc would help).

cpp/src/spatial/hausdorff.cu
cwharris: I wasn't sure, which is why I left it as a literal (hoping this conversation would happen - but forgetting to add a comment myself). It's related to the number of blocks we can have in a grid, I think? Not sure what to name it.

cpp/src/spatial/hausdorff.cu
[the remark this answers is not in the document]
cwharris: If we can flatten the loops and stride them independent of the length of the spaces, that would be ideal, but I think that's out of scope for this PR. I'll get the if statement changed back to the nested version.

cpp/src/spatial/hausdorff.cu
trevorsm7: > Please avoid hard-coding an integer literal... What is the source of 65536 here?

65536 * 65536 is the smallest square that would overflow a 32-bit unsigned int. But then I remembered that `size_type` is signed. That's why I suggested `size_type(ceil(sqrt(std::numeric_limits<size_type>::max())))` instead (which, thanks to `constexpr`, can be evaluated at compile time)

cpp/src/spatial/hausdorff.cu
harrism: That might be the maximum total number of threads, but when I check for the maximum block x-dimension on my Volta, I get 1024...  

```
mharris@canopus:~/rapids/cudf/cpp/build/release$ vim test.cu
mharris@canopus:~/rapids/cudf/cpp/build/release$ nvcc test.cu
mharris@canopus:~/rapids/cudf/cpp/build/release$ ./a.out 
Max x-dimension: 1024
mharris@canopus:~/rapids/cudf/cpp/build/release$ more test.cu
#include <stdio.h>

int main(void) {
    int max_x{0};
    cudaDeviceGetAttribute(&max_x, cudaDevAttrMaxBlockDimX, 0);
    printf("Max x-dimension: %d\n", max_x);
}
```

This really shouldn't be hard coded...

cpp/src/spatial/hausdorff.cu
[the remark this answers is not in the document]
cwharris: Leaving this here in case I somehow forget:

Flatten this algorithm in case I lose it:

```
// ============================================================================
// find Hausdorff distance between spaces A, B, and C
//  - distance is non-symetrical, so calculate both ways
//  - results[i][j] = hausdorff distance from space i to space j
//  - distance between a space and itself is always zero
//    thus the diagonal of the matrix will always be zeroes
//
A = [025]
B = [37]
C = [9]

This is a single-dimension version of the Hausdorff algorithm, but basically `|v2-v1|` can be replaced with `magnitude(ax-bx, ay-by)` for a 2d version... or I guess any number of dimensions?

// ============================================================================
// cartesian product of input spaces (ABC)
//
AA, AB, AC
BA, BB, BC
CA, CB, CC

// expand by the cartesian product of each pair of spaces
//
[025]x[025], [025]x[37], [025]x[9]
 [37]x[025],  [37]x[37],  [37]x[9]
  [9]x[025],   [9]x[37],   [9]x[9]

// group of spaces is indicated by the bars across the top and sides of the matrix ("---==" and "][").
//
//    ================  ==========  ====
//    ================  ==========  ====
// ][ (00), (02), (05), (03), (07), (09)
// ][
// ][ (20), (22), (25), (23), (27), (29)
// ][
// ][ (50), (52), (55), (53), (57), (59)
// 
// ][ (30), (32), (35), (33), (37), (39)
// ][
// ][ (70), (72), (75), (73), (77), (79)
//
// ][ (90), (92), (95), (93), (97), (99)

// ============================================================================
// find the minimum distance from each point in a space to any point in another space
// find the maximum distance of each resulting distance.
// NOTE: word this better.
//
//    --------===================   --------============   --------=====
//    --------===================   --------============   --------=====
// ][ max(min(|0-0|, |0-2|, |0-5|), max(min(|0-3|, |0-7|), max(min(|0-9|),
// ][     min(|2-0|, |2-2|, |2-5|),     min(|2-3|, |2-7|),     min(|2-9|),
// ][     min(|5-0|, |5-2|, |5-5|))     min(|5-3|, |5-7|))     min(|5-9|))
// 
// ][ max(min(|3-0|, |3-2|, |3-5|), max(min(|3-3|, |3-7|), max(min(|3-9|),
// ][     min(|7-0|, |7-2|, |7-5|))     min(|7-3|, |7-7|))     min(|7-9|))
// 
// ][ max(min(|9-0|, |9-2|, |9-5|)) max(min(|9-3|, |9-7|)) max(min(|9-9|))

//    ----=====  ----=====  ----=====
//    ----=====  ----=====  ----=====
// ][ max(|0-0|, max(|0-3|, max(|0-9|,
// ][     |2-2|,     |2-3|,     |2-9|,
// ][     |5-5|)     |5-3|)     |5-9|)
//
// ][ max(|3-2|, max(|3-3|, max(|3-9|,
// ][     |7-5|)     |7-7|)     |7-9|)
//
// ][ max(|9-5|) max(|9-7|) max(|9-9|)

//    ===== ===== =====
// ][ |0-0| |5-3| |0-9|
// ][ |7-5| |7-7| |3-9|
// ][ |9-5| |9-7| |9-9|

//    = = =
// ][ 0 2 9
// ][ 2 0 6
// ][ 4 2 0
```

Translating this to thrust could (hypothetically) be done by
1) compute cartesian product of the input points
2) convert each element in the product to a distance between said points.
3) reduce_by_key where the values are points, keys are the shape of the _first_ point in the cartesian pairs, and the reduce function is `min`.
4) reduce where the values are the min distances and the reduce function is `max`.

note: the lookup "shape of the _first_ point in the cartesian pairs" can be generated by scattering 1s in to an array length of the same length as the number of total points, where the scatter indices are the offsets of the groups. Then pass that through a scan.
`[4,2,3,1] => [0,0,0,0,1,1,2,2,2,3]`

As for the performance, I haven't calculated. This is a rough draft.

cpp/src/spatial/hausdorff.cu
cwharris: I'm leaving this out for the time being, because I do not have a / do not know how to verify this. Instead, I'm documenting what will happen in the event too many spaces are passed to the algorithm.

cwharris: Ended up adding a constant, commenting it, and using it as a guard: `points_per_space.size() > MAX_NUM_SPACES`.

cpp/src/spatial/hausdorff.cu
cwharris: Added a comment for `points_per_space` containing negative values (Which should account for for the `xs[too_low]` case). Added a an EXPECTS for sum of space sizes <= number of points. That should account for the `xs[too_high]` case.

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        auto space_offsets = rmm::device_vector<cudf::size_type>(points_per_space.size());
+
+        thrust::inclusive_scan(rmm::exec_policy(stream)->on(stream),
+                               points_per_space.begin<cudf::size_type>(),
+                               points_per_space.end<cudf::size_type>(),
+                               space_offsets.begin());
+
+        CUSPATIAL_EXPECTS(space_offsets.back() <= xs.size(),
```
trevorsm7, line 142: I'm happy to see input validation, but I think Mark won't like the `.back()` here since we determined that this relies on device synchronization. I'm not sure how to get around this other than using `cudaMemcpyAsync()` and `cudaStreamSync()` directly though.

cpp/src/spatial/hausdorff.cu
```diff
@@ -0,0 +1,200 @@
...
+#include <cudf/column/column.hpp>
+#include <cudf/column/column_device_view.cuh>
+#include <cudf/column/column_factories.hpp>
+#include <cudf/column/column_view.hpp>
+#include <cuspatial/error.hpp>
+#include <cudf/utilities/error.hpp>
+
+namespace {
+
+constexpr cudf::size_type MAX_NUM_SPACES = 46340; // floor(sqrt(numeric_limits<size_type>::max()))
```
trevorsm7, line 30: Why not use the expression and leave the compiler to sort out the value? Floor, sqrt, and max should all be constexpr. That way it will still work if we change `size_type` to 64-bit eventually.

cwharris: They are not constexpr.

cpp/src/spatial/hausdorff.cu
cwharris: Ah, yes, I wrote this at 3am. It looked like a good idea at the time. :laughing:

harrism: This is not the only use of back() in RAPIDS...  Can't fix it nicely until device_uvector is done.

cpp/src/spatial/hausdorff.cu
trevorsm7: They should be, but it looks like only in gcc6 and later so it won't pass our CI :frowning_face: 
https://wandbox.org/permlink/nMwncmdCMWRpOD8Z

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * Example in 1D (this function operates in 2D):
+ * ```
+ * spaces
+ * [0 2 5] [9] [3 7]
+ *
+ * spaces represented as points per space and concatenation of all points
+ * [0 2 5 9 3 7] [3 1 2]
+ *
+ * cartesian product of all
+ * points visually seperated
```
harrism, line 38: Suggest installing a spell checker in your IDE.

```suggestion
 * points visually separated
```

cpp/include/cuspatial/hausdorff.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * by pair of spaces            distance between points
```
harrism, line 39: "visually separated by pair of spaces" is confusing. Do you mean the character "space"? The table on the right has two spaces separating each entry -- is that the visual separation you mean?

cwharris: What about “visually separated to highlight the relationship of a pair of points with the pair of spaces from which it was produced” ?

Wordy, but maybe less ambiguous?

cwharris:
@harrism I had intended to do the cython changes in a followup PR. Is this no longer the practice we want to follow?

cpp/include/cuspatial/hausdorff.hpp
cwharris: Maybe labeling the sets and using those labels at the top and sides of the cells could make sense?

cwharris:
```
 * note: the following matricies are visually separated to highlight the relationship of a pair of
 * points with the pair of spaces from which it is produced
 * 
 * cartesian product of all
 * points by pair of spaces     distance between points
 * +----------+----+-------+    +---------+---+------+
```

cpp/include/cuspatial/hausdorff.hpp
cwharris: But who checks the spellchecker?

I need to check mine. It apparently doesn't look at comments. :(

harrism: I use "code spell checker" for vscode.

harrism:
> @harrism I had intended to do the cython changes in a followup PR. Is this no longer the practice we want to follow?

Up to you!

PR #175, opened 2020-04-14
[REVIEW] Remove use of RMM libraries and use cudf test utils correctly.

thomcom:
Something in the RMM API changed such that `rmmIntialize(nullptr)` needs to be called before `RMM_TRY`. This is now handled in `cubic_spline_test.cpp`.

### DIFF
--- a/cpp/tests/interpolate/cubic_spline_test.cpp
+++ b/cpp/tests/interpolate/cubic_spline_test.cpp
@@ -14,204 +14,105 @@
  * limitations under the License.
  */
 
-#include <time.h>
 #include <sys/time.h>
-#include <string>
+#include <time.h>
 
-#include <tests/utilities/base_fixture.hpp>
-#include <tests/utilities/column_utilities.hpp>
 #include <cudf/column/column_view.hpp>
-#include <cudf/table/table_view.hpp>
 #include <cudf/table/table.hpp>
+#include <cudf/table/table_view.hpp>
 #include <cudf/utilities/type_dispatcher.hpp>
 #include <cuspatial/cubic_spline.hpp>
 #include <cuspatial/error.hpp>
+#include <string>
+#include <tests/utilities/base_fixture.hpp>
+#include <tests/utilities/column_utilities.hpp>
+#include <tests/utilities/column_wrapper.hpp>
 
 struct CubicSplineTest : public cudf::test::BaseFixture {};
 
-template<typename T>
-auto make_device_column(T* const points, int length) {
-    T *d_p = nullptr;
-    RMM_TRY( RMM_ALLOC( &d_p, length * sizeof(T), 0));
-    assert(d_p != nullptr);    
-    CUDA_TRY( cudaMemcpy( d_p, points, length * sizeof(T), cudaMemcpyHostToDevice ) );
-    cudf::column_view col(cudf::data_type{cudf::experimental::type_to_id<T>()}, length, d_p);
-    cudf::column result(col);
-    RMM_FREE(d_p, 0);
-    return result;
-}
-
 auto get_d_expect() {
-  std::array<float, 4> d3_expect{{0.5, -0.5, -0.5, 0.5}};
-  std::array<float, 4> d2_expect{{0, 3, 3, -6}};
-  std::array<float, 4> d1_expect{{-1.5, -4.5, -4.5, 22.5}};
-  std::array<float, 4> d0_expect{{3, 4, 4, -23}};
-  std::vector<std::array<float, 4>> d_expect;
-  d_expect.push_back(d3_expect);
-  d_expect.push_back(d2_expect);
-  d_expect.push_back(d1_expect);
-  d_expect.push_back(d0_expect);
-  return d_expect;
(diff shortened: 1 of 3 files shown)


trxcllnt:
Verified this works locally. @thomcom What happens if you run one of these tests individually, without first running `test_coefficients_single`. Does this need to be called in each test/need to have a corresponding `rmmFinalize()` call?

Or since this is a legacy test, it doesn't really matter?

harrism:
Since rmmInitialize/finalize/alloc/free are now deprecated, we should instead move to using rmm::device_buffer (etc) and RMM memory_resource classes, as libcudf has done.

trxcllnt:
@harrism, I updated the test to use the cudf test utils.

cpp/tests/interpolate/cubic_spline_test.cpp
```diff
@@ -14,202 +14,105 @@
...
+      {3, 2, 3, 4, 3, 3, 2, 3, 4, 3, 3, 2, 3, 4, 3}};
+  cudf::test::fixed_width_column_wrapper<int> ids_column{{0, 0, 1, 2}};
+  cudf::test::fixed_width_column_wrapper<int> prefix_column{{0, 5, 10, 15}};
+
+  auto splines = cuspatial::cubicspline_coefficients(t_column, y_column,
+                                                     ids_column, prefix_column);
+
+  auto d_expect = get_d_expect();
+  for (unsigned int i = 0; i < d_expect.size(); ++i) {
+    auto h_expected_col = d_expect[i];
```
harrism, line 72: I think all the h_ and d_ are unnecessary, e.g.

```suggestion
    auto expected_col = this->reference_spline[i];
```

cpp/tests/interpolate/cubic_spline_test.cpp
```diff
@@ -14,202 +14,105 @@
...
-          device_column.size() * sizeof(float),
-          cudaMemcpyDeviceToHost);
-
-    for(int i = 0 ; i < device_column.size() ; ++i){
-      EXPECT_EQ(x[i], host_data[i]);
-    }
+  auto splines = cuspatial::cubicspline_coefficients(t_column, x_column,
+                                                     ids_column, prefix_column);
+
+  auto interpolates = cuspatial::cubicspline_interpolate(
```
harrism, line 91: I think `interpolants` is the word.

```suggestion
  auto interpolants = cuspatial::cubicspline_interpolate(
```

cpp/tests/interpolate/cubic_spline_test.cpp
```diff
@@ -14,202 +14,105 @@
...
+      {3, 2, 3, 4, 3, 3, 2, 3, 4, 3, 3, 2, 3, 4, 3}};
+  cudf::test::fixed_width_column_wrapper<int> ids_column{{0, 0, 1, 2}};
+  cudf::test::fixed_width_column_wrapper<int> prefix_column{{0, 5, 10, 15}};
+  cudf::test::fixed_width_column_wrapper<int> point_ids_column{
+      {0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2}};
+
+  auto splines = cuspatial::cubicspline_coefficients(t_column, x_column,
+                                                     ids_column, prefix_column);
+
+  auto interpolates = cuspatial::cubicspline_interpolate(
```
harrism, line 112: interpolants

cpp/tests/interpolate/cubic_spline_test.cpp
```diff
@@ -14,202 +14,105 @@
...
+TEST_F(CubicSplineTest, test_coefficients_single) {
+  cudf::test::fixed_width_column_wrapper<float> t_column{{0, 1, 2, 3, 4}};
+  cudf::test::fixed_width_column_wrapper<float> y_column{{3, 2, 3, 4, 3}};
+  cudf::test::fixed_width_column_wrapper<int> ids_column{{0, 0}};
+  cudf::test::fixed_width_column_wrapper<int> prefix_column{{0, 5}};
+
+  auto splines = cuspatial::cubicspline_coefficients(t_column, y_column,
+                                                     ids_column, prefix_column);
+
+  auto d_expect = get_d_expect();
```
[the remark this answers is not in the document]
thomcom: Yeah, `d_` is referring to implementation details - `D0...` are the math names of the coefficient columns. Doesn't need to appear here.

cpp/tests/interpolate/cubic_spline_test.cpp
thomcom: This is a really accurate catch - in this case `h_` refers to host, `d_` refers to implementation naming, so I totally agree.

cpp/tests/interpolate/cubic_spline_test.cpp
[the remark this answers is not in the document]
trxcllnt: :+1: Updated to inline the expected values, rename `d0_expect` to `detail0_expected`, and replace the loops with `expect_tables_equivalent()`.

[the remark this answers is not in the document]
harrism: Not really what I meant in either case, but I've had enough. :)

Issue #168, opened 2020-04-07
[BUG] Add libcuspatial GoogleBenchmarks CMake scaffolding


trxcllnt:
Need to add GoogleBenchmarks CMakeLists.txt for libcuspatial so we can capture current + updated benchmark results.

PR #183, opened 2020-04-29
[REVIEW] libcuspatial benchmark scaffolding

cwharris:
#168

Basic copy-pasta and rejiggering from cudf/rmm benchmark scaffolding.

May want to use cudf's `benchmark_fixture.hpp` and rmm's `synchonrization.hpp` directly, rather than maintain our own copy?


GPUtester:

harrism:
> May want to use cudf's `benchmark_fixture.hpp` and rmm's `synchonrization.hpp` directly, rather than maintain our own copy?

Why not get both from the same repo?

cpp/benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+
+project(CUSPATIAL_BENCHS LANGUAGES C CXX CUDA)
```
harrism, line 3:
```suggestion
project(CUSPATIAL_BENCHMARKS LANGUAGES C CXX CUDA)
```

(or at least spell BENCHES correctly. :) )

cpp/benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+find_library(CUDFTESTUTIL_LIBRARY "libcudftestutil.a"
+             HINTS "$ENV{CUDF_ROOT}/lib")
+
+message(STATUS "CUDFTESTUTIL: CUDFTESTUTIL_LIBRARY set to ${CUDFTESTUTIL_LIBRARY}")
+message(STATUS "CUDFTESTUTIL: CUDF_TEST_INCLUDE set to ${CUDF_TEST_INCLUDE}")
+
+add_library(cudftestutil STATIC IMPORTED ${CUDFTESTUTIL_LIBRARY})
+if (CUDF_TEST_INCLUDE AND CUDFTESTUTIL_LIBRARY)
+    set_target_properties(cudftestutil PROPERTIES IMPORTED_LOCATION ${CUDFTESTUTIL_LIBRARY})
```
harrism, line 20: Might need position independent code option here (@trxcllnt  did that work for you?)

trxcllnt: It did not. I couldn't replicate with gcc 7.5 in rapids-compose, but I think CI might be using gcc 7.3.

cpp/benchmarks/CMakeLists.txt
cwharris: Consistency! :laughing: 

https://github.com/rapidsai/cudf/blob/1046eed3b8a59792fa6ed5f66a34cc5b792ef929/cpp/benchmarks/CMakeLists.txt#L3

cpp/benchmarks/CMakeLists.txt
cwharris: Going to keep this as-is until we can determine what change is needed, if any.

Issue #158, opened 2020-03-27
[BUG] Update point_in_polygon.hpp to libcudf++


harrism:
* Eliminate use of gdf_column in shapefile APIs.
* Remove all raw RMM_ALLOC/RMM_FREE and replace with rmm::device_buffer / device_vector.
* Add appropriate CUSPATIAL_EXPECTS calls. See #153

cwharris:
Related resources:
- https://www.esri.com/library/whitepapers/pdfs/shapefile.pdf
- https://wrf.ecse.rpi.edu//Research/Short_Notes/pnpoly.html
- https://postgis.net/docs/ST_IsValid.html

harrism:
Fixed by #192 and #203

PR #192, opened 2020-05-03
[REVIEW] Update point_in_polygon.hpp to libcudf++

cwharris:
contributes to #158


GPUtester:

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <cudf/column/column_view.hpp>
+#include "cudf/types.hpp"
+
+namespace cuspatial {
+
+/**
+ * @brief Point-in-Polygon (PIP) tests between a column of points and a
+ *        column of polygons
+ *
+ * Performs collision detection for each test point against at most 32 polygons. Polygons are
```
harrism, line 30: This is not collision detection.
```suggestion
 * Tests whether points are inside at most 32 polygons. Polygons are
```

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <memory>
+#include <cudf/column/column.hpp>
+#include <cudf/column/column_view.hpp>
+#include "cudf/types.hpp"
+
+namespace cuspatial {
+
+/**
+ * @brief Point-in-Polygon (PIP) tests between a column of points and a
```
harrism, line 27:
```suggestion
 * @brief Tests whether the specified points are inside any of the specified polygons.
```

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+ *        column of polygons
+ *
+ * Performs collision detection for each test point against at most 32 polygons. Polygons are
+ * a collection of one or more rings. Rings are a collection of three or more vertices.
+ *
+ * @param[in] test_points_x:     x component of target points
```
harrism, line 33:
```suggestion
 * @param[in] test_points_x:     x-coordinates of points to test
```

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+ * @param[in] test_points_y:     y component of target points
```
harrism, line 34:
```suggestion
 * @param[in] test_points_y:     y-coordinates of points to test
```

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+ * @param[in] poly_offsets:      beginning index of the first ring in each polygon
+ * @param[in] poly_ring_offsets: beginning index of the first point in each ring
+ * @param[in] poly_points_x:     x component of polygon points
```
harrism, line 37:
```suggestion
 * @param[in] poly_points_x:     x-coordinates of polygon points
```

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+ * @param[in] poly_points_y:     y component of polygon points
```
harrism, line 38:
```suggestion
 * @param[in] poly_points_y:     y-coordinates of polygon points
```

cpp/include/cuspatial/point_in_polygon.hpp
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @returns A column of cudf::size_type containing one element per input point. Each bit
+ * represents a hit or miss for each of the input polygons in least-significant-bit order.
+ * i.e. `output[3] & 0x0010` indicates a hit or miss for the 3rd point against the 2nd polygon.
```
harrism, line 42: 0x0010 is the hexadecimal number 16, so that represents a hit against the 17th polygon, if bit 0 is the first polygon.

```suggestion
 * i.e. `output[3] & 0x0010` indicates a hit or miss for the 3rd point against the 17th polygon.
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+
+            // for each line segment
+            for (cudf::size_type point_idx = ring_begin; point_idx < ring_end - 1; point_idx++)
+            {
+                T ax = poly_points_x[point_idx];
+                T ay = poly_points_y[point_idx];
+                T bx = poly_points_x[point_idx + 1];
+                T by = poly_points_y[point_idx + 1];
+
+                auto y_between_ay_by = ay <= y && y < by; // is y in range [ay, by) when ay < by?
```
harrism, line 79: Auto is not helpful to me or the compiler here.

```suggestion
                bool y_between_ay_by = ay <= y && y < by; // is y in range [ay, by) when ay < by?
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                auto y_between_by_ay = by <= y && y < ay; // is y in range [by, ay) when by < ay?
```
harrism, line 80:
```suggestion
                bool y_between_by_ay = by <= y && y < ay; // is y in range [by, ay) when by < ay?
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                auto y_in_bounds = y_between_ay_by || y_between_by_ay; // is y in range [by, ay]?
```
harrism, line 81:
```suggestion
                bool y_in_bounds = y_between_ay_by || y_between_by_ay; // is y in range [by, ay]?
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                auto run  = bx - ax;
```
harrism, line 82:
```suggestion
                T run  = bx - ax;
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                auto rise = by - ay;
```
harrism, line 83:
```suggestion
                T rise = by - ay;
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                auto rise_to_point =  y - ay;
```
harrism, line 84:
```suggestion
                T rise_to_point =  y - ay;
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                                        const T* const __restrict__ poly_points_y,
+                                        cudf::size_type* const __restrict__ result)
+{
+    cudf::size_type idx = blockIdx.x * blockDim.x + threadIdx.x;
+
+    if (idx > num_test_points) {
+        return;
+    }
+
+    cudf::size_type hit_mask = 0;
```
harrism, line 53: cudf::size_type should be used to store things that are limited by the size of a cuDF column. It shouldn't be used just for an integer that needs to be limited to 32 bits. It could change in the future, is this function robust to that?

Note that cuDF doesn't use size_type for bitmasks, which this is. Suggest using an unsigned type.

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+struct point_in_polygon_functor
+{
+    template <typename T>
+    static constexpr bool is_supported()
+    {
+        return std::is_floating_point<T>::value;
+    }
+
+    template <typename T, std::enable_if_t< !is_supported<T>() >* = nullptr, typename... Args>
+    std::unique_ptr<cudf::column> operator()(Args&& ...)
```
harrism, line 107: This is a nice way to not repeat all the args in the unsupported path, I'll have to use this idea.

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+        auto results = cudf::make_fixed_width_column(type,
+                                                     size,
+                                                     cudf::mask_state::UNALLOCATED,
+                                                     stream,
+                                                     mr);
+
+        if (results->size() == 0)
+        {
+            return results;
+        }
```
harrism, line 135: cudf copying utils can simplify this a lot.
```suggestion
        if (test_points_y.is_empty()) return cudf::experimental::empty_like(test_points_y);
        auto results = cudf::expermental::detail::allocate_like(test_points_y, mask_allocation_policy::RETAIN, mr, stream);
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                 cudf::column_view const& test_points_y,
+                 cudf::column_view const& poly_offsets,
+                 cudf::column_view const& poly_ring_offsets,
+                 cudf::column_view const& poly_points_y,
+                 cudf::column_view const& poly_points_x,
+                 rmm::mr::device_memory_resource* mr)
+{
+    CUSPATIAL_EXPECTS(test_points_x.size() == test_points_x.size() and
+                      poly_points_x.size() == poly_points_y.size(),
+                      "All points must have both x and y values");
```
harrism, line 199: Typo (x == x) -- must not be testing this expectation. Also I don't think separate expectations should be combined, makes testing harder.

```suggestion
    CUSPATIAL_EXPECTS(test_points_x.size() == test_points_y.size(),
                      "Size mismatch between x and y test points");
    CUSPATIAL_EXPECTS(poly_points_x.size() == poly_points_y.size(),
                      "Size mismatch between x and y polygon points");
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+
+    CUSPATIAL_EXPECTS(test_points_y.type() == test_points_x.type() and
+                      poly_points_x.type() == test_points_x.type() and
+                      poly_points_y.type() == test_points_x.type(),
+                      "All points much have the same type for both x and y");
```
harrism, line 204:
```suggestion
    CUSPATIAL_EXPECTS(test_points_x.type() == test_points_y.type(), "Type mismatch in test points x / y");
    CUSPATIAL_EXPECTS(poly_points_x.type() == test_points_y.type(), "Type mismatch in polygon points x / y");
    CUSPATIAL_EXPECTS(poly_points_x.type() == test_points_x.type(), "Type mismatch between test points and polygon points");
```

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -0,0 +1,233 @@
...
+                      not test_points_y.has_nulls(),
+                      "Test points must not contain nulls");
+
+    CUSPATIAL_EXPECTS(not poly_points_y.has_nulls() &&
+                      not poly_points_x.has_nulls(),
+                      "Polygon points must not contain nulls");
+
+    CUSPATIAL_EXPECTS(poly_offsets.size() <= (cudf::size_type) sizeof(cudf::size_type) * 8,
+                      "Number of polygons cannot exceed bitmap capacity (32 for cudf::size_type)");
+
```
harrism, line 216: Again, I don't think size_type is appropriate for a bitmask. Notably, your size calculation should be replaced with `std::numeric_limits<mask_type>::digits`, and if you use a signed int type you will get 31 instead of 32.

cpp/include/cuspatial/point_in_polygon.hpp
cwharris: Good catch. `0x0010` should have been `0b0010`.

cpp/src/spatial/point_in_polygon.cu
cwharris: I started with `uint32_t`, but found out `cudf::column` doesn't support unsigned types... ? `cudf::bitmask_type` is `uint32_t`, but that's stored as a `device_buffer` and can be materialized to a `bool` column.

Do we support unsigned columns, and I'm just missing something?

cpp/src/spatial/point_in_polygon.cu
cwharris: Can't take credit for that one!

https://github.com/rapidsai/cudf/pull/3402#discussion_r358425449

cpp/src/spatial/point_in_polygon.cu
cwharris: Unfortunately `empty_like` and `allocate_like` can't be used here because the result type differs from the input type.

cpp/src/spatial/point_in_polygon.cu
cwharris: Afaik the signedness of the type does not matter because we're doing bit manipulation, and we're limited to signed types for cudf columns. I've changed this from `size_type` to `int32_t`. The consumer should not be inspecting the value as if it were a number, but as a per-point hit-mask. Please correct me if an int32 cannot represent a 32 bit mask - I have no experience doing this in c++.

I will add test cases for 31, 32, and 33 poly inputs, depending on which can be supported

cwharris: Am, I'm seeing how _at the very least_ this is highly inconvenient. :disappointed:

cwharris: Going with `int32_t` and supporting only 31 polygons for now.

PR #195, opened 2020-05-04
[REVIEW] Fix point_in_polygon docs

### LINKED ISSUE #187: [BUG] point_in_polygon_bitmap python docs are wrong
**Describe the bug**
The docs for this fn incorrectly explain what the input parameters are.

I'll update the docs with the correct explanation and tests, too.

thomcom:
The `point_in_polygon_bitmap` docs incorrectly explained the use of the `fpos` and `rpos` columns. This push rectifies that, and improves test coverage.

Fixes #187


GPUtester:

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -126,9 +126,8 @@ def point_in_polygon_bitmap(
     params
     x_points: x coordinates of points to test
     y_points: y coordinates of points to test
-    polygon_ids: a unique integer id for each polygon
-    polygon_end_indices: the (n+1)th vertex of the final coordinate of each
-                         polygon in the next parameters
+    polygon_fpos: the (n+1)th ring coordinate for each feature/polygon.
```
cwharris, line 129: Are `fpos` and `rpos` standard names? While porting, I found these confusing. Can we name them something like `poly_ring_offsets` `poly_ring_points_offsets`?

harrism: Agree, names that explain themselves are good.

zhangjianting: fpos and rpos are just two vector variables to index rings and vertices, respectively. They are not standard names. 
I think ring_offsets and vertex_offsets are good names.  Personally I do not like very long variable names.  
Ideally, when nested columns are supported, these variables will be internal to a polygon column and end users do not need to deal with them directly.

harrism: But there is no "f" in ring, and "r" surely does not correspond to the "r" in the middle of "vertices".

Short or long, names need to be meaningful to humans. :)

cwharris: After rummaging through the esri file format documentation, and various geospatial resources on the web, and porting the shapefile reader, I now understand where `fpos` and `rpos` are coming from.

`fpos` is the "feature" positions (a feature is some esoteric idea of a "thing" in space, especially on earth - they generally have shapes and custom attributes, like say polygons made of rings with a "danger zone" attribute). So `fpos` is basically `feature_position`, where "position" means "offset". `rpos` is the "ring" positions. Similar to `fpos`, but indicates the position or length of each ring in a feature (read: polygon). A feature can have one or more geometric entities associated with it. Right now we are only dealing with polygons.

`fpos` lets us determine how many rings are in a polygon and/or where the first ring in a given polygon begins.
`rpos` lets us determine how many vertices are in a ring and or where the first vertex in a given ring begins.

cwharris: Suggested name changes:
 - `poly_offsets`
 - `poly_ring_offsets`

I'm favoring "poly" over "feature" in this case because that's the only geometry this function supports right now.

cwharris: Also, I'm happy to go adjust these as I make my way thought the port.

harrism: I like it.

zhangjianting: > But there is no "f" in ring, and "r" surely does not correspond to the "r" in the middle of "vertices".
> 
> Short or long, names need to be meaningful to humans. :)

fpos indexes rings and rpos indexes vertices in the original design.

cwharris: > fpos indexes rings and rpos indexes vertices in the original design

That description assumes the reader understands the implementation details of the algorithm. I interpreted a description similar to this as “fpos is a lookup table for an ordered list of non-unique rings.” Which is clearly incorrect _after_ you know everything else about the algorithm. But consumers (and other developers) don’t necessarily know. A good description should be complete, concise, correct, and unambiguous. Here’s my attempt to summarize my understand:

```Polygons are comprised of 1 or more rings.
Rings are comprised of 3 or more vertices.
Rings looping clockwise represent positive space (presence).
Rings looping anti-clockwise represent negative space (absence).

`poly_offsets` is a list containing the index of the first ring of each polygon in ascending order. The total size of the first polygon is `poly_offsets[0]`. The total size of the remaining polygons are given by `poly_offsets[i] - poly_offsets[i - 1]`

`poly_ring_offsets` is a list containing the index of the first vertex of each ring in ascending order. The total size of the first ring is `poly_ring_offsets[0]`. The total size of the remaining rings are given by `poly_ring_offsets[i] - poly_ring_offsets[i - 1]`
```

PR #197, opened 2020-05-08
Move spatial_window_points/query.hpp to legacy

harrism:
This is the first step in #157 , simply moving the old files to legacy folders.


GPUtester:

trxcllnt:
Looks like CI is failing because this line in the Cython needs to change to `legacy/query.hpp`:
https://github.com/rapidsai/cuspatial/blob/55311bf5b42bb5c0ef454299d3dd5c16f38c81e6/python/cuspatial/cuspatial/_lib/spatial.pxd#L31

harrism:
> How come the cu files don't show up as a move?

No idea. Probably moved it in vscode rather than git. Sorry.

cwharris:
@harrism I move all my files in vscode by drag-and-drop. 🤷🏻‍♂️

Sometimes I wish I knew git better... Sometimes.

PR #198, opened 2020-05-08
[REVIEW] Port spatial window query to libcudf++

### LINKED ISSUE #157: [BUG] Update query.hpp to libcudf++
* Eliminate use of gdf_column in shapefile APIs.
* Remove all raw RMM_ALLOC/RMM_FREE and replace with rmm::device_buffer / device_vector.
* Add appropriate CUSPATIAL_EXPECTS calls. See #153

harrism:
Porting `cuspatial::spatial_window_points` to use libcudf++ types.

Depends on #197 
Closes #157  

Now called `cuspatial::points_in_spatial_window`.

TODO:
 - [x] Cython / Python changes ~(help wanted)~
 - [x] (optional for this PR) More testing


cpp/include/cuspatial/spatial_window.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * @param[in] mr     Optional `device_memory_resource` to use for allocating the output table
+ *
+ * @returns A table with two columns of the same type as the input columns. Columns 0, 1 are the
+ * (x, y) coordinates of the points in the input that fall within the query window.
+ */
+std::unique_ptr<cudf::experimental::table> points_in_spatial_window(
+  double left,
+  double bottom,
+  double right,
+  double top,
```
cwharris, line 44: Consider using orientation-agnostic variable names.
```suggestion
  double min_x,
  double min_y,
  double max_x,
  double max_y,
```

cpp/src/query/spatial_window.cu
```diff
@@ -0,0 +1,146 @@
...
+
+  CUSPATIAL_EXPECTS(not(x.has_nulls() || y.has_nulls()), "NULL point data not supported");
+
+  return cudf::experimental::type_dispatcher(x.type(),
+                                             spatial_window_dispatch(),
+                                             left,
+                                             bottom,
+                                             right,
+                                             top,
+                                             cudf::table_view{{x, y}},
```
cwharris, line 123: It's an anonymous namespace, so it doesn't matter, but why create a `table_view`?

cpp/src/query/spatial_window.cu
```diff
@@ -0,0 +1,146 @@
...
+  double bottom,
+  double right,
+  double top,
+  cudf::column_view const& x,
+  cudf::column_view const& y,
+  cudaStream_t stream,
+  rmm::mr::device_memory_resource* mr)
+{
+  CUSPATIAL_EXPECTS(x.type() == y.type(), "point type mismatch between x/y arrays");
+  CUSPATIAL_EXPECTS(x.size() == y.size(), "#of points mismatch between x/y arrays");
```
cwharris, line 113:
```suggestion
  CUSPATIAL_EXPECTS(x.size() == y.size(), "# of points mismatch between x/y arrays");
```

cpp/include/cuspatial/spatial_window.hpp
harrism: I am on the fence, because it's very common in both computer graphics and geo APIs to specify rectangles as bottom-left top-right or top-left bottomr-right.  D3D, Win32, some geo DB queries, etc.

cpp/src/query/spatial_window.cu
harrism: Because `copy_if` takes a `table_view`, not two `column_view`s.  But you have a point, so I pushed down the `table_view` creation to the call to `copy_if`.

cpp/src/query/spatial_window.cu
harrism: Fixed, thank.

cpp/src/query/spatial_window.cu
```diff
@@ -0,0 +1,144 @@
...
+                        cudf::column_device_view const& y)
+    : left{left}, bottom{bottom}, right{right}, top{top}, points_x{x}, points_y{y}
+  {
+  }
+
+  __device__ inline bool operator()(cudf::size_type i)
+  {
+    auto x = points_x.data<T>()[i];
+    auto y = points_y.data<T>()[i];
+    return x > left && x < right && y > bottom && y < top;
```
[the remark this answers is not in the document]
harrism: Good observation. Yes, the float would be promoted to double.

cpp/include/cuspatial/spatial_window.hpp
cwharris: That's true, but those systems may differ in what is treated as left and right, top and bottom, and I think cuspatial is intended to help accelerate many different systems. It could be beneficial to remain agnostic - easier to answer "is min_x left?" than "why is my code returning unexpected results?"

just my 2 cents.

cpp/src/query/spatial_window.cu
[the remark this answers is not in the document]
harrism: Parameterize the type.

cpp/include/cuspatial/spatial_window.hpp
harrism: Changed. Also now handles the case where min_x > max_x or min_y > max_y.

harrism:
I finished the Cython / Python port. I'm surprised it didn't prompt for a python-codeowners review. @trxcllnt @cwharris @thomcom would you mind reviewing the Python/Cython changes?

python/cuspatial/cuspatial/_lib/spatial_window.pxd
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2019-2020, NVIDIA CORPORATION.
+
+from cudf._lib.legacy.cudf cimport *
+
+from cudf._lib.column cimport column, column_view
+from cudf._lib.table cimport table, table_view
+from cudf._lib.move cimport move, unique_ptr
+
+cdef extern from "spatial_window.hpp" namespace "cuspatial" nogil:
+    cdef unique_ptr[table] cpp_points_in_spatial_window \
```
cwharris, line 10: In libcudf we made the decision to keep the `pxd` files in a separate directory (`_lib/cpp/`) from the `pyx` files because files within the same directory end up "smashed together" in the same "namespace". This was done so we could remove the `cpp_` prefix without conflict. Dependent files could then explicitly `import ... as ...` in the usual manner.

```suggestion
    cdef unique_ptr[table] points_in_spatial_window \
```

python/cuspatial/cuspatial/_lib/spatial_window.pyx
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
cwharris, line 1:
```suggestion
# Copyright (c) 2020, NVIDIA CORPORATION.
```

python/cuspatial/cuspatial/core/spatial_window.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
cwharris, line 1:
```suggestion
# Copyright (c) 2020, NVIDIA CORPORATION.
```

cpp/src/spatial_window/spatial_window.cu
```diff
@@ -0,0 +1,158 @@
...
+// This is passed to cudf::detail::copy_if
+template <typename T>
+struct spatial_window_filter {
+  spatial_window_filter(T window_min_x,
+                        T window_max_x,
+                        T window_min_y,
+                        T window_max_y,
+                        cudf::column_device_view const& x,
+                        cudf::column_device_view const& y)
+    : min_x{std::min(window_min_x, window_max_x)},  // support mirrored rectangles
```
cwharris, line 39: I'm on the fence about supporting mirrored windows.

There is a case where keeping the values as-is becomes a feature in and of itself. If an algorithm upstream _intentionally_ produces a "negative-size" window, that is an implicit zero-results query we can optimize for.

On the one hand, a user might misunderstand the API and wonder why no results are being returned.
On the other hand, a user might provide a "negative-size" or "too small" window size intentionally, only to have us rectify it, and wonder why _any_ results are returning at all.

If we rectify the window, we actually lose flexibility in how the API can be used (and complicate the documentation, since there are now 4 "quadrants" for rectifiable window coordinates).

cwharris: Alternatively, we could move the rectify to the python layer and make optional.

python/cuspatial/cuspatial/_lib/spatial_window.pyx
```diff
@@ -0,0 +1,33 @@
...
+
+from cudf._lib.table cimport table, Table
+from cudf._lib.move cimport move, unique_ptr
+from cudf._lib.column cimport column, Column
+
+cpdef points_in_spatial_window(
```
cwharris, line 7: I think this can be cdef. the rules behind def, cdef, and cpdef are weird. I'm not sure if any one of them effects compilation time, but it takes practically no time to compile cython compared to the rest of the code base anyway, so my comment is mostly irrelevant.

cpp/src/spatial_window/spatial_window.cu
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright (c) 2019, NVIDIA CORPORATION.
```
cwharris, line 2:
```suggestion
 * Copyright (c) 2020, NVIDIA CORPORATION.
```

cpp/tests/spatial_window/spatial_window_test.cpp
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright (c) 2019, NVIDIA CORPORATION.
```
cwharris, line 2:
```suggestion
 * Copyright (c) 2020, NVIDIA CORPORATION.
```

python/cuspatial/cuspatial/core/spatial_window.py
```diff
@@ -0,0 +1,37 @@
...
+
+from cudf import DataFrame
+from cudf.core.column import as_column
+
+from cuspatial._lib import spatial_window
+
+
+def points_in_spatial_window(min_x, max_x, min_y, max_y, xs, ys):
+    """ Return only the subset of coordinates that fall within the numerically
+    closed borders [,] of the defined bounding box.
```
cwharris, line 11: The implementation indicates that the borders would be open, I think? (I'm still fuzzy about open/closed set terminology).
`return x > min_x && x < max_x && y > min_y && y < max_y;`

It's possible we may want to use `>=`/`<=` in the implementation?

cwharris:
For the `check-style` CI failure, you can either run the commands individually or use pre-commit from inside a repo directory (this works in host and rapids-compose container, I believe, but I usually run inside the container).

```sh
conda install pre-commit # or pip install pre-commit
pre-commit run
```
You can also install them to run automatically on commit
```sh
pre-commit install
```

trxcllnt:
You can also run `lint-cuspatial-python` in the rapids-compose container.

cpp/src/spatial_window/spatial_window.cu
harrism: The previous behavior was to return an error, rather than zero results.

python/cuspatial/cuspatial/core/spatial_window.py
harrism: I actually didn't understand what that meant in the docs (`[, ]` is not obvious!). Now I do.  

I don't think "closed" is a useful term here. Typically "inclusive/inclusive" are used. Which does that mean? The implementation always used < not <=.   <= might be more intuitive...  But I'm worried we're going to get the usual misunderstandings about floating point representation...

cpp/src/spatial_window/spatial_window.cu
harrism: I'm on the fence too, so we'll just leave it until someone files an issue. :)

python/cuspatial/cuspatial/core/spatial_window.py
harrism: I fixed the documentation to match the real behavior (<, not <=) and not use vague terms.

cpp/src/spatial_window/spatial_window.cu
harrism: Updated docs to reflect real behavior.

python/cuspatial/cuspatial/core/spatial_window.py
cwharris: My understanding is that a closed interval `[a, b]` is a closed set.

(I don't think the inverse is true always).

python/cuspatial/cuspatial/_lib/spatial_window.pyx
harrism: When I change this to cdef the python function fails to call it, and I get a Cython warning about unused function. I think I'll leave it.

Issue #154, opened 2020-03-27
[BUG] Update shapefile_readers.hpp to libcudf++


harrism:
* Eliminate use of gdf_column in shapefile APIs.
* Remove all raw `RMM_ALLOC/RMM_FREE` and replace with `rmm::device_buffer / device_vector`.
* Add appropriate `CUSPATIAL_EXPECTS` calls. See #153

harrism:
Fixed by #204 and #200.

PR #200, opened 2020-05-09
[REVIEW] Update shapefile_readers.hpp to libcudf++

cwharris:
Contributes to #154.


GPUtester:

cpp/include/cuspatial/shapefile_reader.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <cudf/column/column.hpp>
+#include <cudf/types.hpp>
+
+namespace cuspatial {
+
+/**
+ * @brief read polygon data from an ESRI Shapefile.
+ *
+ * @param[in] filename: ESRI Shapefile file path (usually ends in .shp)
+ * @param[in] mr:       Optional, The resource to use for all allocations
```
harrism, line 28:
```suggestion
 * @param[in] mr:       Optional, The resource to use to allocate the returned data
```

harrism: Only output data should be allocated with mr. Internal temporary allocations should use the default resource.

cpp/src/io/shp/polygon_shapefile_reader.cu
```diff
@@ -0,0 +1,103 @@
...
+
+#include <memory>
+#include <string>
+
+namespace cuspatial {
+namespace detail {
+
+template<typename T>
+std::unique_ptr<cudf::column>
+make_column(std::vector<T> source, cudaStream_t stream, rmm::mr::device_memory_resource* mr)
```
harrism, line 37: Hmmm, seems like we are missing a make_column_from_vector factory in cuDF?

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+namespace {
+
+cudf::size_type read_ring(OGRLinearRing const& ring,
+                          std::vector<double>& xs,
+                          std::vector<double>& ys)
+{
+    cudf::size_type num_vertices = ring.getNumPoints();
+
+    // append points in reverse order
```
harrism, line 34: Why in reverse order?

cwharris: Good question. That was the original implementation, I believe. I’ll see if I can find out _why_, though.

cwharris: @zhangjianting may know the answer.

cpp/src/io/shp/polygon_shapefile_reader.cu
cwharris: Opened rapidsai/cudf#5150 to track this.

cpp/include/cuspatial/shapefile_reader.hpp
```diff
@@ -0,0 +1,47 @@
...
+ *  2 - FLOAT64: x component of polygon points
+ *  3 - FLOAT64: y component of polygon points
+ *
+ * @note The number of polygons is equal to the length of the first column
+ *
+**/
+std::tuple<std::unique_ptr<cudf::column>,
+           std::unique_ptr<cudf::column>,
+           std::unique_ptr<cudf::column>,
+           std::unique_ptr<cudf::column>>
```
trxcllnt, line 42: I just remembered we can't read tuples of non-copyable types in Cython, so we need to update this to return a `vector<unique_ptr<column>>`

trxcllnt: Done.

trxcllnt:
Remembered Cython can't unpack `std::tuple` of non-copyable types like `unique_ptr`. I changed the `read_polygon_shapefile` C++ return type to `std::vector<std::unique_ptr<cudf::column>>`, all good now.

cwharris:
@trxcllnt That applies to tuples, in general, but actually cython doesn't have to unpack _pairs_, we can move them individually as `.first` and`.last` instead.

PR #199, opened 2020-05-09
[REVIEW] fix coordinate transform tests

cwharris:
Addresses #174 and #192 build errors.

PR #201, opened 2020-05-10
[REVIEW] Update trajectory cython to libcudf++

trxcllnt:
This PR updates the Cython and related tests to call the new libcuspatial++ trajectory APIs.


trxcllnt:
CI seems blocked until https://github.com/rapidsai/cuspatial/pull/199 is merged.

python/cuspatial/cuspatial/utils/traj_utils.py
```diff
@@ -1,3 +1,47 @@
+import numpy as np
+
+from cudf.utils.dtypes import is_datetime_dtype
+
+
+def normalize_point_columns(xs, ys):
```
harrism, line 6: Should this be a utility for more than just trajectories? spatial_window could use this too, I think.

python/cuspatial/cuspatial/tests/test_traj.py
```diff
@@ -9,75 +9,49 @@
 import cuspatial
```
harrism, line 10: Consider renaming file to `test_trajectory.py`

python/cuspatial/cuspatial/utils/traj_utils.py
```diff
@@ -1,3 +1,47 @@
+import numpy as np
```
harrism, line 1: Consider renaming file to `trajectory_utils.py`

python/cuspatial/cuspatial/utils/traj_utils.py
trxcllnt: Yes that seems appropriate.

python/cuspatial/cuspatial/__init__.py
```diff
@@ -7,12 +7,7 @@
 )
 from .core.interpolate import CubicSpline
 from .core.spatial_window import points_in_spatial_window
-from .core.trajectory import (
-    derive,
-    distance_and_speed,
-    spatial_bounds,
-    subset_trajectory_id,
-)
+from .core.trajectory import derive, distance_and_speed, spatial_bounds
```
trxcllnt, line 10: Do we want to be re-exporting these functions to the top-level `cuspatial` namespace? Worried things like this could lead to confusion:
```
# derive what? should this be renamed?
cuspatial.derive(...)

# or should we force this?
cuspatial.trajectory.derive(...)
```

harrism: I defer to @thomcom @kkraus14 .

kkraus14: I'm not familiar with this API space as well where I would defer to @thomcom as well.

cwharris: IMO force `cuspatial.trajectory.derive`, because we can always add the other later, whereas removing one is way harder. 2cents.

cwharris: @trxcllnt proposed calling the function `device_trajectory`, which sounds good to me.

trxcllnt: Yeah having different names from C++ seemed unclear, so I changed Python to be the same:
```python
cuspatial.derive_trajectories()
cuspatial.trajectory_bounding_boxes()
cuspatial.trajectory_distances_and_speeds()
```

PR #202, opened 2020-05-12
[REVIEW] Update haversine cython to libcudf++

trxcllnt:
Based on https://github.com/rapidsai/cuspatial/pull/189


cwharris:
LGTM, waiting on base PR before approving.

python/cuspatial/cuspatial/_lib/spatial.pyx
```diff
@@ -13,6 +13,11 @@ from cudf._lib.legacy.cudf import *
 from cuspatial._lib.cpp.coordinate_transform cimport (
```
harrism, line 13: I didn't really understand why everything was lumped in a file called "spatial", which is why I created a new file for spatial_window...

Perhaps distance metrics should get their own file too? (Distance_metrics.pyx)?

python/cuspatial/cuspatial/_lib/spatial.pyx
```diff
@@ -55,36 +60,18 @@ cpdef cpp_point_in_polygon_bitmap(
...
-    y1 = y1.astype('float64')._column
-    x2 = x2.astype('float64')._column
-    y2 = y2.astype('float64')._column
-
-    cdef gdf_column* c_x1 = column_view_from_column(x1)
-    cdef gdf_column* c_y1 = column_view_from_column(y1)
-    cdef gdf_column* c_x2 = column_view_from_column(x2)
-    cdef gdf_column* c_y2 = column_view_from_column(y2)
+cpdef haversine_distance(Column x1, Column y1, Column x2, Column y2):
+    cdef column_view c_x1 = x1.view()
```
harrism, line 64: When is it necessary in a cpdef to use explicit types? I didn't need them in points_in_spatial_window, I don't think.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -1,11 +1,14 @@
 # Copyright (c) 2019-2020, NVIDIA CORPORATION.
 
+import numpy as np
```
harrism, line 3: Likewise, does it make sense to lump everything in a file called gis.py? I moved spatial_window out... Did I do the wrong thing?

python/cuspatial/cuspatial/_lib/spatial.pyx
cwharris: The python community as a whole seems to like keeping everything in the same file. :disappointed: 

https://github.com/rapidsai/cudf/blob/branch-0.14/python/cudf/cudf/core/dataframe.py

@thomcom any opinion here

python/cuspatial/cuspatial/_lib/spatial.pyx
cwharris: Adding types allows Cython to do some runtime type checking for function parameters, and compile-time checking for local variables, _I think_.

trxcllnt: Yeah Cython can infer the type of `c_x1` because `x1` has a type, but I think the cudf convention is to add explicit types to C variables to distinguish them from Python variables.

python/cuspatial/cuspatial/_lib/spatial.pyx
trxcllnt: Yeah I agree. Honestly I'd be fine with naming it after the C++ header like we do in cudf. Also I just noticed we need to fix the Cython build's include paths. The header paths should be this:

```diff
- cdef extern from "haversine.hpp" namespace "cuspatial" nogil:
+ cdef extern from "cuspatial/haversine.hpp" namespace "cuspatial" nogil:
```

python/cuspatial/cuspatial/core/gis.py
trxcllnt: No I think you did the right thing. I left the function in place to make it easier/faster to review.

I'm somewhat ambivalent about the file structure at the moment since everything's re-exported on the top-level `cuspatial` module. But if you or others think there's benefit to splitting these out into separate/more descriptive files as we go, I'd be happy to do that too.

cpp/tests/spatial/point_in_polygon_test.cpp
```diff
@@ -222,6 +223,7 @@ TYPED_TEST(PointInPolygonTest, 31PolygonSupport)
                 return  1.0;
             case 3:
             case 4:
+            default:
```
cwharris, line 226: Should probably be `CUSPATIAL_FAIL`, but it makes absolutely no difference. TBH I should have written it as
```
case 1:
case 2:
  return  1.0;
default:
  return -1.0;
```

trxcllnt: I considered that too, but left the current behavior in place since the behavior is more explicit.

PR #203, opened 2020-05-13
[REVIEW] Port point_in_polygon.hpp cython files

cwharris:
Contributes to #158


GPUtester:

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -163,16 +172,36 @@ def point_in_polygon_bitmap(
...
+        test_points_y,
+        as_column(poly_offsets, dtype="int32"),
+        as_column(poly_ring_offsets, dtype="int32"),
+        poly_points_x,
+        poly_points_y,
     )
 
-    result_binary = gis_utils.pip_bitmap_column_to_binary_array(
-        polygon_bitmap_column=bitmap_result, width=len(polygon_fpos)
+    result = gis_utils.pip_bitmap_column_to_binary_array(
```
harrism, line 200: Wow, so we are exploding the bitmaps to boolean columns anyway... So it probably would be more efficient (and eliminate the 32-poly limit) to just do that in C++...

python/cuspatial/cuspatial/utils/gis_utils.py
```diff
@@ -33,7 +33,7 @@ def apply_binarize(in_col, width):
 
 
 def pip_bitmap_column_to_binary_array(polygon_bitmap_column, width):
```
harrism, line 35: Yikes, libcudf probably has a function for this!

python/cuspatial/cuspatial/core/gis.py
trxcllnt: I think this to cast the matrix columns to `dtype="bool"` just below, we could definitely do this without numba.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -163,16 +172,36 @@ def point_in_polygon_bitmap(
     DataFrame: a DataFrame of Boolean values indicating whether each point
     falls within each polygon.
     """
-    bitmap_result = cpp_point_in_polygon_bitmap(
-        x_points, y_points, polygon_fpos, polygon_rpos, polygons_x, polygons_y
+
+    if len(poly_offsets) == 0:
+        return DataFrame()
+
```
cwharris, line 178: fwiw, the reason we're doing this is because: https://github.com/rapidsai/cudf/pull/5162

PR #205, opened 2020-05-13
[REVIEW] Port hausdorff.hpp cython to libcudf++

cwharris:
Contributes to #160.


GPUtester:

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -62,12 +65,20 @@ def directed_hausdorff_distance(x, y, count):
...
-    return DataFrame.from_gpu_matrix(result.to_gpu_array().reshape(dim, dim))
+    num_spaces = len(points_per_space)
+    if num_spaces == 0:
+        return DataFrame()
+    result = cpp_directed_hausdorff_distance(
+        xs.astype("float64")._column,
+        ys.astype("float64")._column,
+        points_per_space.astype("int32")._column,
+    )
+    result = result.to_gpu_array()
```
harrism, line 79: Is the C++ API not outputting in the format expected? Is this just zero-copy reshaping?

cwharris: the cpp deals in columns and returns a flattened matrix as a column for perf reasons (assuming), and the python layer wants it in a matrix. I've been leaving most of the cython/python code as-is, mostly because the answer to a lot of these questions is "I don't know for sure", and my primary focus has been deleting legacy. :man_shrugging: .

trxcllnt: `to_gpu_array()` might've changed since this was written -- it used to return a zero-copy view, but now it does `dropna()` or `fillna()`, and returns the view of the non-null result. So likely better to change this to:
```suggestion
    result = cupy.asarray(result.data_array_view)
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -62,12 +65,20 @@ def directed_hausdorff_distance(x, y, count):
...
-    dim = len(count)
-    return DataFrame.from_gpu_matrix(result.to_gpu_array().reshape(dim, dim))
+    num_spaces = len(points_per_space)
+    if num_spaces == 0:
+        return DataFrame()
+    result = cpp_directed_hausdorff_distance(
+        xs.astype("float64")._column,
+        ys.astype("float64")._column,
+        points_per_space.astype("int32")._column,
+    )
```
trxcllnt, line 78:
```suggestion
    xs, ys = normalize_point_columns(as_column(xs), as_column(ys))
    result = cpp_directed_hausdorff_distance(
        xs,
        ys,
        as_column(points_per_space, dtype="int32"),
    )
```

PR #211, opened 2020-05-16
[WIP] Simplify Hausdorff API and Implementation

### LINKED ISSUE #210: [BUG] hausdorff.cu uses custom kernel unnecessarily
hausdorff.cu uses a custom kernel to compute Hausdorff distances between a set of spaces.

Rather than a custom kernel, use thrust to eliminate the one-result-per-warp limitation and flatten nested loops. Hypothetically this should increase performance, especially when processing more spaces of different sizes.

- [x] Create Hausdorff Benchmarks
- [x] Replace custom kernel
- [x] Verify performance is at least as good as the custom kernel.

cwharris:
Closes #210
- [x] simplify hausdorff python tests
- [x] let hausdorff apis accept offsets instead of lengths and adjust tests
- [x] port kernel to thrust
- [ ] add comprehensive tests


GPUtester:

PR #214, opened 2020-05-21
[REVIEW] Install gdal>=3.0.2 in build.sh

trxcllnt:
Testing if this will fix the issue where gdal 2.4.4 is getting installed instead of >=3.0.2.


GPUtester:

trxcllnt:
Comparing to other CI jobs like [this one](https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fcuspatial%2Fprb%2Fcuspatial-gpu-build/detail/cuspatial-gpu-build/1993/pipeline), it seems like this change does indeed install gdal >= 3.0.2. Maybe this will propagate to the nightly packages/containers?

dillon-cullinan:
> Maybe this will propagate to the nightly packages/containers?

Nightly packages are conda based, so you would have to add this as a dependency into the conda meta.yaml (which seems to have already been done).

trxcllnt:
This is a runtime dependency. I've already updated it in the `meta.yml` files here:
https://github.com/rapidsai/cuspatial/blob/branch-0.14/conda/recipes/cuspatial/meta.yaml#L33. But for some reason, 2.4.4 is still in the nightly containers, and this is the only reference to the old gdal version that I've found.

mike-wendt:
@trxcllnt you need to update this as well https://github.com/rapidsai/integration/blob/branch-0.14/conda/recipes/nightly-versions.yaml#L59 this is where the nightly containers get their versioning and soon to be gpuCI containers are well. The overrides that you have in the `ci/gpu/build.sh` will be removed shortly and all versioning will be in the two YAML files here https://github.com/rapidsai/integration/blob/branch-0.14/conda/recipes/

PR #221, opened 2020-05-23
[REVIEW] Add python methods to api.rst, fix formatting

trxcllnt:
This PR fixes a few formatting issues in our Python docs and adds all the new methods to the Sphinx `api.rst` manifest.

If you're using `rapids-compose`, here's an easy way to build the python docs:
```shell
make html -C $CUSPATIAL_HOME/docs
```

Or if you're doing a lot of docs editing and want to auto-rebuild the docs on save, you can do this:
```shell
# Install `entr` to watch files (https://systutorials.com/docs/linux/man/1-entr/)
echo "rapids" | sudo -S apt update && echo "rapids" | sudo -S apt install entr

# Launch `entr` to re-run `make html` any time the source or config files change
find $CUSPATIAL_HOME/python $CUSPATIAL_HOME/docs/source \
  | entr sh -c "make html -C $CUSPATIAL_HOME/docs"
```

<details><summary>Expand this to see a screenshot of the rendered docs</summary>

</details>

### DIFF
--- a/python/cuspatial/cuspatial/core/gis.py
+++ b/python/cuspatial/cuspatial/core/gis.py
@@ -21,53 +21,61 @@ def directed_hausdorff_distance(xs, ys, points_per_space):
     """Compute the directed Hausdorff distances between all pairs of
     spaces.
 
-    params
-    xs: x-coordinates
-    ys: y-coordinates
-    points_per_space: number of points in each space
-
     Parameters
     ----------
-    {params}
+    xs
+        column of x-coordinates
+    ys
+        column of y-coordinates
+    points_per_space
+        number of points in each space
 
-    Example
+    Returns
     -------
+    result : cudf.DataFrame
+        The pairwise directed distance matrix with one row and one
+        column per input space; the value at row i, column j represents the
+        hausdorff distance from space i to space j.
+
+    Examples
+    --------
     The directed Hausdorff distance from one space to another is the greatest
     of all the distances between any point in the first space to the closest
     point in the second.
 
-    [Wikipedia](https://en.wikipedia.org/wiki/Hausdorff_distance)
+    `Wikipedia <https://en.wikipedia.org/wiki/Hausdorff_distance>`_
 
-    Consider a pair of lines on a grid.
+    Consider a pair of lines on a grid::
 
-     |
-     o
-    -oxx-
-     |
-    o = [[0, 0], [0, 1]]
-    x = [[1, 0], [2, 0]]
+             :
+             x
+        -----xyy---
+             :
+             :
 
-    o[0] is the closest point in o to x. The distance from o[0] to the farthest
-    point in x = 2.
-    x[0] is the closest point in x to o. The distance from x[0] to the farthest
-    point in o = 1.414.
+    x\\ :sub:`0` = (0, 0), x\\ :sub:`1` = (0, 1)
 
-        result = cuspatial.directed_hausdorff_distance(
-            cudf.Series([0, 1, 0, 0]),
-            cudf.Series([0, 0, 1, 2]),
-            cudf.Series([2, 2,]),
+    y\\ :sub:`0` = (1, 0), y\\ :sub:`1` = (2, 0)
+
+    x\\ :sub:`0` is the closest point in ``x`` to ``y``. The distance from
+    x\\ :sub:`0` to the farthest point in ``y`` is 2.
+
+    y\\ :sub:`0` is the closest point in ``y`` to ``x``. The distance from
+    y\\ :sub:`0` to the farthest point in ``x`` is 1.414.
+
+    Compute the directed hausdorff distances between a set of spaces
+
+    >>> result = cuspatial.directed_hausdorff_distance(
+            [0, 1, 0, 0], # xs
+            [0, 0, 1, 2], # ys
+            [2,    2],    # points_per_space
         )
-        print(result)
+    >>> print(result)
              0         1
         0  0.0  1.414214
         1  2.0  0.000000
-
-    Returns
-    -------
-    DataFrame: The pairwise directed distance matrix with one row and one
-    column per input space; the value at row i, column j represents the
-    hausdorff distance from space i to space j.
     """
+
     num_spaces = len(points_per_space)
     if num_spaces == 0:
         return DataFrame()
@@ -84,19 +92,23 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
     """ Compute the haversine distances between an arbitrary list of lon/lat
     pairs
 
-    params
-    p1_lon: longitude of first set of coords
-    p1_lat: latitude of first set of coords
-    p2_lon: longitude of second set of coords
-    p2_lat: latitude of second set of coords
-
     Parameters
     ----------
-    {params}
+    p1_lon
+        longitude of first set of coords
+    p1_lat
+        latitude of first set of coords
+    p2_lon
+        longitude of second set of coords
+    p2_lat
+        latitude of second set of coords
 
-    returns
-    Series: distance between all pairs of lat/lon coords
+    Returns
+    -------
+    result : cudf.Series
+        The distance between all pairs of lat/lon coordinates
     """
+
     p1_lon, p1_lat, p2_lon, p2_lat = normalize_point_columns(
         as_column(p1_lon),
         as_column(p1_lat),
@@ -107,21 +119,28 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
 
 
 def lonlat_to_cartesian(origin_lon, origin_lat, input_lon, input_lat):
-    """ Convert lonlat coordinates to km x,y coordinates based on some camera
-    origin.
-
-    params
-    origin_lon: float64 - longitude camera
-    origin_lat: float64 - latitude camera
-    input_lon: Series of longitude coords to convert to x
-    input_lat: Series of latitude coords to convert to y
+    """
+    Convert lon/lat coordinates to kilometer x,y coordinates with respect to
+    an origin camera position.
 
     Parameters
     ----------
-    {params}
+    origin_lon : ``number``
+        camera longitude
+    origin_lat : ``number``
+        camera latitude
+    input_lon : ``Series`` or ``list``
+        longitude coordinates to convert to x
+    input_lat : ``Series`` or ``list``
+        latitude coordinates to convert to y
 
-    returns
-    DataFrame: 'x', 'y' columns for new km positions of coords
+    Returns
+    -------
+    result : cudf.DataFrame
+        x : cudf.Series
+            x-coordinate of the new positions in kilometers
+        y : cudf.Series
+            y-coordinate of the new positions in kilometers
     """
     result = cpp_lonlat_to_cartesian(
         origin_lon, origin_lat, input_lon._column, input_lat._column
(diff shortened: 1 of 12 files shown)


GPUtester:

python/cuspatial/cuspatial/core/interpolate.py
```diff
@@ -46,16 +46,16 @@ class CubicSpline:
...
         y = cudf.Series(np.random.random(100*1000)).astype('float32')
         prefix_sum = cudf.Series(cp.arange(1000)*100).astype('int32')
-        new_samples = cudf.Series(np.repeat(np.linspace(0, 100, 1000), 1000)
-            .astype('float32'))
+        new_samples = cudf.Series(
+            np.repeat(np.linspace(0, 100, 1000), 1000)
+        ).astype('float32')
 
         curve = cuspatial.CubicSpline(t, y, prefixes=prefix_sum)
         new_points = curve(new_samples, prefix_sum*10)
```
trxcllnt, line 61: I tried running this example in a notebook, but I'm getting CUDA illegal address errors. Does this need updating for any recent API or implementation changes?

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -84,19 +92,23 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
...
+        longitude of second set of coords
+    p2_lat
+        latitude of second set of coords
 
-    returns
-    Series: distance between all pairs of lat/lon coords
+    Returns
+    -------
+    result : cudf.Series
+        The distance between all pairs of lat/lon coordinates
```
harrism, line 109: Order seems to switch here?

```suggestion
        The distance between all pairs of lon/lat coordinates
```

trxcllnt:
@thomcom @cwharris do you guys have feedback for this PR?

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -21,53 +21,61 @@ def directed_hausdorff_distance(xs, ys, points_per_space):
...
+        column of y-coordinates
+    points_per_space
+        number of points in each space
 
-    Example
+    Returns
     -------
+    result : cudf.DataFrame
+        The pairwise directed distance matrix with one row and one
+        column per input space; the value at row i, column j represents the
```
cwharris, line 37: Did we verify this is correct (aka not vice versa)? I never did. I wrote it so the existing tests would pass, and I believe the existing docs state it this way, but I never confirmed the existing docs were correct. :/

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -21,53 +21,61 @@ def directed_hausdorff_distance(xs, ys, points_per_space):
...
-    [Wikipedia](https://en.wikipedia.org/wiki/Hausdorff_distance)
+    `Wikipedia <https://en.wikipedia.org/wiki/Hausdorff_distance>`_
+
+    Consider a pair of lines on a grid::
+
+             :
+             x
+        -----xyy---
+             :
+             :
```
cwharris, line 54: What does this have to do with Hausdorff?

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -107,21 +119,28 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
...
-    origin.
-
-    params
-    origin_lon: float64 - longitude camera
-    origin_lat: float64 - latitude camera
-    input_lon: Series of longitude coords to convert to x
-    input_lat: Series of latitude coords to convert to y
+    """
+    Convert lon/lat coordinates to kilometer x,y coordinates with respect to
+    an origin camera position.
```
cwharris, line 124: Convert lon/lat to x,y coordinates relative to an origin lon/lat. Results are scaled relative to the size of Earth in kilometers.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -107,21 +119,28 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
...
 
     Parameters
     ----------
-    {params}
+    origin_lon : ``number``
+        camera longitude
+    origin_lat : ``number``
+        camera latitude
```
cwharris, line 131:
```suggestion
    origin_lon : ``number``
        longitude offset  (this is subtracted from each input before converting to x,y)
    origin_lat : ``number``
        latitude offset (this is subtracted from each input before converting to x,y)
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -107,21 +119,28 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
...
 
-    returns
-    DataFrame: 'x', 'y' columns for new km positions of coords
+    Returns
+    -------
+    result : cudf.DataFrame
+        x : cudf.Series
+            x-coordinate of the new positions in kilometers
+        y : cudf.Series
+            y-coordinate of the new positions in kilometers
```
cwharris, line 143:
```suggestion
    result : cudf.DataFrame
        x : cudf.Series
            x-coordinates of the input relative to the size of Earth in kilometers.
        y : cudf.Series
            y-coordinate of the input relative to the size of Earth in kilometers
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -142,51 +161,55 @@ def point_in_polygon(
...
+
+    Test whether 3 points fall within either of two polygons
+
+    >>> result = cuspatial.point_in_polygon(
+        [0, -8, 6.0],                             # test_points_x
+        [0, -8, 6.0],                             # test_points_y
+        cudf.Series([0, 1], index=['nyc', 'dc']), # poly_offsets
+        [0, 3],                                   # ring_offsets
+        [-10, 5, 5, -10, 0, 10, 10, 0],           # poly_points_x
+        [-10, -10, 5, 5, 0, 0, 10, 10],           # poly_points_y
```
cwharris, line 190: I believe the first point in each ring must be repeated at the end of the ring (completing the loop). This is contrary to the c++ documentation. My initial tests seemed to indicate otherwise, but I believe there are situations when the first points must be repeated.  Specifically in the case when the "missing edge" should intersects an x-axis determined by the y-coordinate of the input point. This still needs to be verified. For example:

```
// 3 -> 0 may need to be explicitly indicated in the ring by repeating the first point at the end of the ring.
// the test for point A will be correct, but tests for point B and C will be incorrect.

ring points: [0, 1, 2, 3]

   0-----1
         |
A     B  |  C
         |
   3-----2
```

This didn't come up in initial testing because the ring's last edge was aligned with the x-axis:
```
ring points: [3, 0, 1, 2]:

   0-----1
   |     |
A  |  B  |  C
   |     |
   3     2
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -142,51 +161,55 @@ def point_in_polygon(
...
-        # Point 0: (0, 0) falls in both polygons
-        # Point 1: (-8, -8) falls in the first polygon
-        # Point 2: (6.0, 6.0) falls in the second polygon
+
+    Test whether 3 points fall within either of two polygons
+
+    >>> result = cuspatial.point_in_polygon(
+        [0, -8, 6.0],                             # test_points_x
+        [0, -8, 6.0],                             # test_points_y
+        cudf.Series([0, 1], index=['nyc', 'dc']), # poly_offsets
```
cwharris, line 187: Maybe use hudson river in place of dc, since it partially overlaps with nyc (may be less confusing).
```suggestion
        [0, -8, 6.0],                             # test_points_y
        cudf.Series([0, 1], index=['nyc', 'hudson river']), # poly_offsets
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -142,51 +161,55 @@ def point_in_polygon(
...
+        [0, 3],                                   # ring_offsets
+        [-10, 5, 5, -10, 0, 10, 10, 0],           # poly_points_x
+        [-10, -10, 5, 5, 0, 0, 10, 10],           # poly_points_y
+    )
+    # The result of point_in_polygon is a DataFrame of Boolean
+    # values indicating whether each point (rows) falls within
+    # each polygon (columns).
+    >>> print(result)
+                nyc            dc
```
cwharris, line 196:
```suggestion
                nyc            hudson river
```

python/cuspatial/cuspatial/core/interpolate.py
cwharris: @thomcom

python/cuspatial/cuspatial/core/gis.py
trxcllnt: I don't know, I assumed you wrote this diagram? I just corrected the formatting.

python/cuspatial/cuspatial/core/gis.py
trxcllnt: please make a github suggestion as to what to change?

python/cuspatial/cuspatial/io/shapefile.py
```diff
@@ -8,7 +8,28 @@
...
+        ESRI Shapefile file path (usually ends in ``.shp``)
+
+    Returns
+    -------
+    result  : tuple (cudf.Series, cudf.Series, cudf.DataFrame)
+    poly_offsets   : cudf.Series(dtype=np.int32)
+        Offsets of the first ring in each polygon
+    ring_offsets   : cudf.Series(dtype=np.int32)
+        Offsets of the first point in each ring
+    points  : cudf.DataFrame
```
cwharris, line 26: It's strange that we don't return a single DataFrame, or four Series... But this is a documentation PR. :)

python/cuspatial/cuspatial/core/trajectory.py
```diff
@@ -92,35 +131,49 @@ def trajectory_bounding_boxes(num_trajectories, object_ids, xs, ys):
...
     ----------
-    {params}
+    num_trajectories
+        number of trajectories (unique object ids)
+    object_ids
+        column of object (e.g., vehicle) ids
+    xs
+        column of x-coordinates (in kilometers)
+    ys
+        column of y-coordinates (in kilometers)
```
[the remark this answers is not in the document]
trxcllnt: The inputs must be in kilometers if meters/second is the desired output. Though if the inputs were in miles, the output speeds would be milli-miles/second since we're just dividing the input units 1000.

python/cuspatial/cuspatial/io/shapefile.py
trxcllnt: I believe all these could be different lengths?

python/cuspatial/cuspatial/core/gis.py
cwharris: I don’t believe I wrote these docs. Doesn't seem to be relevant to the rest of the example.

python/cuspatial/cuspatial/core/trajectory.py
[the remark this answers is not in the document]
cwharris: Ah, I see. Perhaps We should leave the results in kilometers/s, or add an optional multiplier? Not relevant this PR, though.

python/cuspatial/cuspatial/io/shapefile.py
cwharris: Right. That makes sense. 🤦

PR #222, opened 2020-05-23
[REVIEW] Fix potential thrust launch failure in quadtree building

trxcllnt:
This PR fixes a potential thrust launch failure in the case there are no valid non-leaf quadrants. This is easy to trigger if the user-provided `scale` value doesn't lead to computing any common Morton code keys, and all points are grouped into separate top-level leaf quadrants.

### DIFF
--- a/cpp/src/indexing/construction/detail/phase_1.cuh
+++ b/cpp/src/indexing/construction/detail/phase_1.cuh
@@ -292,16 +292,16 @@ inline auto make_full_levels(cudf::column_view const &x,
   // This allows us to avoid building the "full" quads for each level only to turn around and
   // remove them for having no children. This scenario causes the `thrust::transform` call in
   // `construct_non_leaf_indicator()` to launch on zero elements.
-  if (num_top_quads >= static_cast<cudf::size_type>(quad_keys.size())) {
-    return std::make_tuple(std::move(point_indices),
-                           std::move(quad_keys),
-                           std::move(quad_point_count),
-                           std::move(rmm::device_vector<uint32_t>(0)),
-                           std::move(rmm::device_vector<int8_t>(0)),
-                           num_top_quads,
-                           0,
-                           0);
-  }
+  // if (num_top_quads >= static_cast<cudf::size_type>(quad_keys.size())) {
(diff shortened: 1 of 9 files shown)


GPUtester:

cpp/src/indexing/construction/detail/phase_1.cuh
```diff
@@ -287,6 +287,22 @@ inline auto make_full_levels(cudf::column_view const &x,
                                               thrust::plus<uint32_t>(),
                                               stream);
 
+  // Optimization: return early if all the nodes are top-level leaf children of the root.
+  //
+  // This allows us to avoid building the "full" quads for each level only to turn around and
+  // remove them for having no children. This scenario causes the `thrust::transform` call in
+  // `construct_non_leaf_indicator()` to launch on zero elements.
+  // if (num_top_quads >= static_cast<cudf::size_type>(quad_keys.size())) {
```
harrism, line 295: Remove commented out code...

cpp/src/indexing/construction/detail/phase_2.cuh
```diff
@@ -235,6 +235,12 @@ inline std::pair<uint32_t, uint32_t> remove_unqualified_quads(
   cudf::size_type level_1_size,
   cudaStream_t stream)
 {
+  std::cerr << "min_size: " << min_size << std::endl;
```
harrism, line 238: Remove debugging code...

cpp/src/indexing/construction/detail/phase_1.cuh
trxcllnt: Ah yes, I pushed this yesterday so @zhangjianting could try out some different fixes. Forgot to change the title back to WIP.

trxcllnt:
I figured out the issue. After computing `num_top_quads`, we need to resize the vectors so there's enough space to write all the non-empty sub quadrants:
https://github.com/rapidsai/cuspatial/blob/c53ccf40e74f80230b2b8c6066c225f88895cac4/cpp/src/indexing/construction/detail/phase_1.cuh#L294-L296

PR #225, opened 2020-05-27
[REVIEW] Add short commit to conda package

dillon-cullinan:
Removing unnecessary git clone from build script as well. cuDF should not need to be cloned in the CPU build script since the conda build should download the correct version.


GPUtester:

PR #238, opened 2020-06-09
[REVIEW] Fix library and include paths in CMakeLists.txt and setup.py

trxcllnt:
Applies similar changes as https://github.com/rapidsai/cudf/pull/5398,

* Remove incorrect include path in `setup.py`
* Use FetchContent to get cub, thrust, and libcudacxx
* use `GDAL::GDAL` from CMake's `find_package(GDAL)`

The last two changes should make our CMakeLists.txt a bit more portable/compatible with CMake `ExternalProject_Add()`.


python/cuspatial/setup.py
```diff
@@ -31,6 +31,8 @@
 
 cuda_include_dir = os.path.join(CUDA_HOME, "include")
 
+CUSPATIAL_ROOT = os.environ.get("CUSPATIAL_ROOT", "../../cpp/build/")
```
rlratzel, line 34: I'm curious, when running a setup.py, what reason would someone have for **not** using the build in `../../cpp/build`?  Seems like that would be confusing, but maybe I'm missing an important use case.  If so, maybe add a comment?

trxcllnt: I don't want to make any assumptions about where the C++ build directory is. In `rapids-compose` it's `cpp/build/release` (and that's a symlink to `cpp/build/cuda-10.2/fix__cmake-paths/release`).

Most of our other scripts assume a convention that `$<PROJ>_ROOT` is the C++ build dir, so this follows that if set, and it not it uses our CI paths as the fallback path.

cpp/CMakeLists.txt
```diff
@@ -183,22 +198,21 @@ endif (RMM_INCLUDE AND RMM_LIBRARY)
 # - CUDF -------------------------------------------------------------------------------------------
 
 find_path(CUDF_INCLUDE "cudf"
-          HINTS "$ENV{CUDF_ROOT}/cpp/include")
+          HINTS "$ENV{CUDF_ROOT}/include")
```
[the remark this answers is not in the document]
trxcllnt: From what I know about `CUDF_ROOT` usage, this path should have never worked. I'll try to build this branch in the devel container to make sure though.

python/cuspatial/setup.py
rlratzel: If there's a convention in place then I'm fine with no comment.

harrism:
Does cuSpatial need libcu++?

cpp/tests/CMakeLists.txt
```diff
@@ -65,29 +65,29 @@ include_directories("${CMAKE_BINARY_DIR}/include"
...
+                    "${GDAL_INCLUDE}"
                     "${GTEST_INCLUDE_DIR}"
                     "${RMM_INCLUDE}"
                     "${CUDF_INCLUDE}"
                     "${CUDF_SRC_INCLUDE}"
                     "${CUDF_TEST_INCLUDE}"
-                    "${CUB_INCLUDE}"
+                    "${CUB_INCLUDE_DIR}"
+                    "${THRUST_INCLUDE_DIR}"
+                    "${LIBCUDACXX_INCLUDE_DIR}"
```
harrism, line 76: Do cuspatial tests really need libcudacxx?

cpp/tests/CMakeLists.txt
```diff
@@ -65,29 +65,29 @@ include_directories("${CMAKE_BINARY_DIR}/include"
...
-                    "${CMAKE_SOURCE_DIR}/../thirdparty/libcudacxx/include"
-                    "${CMAKE_SOURCE_DIR}/thirdparty/dlpack/include"
+                    "${GDAL_INCLUDE}"
                     "${GTEST_INCLUDE_DIR}"
                     "${RMM_INCLUDE}"
                     "${CUDF_INCLUDE}"
                     "${CUDF_SRC_INCLUDE}"
                     "${CUDF_TEST_INCLUDE}"
-                    "${CUB_INCLUDE}"
+                    "${CUB_INCLUDE_DIR}"
```
harrism, line 74: CUB and Thrust should be included before the CUDA toolkit include path to avoid getting the CUDA toolkit version of them. See how this is done in libcudf cmake.

cpp/CMakeLists.txt
```diff
@@ -261,14 +287,13 @@ endif(CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES)
...
-                    "${CMAKE_SOURCE_DIR}/../thirdparty/cub"
-                    "${CMAKE_SOURCE_DIR}/../thirdparty/libcudacxx/include"
-                    "${GDAL_INCLUDE_DIRS}"
+                    "${GDAL_INCLUDE}"
                     "${RMM_INCLUDE}"
                     "${CUDF_SRC_INCLUDE}"
                     "${CUDF_INCLUDE}"
-                    "${CUB_INCLUDE}"
-                    "${CUDACXX_INCLUDE}")
+                    "${CUB_INCLUDE_DIR}"
```
harrism, line 294: All 3 of these are in the CUDA toolkit as of CUDA 11 so need to be moved above the CUDA toolkit include paths (which need to be explicitly added). See libcudf cmake.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -43,17 +43,16 @@ include_directories("${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}"
...
-                    "${CMAKE_SOURCE_DIR}/../thirdparty/libcudacxx/include"
+                    "${GDAL_INCLUDE}"
                     "${GTEST_INCLUDE_DIR}"
                     "${GBENCH_INCLUDE_DIR}"
                     "${RMM_INCLUDE}"
                     "${CUDF_INCLUDE}"
                     "${CUDF_SRC_INCLUDE}"
                     "${CUDF_TEST_INCLUDE}"
-                    "${CUB_INCLUDE}"
+                    "${CUB_INCLUDE_DIR}"
```
harrism, line 53: Before CUDA toolkit includes.

cpp/tests/CMakeLists.txt
trxcllnt: I believe we do, because we import `cudf/types.hpp`

cpp/CMakeLists.txt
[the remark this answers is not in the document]
rlratzel: marking as resolved assuming your `devel` build worked

cpp/tests/CMakeLists.txt
harrism: And I remembered that we use timestamps.

cpp/CMakeLists.txt
```diff
@@ -183,22 +198,19 @@ endif (RMM_INCLUDE AND RMM_LIBRARY)
...
+          HINTS "$ENV{CUDF_ROOT}/include")
 
 find_library(CUDF_LIBRARY "cudf"
              HINTS "$ENV{CUDF_ROOT}/lib")
 
-SET(CUDF_SRC_INCLUDE "$ENV{CUDF_HOME}/cpp/src")
-SET(CUB_INCLUDE "$ENV{CUDF_HOME}/cpp/thirdparty/cub")
-SET(CUDACXX_INCLUDE "$ENV{CUDF_HOME}/cpp/thirdparty/libcudacxx/include")
-SET(CUDF_TEST_INCLUDE "$ENV{CUDF_HOME}/cpp/")
+set(CUDF_TEST_INCLUDE "$ENV{CUDF_HOME}/cpp")
```
kkraus14, line 206: Reaching into the libcudf source like this makes me cautious. Are these headers not shipped in a libcudf install?

cpp/CMakeLists.txt
```diff
@@ -326,14 +337,24 @@ endif(USE_NVTX)
...
+        DESTINATION include
+        COMPONENT cuspatial)
+
+install(DIRECTORY ${CMAKE_BINARY_DIR}/include/libcxx
+        DESTINATION include/libcuspatial
+        COMPONENT cuspatial)
+
+install(DIRECTORY ${CMAKE_BINARY_DIR}/include/libcudacxx
+        DESTINATION include/libcuspatial
+        COMPONENT cuspatial)
```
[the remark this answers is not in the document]
trxcllnt: We don't expose them explicitly, but the Cython doesn't compile unless we have them.

cpp/CMakeLists.txt
trxcllnt: They're not, unfortunately. AFAIK cuSpatial is the only external consumer of `libcudftestutil.a` so far.

PR #244, opened 2020-06-15
Pin gdal version

raydouglass:
Need to restrict the gdal version more closely since 3.0.2 and 3.1.0 are linked in an incompatible way. Since 3.1.0 works, just restrict to that minor version.


GPUtester:

raydouglass:
rerun tests

PR #246, opened 2020-06-18
[REVIEW] directed_hausdorff_distance benchmark and perf improvement.

cwharris:
This started as an experiment with Thrust and ended up being a lot of research involving prefix sums, noncommutative operations, and Thrust code diving. As it turns, Thrust does not yet have a reduce function which supports noncommutative operations (aka, `accumulate`), and given that Thrust needs to support multiple systems (sequential, cuda, tbb, etc), it is not straightforward to implement such a function. Given this limitation, and a desire to learn more about Thrust, I decided to use `inclusive_scan_by_key` in conjunction with a custom output iterator which can filter outputs based on a scattermap. This is overkill. In hindsight, it would have been far easier and more straightforward to write a kernel. The experience gained and information gathered along the way is still beneficial to other endeavors, however, and so the it was not a total waste of time.

While this implementation is faster than the original (need to generate this data again), it has a limitation due to a bug in Thrust. Because this operates in O(N^2) time using an iterator which calculates the distance between two points from two different spaces, it requires a Traversal type of `int64_t` to support `2^32` inputs, since  `2^32^2 = 2^64`. The bug blocks us from having this 64 bit iterator. Therefore the maximum number of points this algorithm supports (until Thrust's bug is fixed) is `46340` total. Not per space. However, computing anything higher than this will take several seconds and may better suited to an `n log n` time solution using spatial partition.


GPUtester:

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,178 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required point_b_y applicable law or agreed to in writing, software
```
trxcllnt, line 10:
```suggestion
 * Unless required by applicable law or agreed to in writing, software
```

cpp/src/utility/scatter_output_iterator.cuh
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required point_b_y applicable law or agreed to in writing, software
```
trxcllnt, line 10:
```suggestion
 * Unless required by applicable law or agreed to in writing, software
```

cpp/src/utility/size_from_offsets.cuh
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required point_b_y applicable law or agreed to in writing, software
```
trxcllnt, line 10:
```suggestion
 * Unless required by applicable law or agreed to in writing, software
```

cpp/include/cuspatial/detail/hausdorff.cuh
cwharris: Intellisense fails on me often enough that I've started using find replace, and apparently that's _not_ gonna work. >.<

cwharris:
Current implementation:
```
-----------------------------------------------------------------------------------------
Benchmark                                               Time             CPU   Iterations
-----------------------------------------------------------------------------------------
HausdorffBenchmark/hausdorff/128/manual_time          607 ms          607 ms            1
HausdorffBenchmark/hausdorff/512/manual_time         2183 ms         2183 ms            1
HausdorffBenchmark/hausdorff/4096/manual_time       17240 ms        17240 ms            1
HausdorffBenchmark/hausdorff/32768/manual_time     141208 ms       141206 ms            1
```
Proposed implementation
```
-----------------------------------------------------------------------------------------
Benchmark                                               Time             CPU   Iterations
-----------------------------------------------------------------------------------------
HausdorffBenchmark/hausdorff/128/manual_time        0.084 ms        0.103 ms         7141
HausdorffBenchmark/hausdorff/512/manual_time        0.663 ms        0.684 ms         1044
HausdorffBenchmark/hausdorff/4096/manual_time        26.8 ms         26.9 ms           24
HausdorffBenchmark/hausdorff/32768/manual_time       1694 ms         1694 ms            1
```

cwharris:
I had some rebase issues before creating this PR, so I'm addressing those now.

cwharris:
I played around with a version of this that doesn't use a pre-computed lookup list, thereby avoiding a temporary memory allocation of `sizeof(int32_t) * num_points`. Here are the results.

The first number is the total number of spaces.
The second number is the total number of points.
If there are not enough points for each space to have at least one point, then less spaces are used.

```
----------------------------------------------------------------------------------------------
Benchmark                                                    Time             CPU   Iterations
----------------------------------------------------------------------------------------------
HausdorffBenchmark/hausdorff/128/128/manual_time         0.056 ms        0.073 ms        12524
HausdorffBenchmark/hausdorff/512/128/manual_time         0.055 ms        0.073 ms        12449
HausdorffBenchmark/hausdorff/4096/128/manual_time        0.055 ms        0.073 ms        12521
HausdorffBenchmark/hausdorff/8192/128/manual_time        0.055 ms        0.073 ms        12403
HausdorffBenchmark/hausdorff/128/512/manual_time         0.452 ms        0.471 ms         1546
HausdorffBenchmark/hausdorff/512/512/manual_time         0.486 ms        0.501 ms         1438
HausdorffBenchmark/hausdorff/4096/512/manual_time        0.486 ms        0.501 ms         1440
HausdorffBenchmark/hausdorff/8192/512/manual_time        0.486 ms        0.501 ms         1434
HausdorffBenchmark/hausdorff/128/4096/manual_time         14.9 ms         14.9 ms           46
HausdorffBenchmark/hausdorff/512/4096/manual_time         16.9 ms         16.9 ms           41
HausdorffBenchmark/hausdorff/4096/4096/manual_time        19.0 ms         19.0 ms           37
HausdorffBenchmark/hausdorff/8192/4096/manual_time        19.0 ms         19.0 ms           37
HausdorffBenchmark/hausdorff/128/32768/manual_time         962 ms          962 ms            1
HausdorffBenchmark/hausdorff/512/32768/manual_time        1013 ms         1013 ms            1
HausdorffBenchmark/hausdorff/4096/32768/manual_time       1413 ms         1413 ms            1
HausdorffBenchmark/hausdorff/8192/32768/manual_time       1351 ms         1351 ms            1
```

In the worst case (8k spaces, 32k points), we're .5s slower. For lower numbers of spaces, the performance is better (this makes sense because we have less spaces to binary search).

Given that the advantage of this `O(n^2)` algorith is that it _can_ be done with no temporary memory allocations, and that the performance is similar for lower numbers of spaces, I'm inclined to advocate for the binary search in the long run. In the future we can have an `O(n log n)` algorithm which will be significantly faster than the `O(n^2)`.

Thoughts?

cpp/src/spatial/hausdorff.cu
```diff
@@ -103,49 +135,92 @@ struct hausdorff_functor {
...
-    auto result = cudf::make_fixed_width_column(cudf::data_type{tid},
-                                                points_per_space.size() * points_per_space.size(),
-                                                cudf::mask_state::UNALLOCATED,
-                                                stream,
-                                                mr);
+    size_type num_points  = xs.size();
+    size_type num_spaces  = space_offsets.size();
+    size_type num_results = num_spaces * num_spaces;
+
+    if (num_results == 0) { return make_column<T>(0, stream, mr); }
```
harrism, line 146: Can't you use `make_empty_column` from column_factories.hpp?

cpp/src/spatial/hausdorff.cu
```diff
@@ -103,49 +135,92 @@ struct hausdorff_functor {
...
 
-    if (result->size() == 0) { return result; }
+    // ===== Make Lookup for Space by Point ========================================================
+    // these space lookups could be replaced with a `lower_bound` to reduce temporary memory
 
-    auto space_offsets = rmm::device_vector<cudf::size_type>(points_per_space.size());
+    auto temp_space_lookup = rmm::device_vector<size_type>(num_points);
```
harrism, line 151:
```suggestion
    auto temp_space_lookup = rmm::device_uvector<size_type>(num_points, stream);
```

cpp/src/spatial/hausdorff.cu
```diff
@@ -103,49 +135,92 @@ struct hausdorff_functor {
...
+
+    thrust::scatter(rmm::exec_policy(stream)->on(stream),
```
harrism, line 153: Add a comment explaining what is happening.

cpp/src/spatial/hausdorff.cu
```diff
@@ -103,49 +135,92 @@ struct hausdorff_functor {
...
+    auto space_offset_iterator =
+      thrust::make_transform_iterator(thrust::make_counting_iterator<int32_t>(0),
+                                      size_from_offsets_functor{*d_space_offsets, xs.size()});
+
+    // ===== Make Hausdorff Accumulator ============================================================
 
-    // utilize one block per result (pair of spaces).
-    size_type num_blocks_x = min(result->size(), MAX_NUM_BLOCKS_X);
-    size_type num_blocks_y = ceil(result->size() / (float)MAX_NUM_BLOCKS_X);
+    auto num_distances = static_cast<int32_t>(num_points) * static_cast<int32_t>(num_points);
```
harrism, line 174: Why do you need these casts?

cpp/src/spatial/hausdorff.cu
```diff
@@ -103,49 +135,92 @@ struct hausdorff_functor {
...
+
+    auto scatter_map = thrust::make_transform_iterator(
+      hausdorff_index_iter, [] __device__(hausdorff_index acc) { return acc.result_idx; });
+
+    // the following output iterator and `inclusive_scan_by_key` could be replaced by a
+    // reduce_by_key, if it supported noncommutative operators.
+
+    auto scatter_out = make_scatter_output_iterator(result_temp_iter, scatter_map);
+
+    thrust::inclusive_scan_by_key(rmm::exec_policy(stream)->on(stream),
```
harrism, line 205: Add a comment explaining what is happening.

cpp/src/spatial/hausdorff.cu
```diff
@@ -103,49 +135,92 @@ struct hausdorff_functor {
...
+                                  hausdorff_acc_iter,
+                                  hausdorff_acc_iter + num_distances,
+                                  hausdorff_acc_iter,
+                                  scatter_out,
+                                  hausdorff_key_compare<T>{});
+
+    thrust::transform(rmm::exec_policy(stream)->on(stream),
```
harrism, line 212: This cast can't be fused with the above scan?

cpp/src/utility/scatter_output_iterator.cuh
```diff
@@ -0,0 +1,102 @@
...
+#include <thrust/iterator/iterator_adaptor.h>
+
+#pragma once
+
+namespace cuspatial {
+namespace detail {
+
+/** @brief scatters outputs to a given index, according to the input scatter map.
+ *
+ * @note: if the scatter map produces a negative index, the corrosponding output is ignored.
```
harrism, line 27:
```suggestion
 * @note: if the scatter map produces a negative index, the corresponding output is ignored.
```

cpp/src/utility/scatter_output_iterator.cuh
```diff
@@ -0,0 +1,102 @@
...
+                                                    const OutputIterator& out,
+                                                    const ScatterIterator& scatter_map)
+    : begin(begin), out(out), scatter_map(scatter_map)
+  {
+  }
+
+  template <typename T>
+  __host__ __device__ scatter_output_iterator_proxy operator=(const T& element)
+  {
+    auto scatter_idx = *(scatter_map + thrust::distance(begin, out));
```
harrism, line 42: What's happening here and why? I think more documentation would help.

cpp/src/utility/scatter_output_iterator.cuh
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+ private:
+  OutputIterator begin;
+  OutputIterator out;
+  ScatterIterator scatter_map;
+};
+
+template <typename OutputIterator, typename ScatterIterator>
+class scatter_output_iterator;
```
harrism, line 56: Wow there are lots of classes, structs, proxies. Please document, I'm confused.

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,144 @@
...
+        out.max = std::max(out.max, lhs.min_r);
+      }
+
+      // roll the RHS inner minimum into output (output lhs, rhs, or max)
+      if (matching_r) {
+        out.min_r = std::min(out.min_r, rhs.min_l);
+      } else if (matching_m) {
+        out.min_l = std::min(out.min_l, rhs.min_l);
+      } else {
+        out.max = std::max(out.max, rhs.min_l);
```
trxcllnt, line 98:
```suggestion
        out.max = max(out.max, rhs.min_l);
```

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,144 @@
...
+   * @returns Directed hausdorff distance
+   */
+  __host__ __device__ explicit operator T() const
+  {
+    auto is_open = this->col_l == this->col_r;
+
+    auto partial_max =
+      is_open ? std::min(this->min_l, this->min_r) : std::max(this->min_l, this->min_r);
+
+    return std::max(this->max, partial_max);
```
trxcllnt, line 117:
```suggestion
    return max(this->max, partial_max);
```

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,144 @@
...
+  {
+    auto const& lhs = *this;
+
+    auto out = hausdorff_acc<T>{lhs.key,
+                                rhs.result_idx,
+                                lhs.col_l,
+                                rhs.col_r,
+                                lhs.min_l,
+                                rhs.min_r,
+                                std::max(lhs.max, rhs.max)};
```
trxcllnt, line 73:
```suggestion
                                max(lhs.max, rhs.max)};
```

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,144 @@
...
+* auto d4 = hausdorff_acc<float>(..., 2, distance_21);
+* auto d5 = hausdorff_acc<float>(..., 3, distance_22);
+* auto d6 = hausdorff_acc<float>(..., 3, distance_23);
+*
+* auto distance_2_to_3 = static_cast<float>(d1 + d2 + d3 + d4 + d5 + d6);
+* ```
+*/
+template <typename T>
+struct hausdorff_acc {
+  __host__ __device__ hausdorff_acc<T> operator+(hausdorff_acc<T> const& rhs) const
```
trxcllnt, line 63: Does this need to be `__host__ __device__` or can it be `__device__`?

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,144 @@
...
+
+    return out;
+  }
+
+  /**
+   * @brief Converts hausdorff reduction to directed hausdorff distance
+   *
+   * @returns Directed hausdorff distance
+   */
+  __host__ __device__ explicit operator T() const
```
trxcllnt, line 110:
```suggestion
  __device__ explicit operator T() const
```

trxcllnt:
Does this PR supersede https://github.com/rapidsai/cuspatial/pull/211? If so, can you close #211?

cpp/benchmarks/hausdorff_benchmark.cpp
```diff
@@ -0,0 +1,63 @@
...
+static void BM_hausdorff(benchmark::State& state)
+{
+  int32_t num_points           = state.range(0);
+  int32_t num_spaces           = std::min(num_points, 8);
+  int32_t num_points_per_space = num_points / num_spaces;
+
+  auto counting_iter = thrust::counting_iterator<int32_t>();
+  auto zero_iter     = thrust::make_transform_iterator(counting_iter, [](auto idx) { return 0; });
+
+  auto space_offset_iter = thrust::make_transform_iterator(
```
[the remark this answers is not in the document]
cwharris: I believe Thrust is intended to support this use case, and I have not had any issues building. :man_shrugging:

cpp/include/cuspatial/detail/hausdorff.cuh
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+* @brief Hausdorff reduction data structure
+*
+* Data structure for computing directed hausdorff distance as a noncommutative reduction.
+
+* Given one `hausdorff_acc<T>` for each distance between all points in two spaces (O(N^2)) and a
+* binary reduce algorithm which supports noncommutative operations, this data structure and
+* accompoanying reduce operator (`+`) can be used to calculate the directed hausdorff distance
+* between those two spaces. One of two asymetric directed distances can be deduced. The asymetric
```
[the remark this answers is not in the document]
cwharris: Ah, turns out spell checker was not enabled for cuda files. fixed.

cpp/src/spatial/hausdorff.cu
cwharris: https://github.com/thrust/thrust/issues/1199

Haven't tested with latest Thrust yet, but it does not work with `1.9.10`.

cpp/src/utility/scatter_output_iterator.cuh
cwharris: Attempted to address this concern in latest commits, but ultimately it's the same iterator proxy pattern used by Thrust. At first I didn't understand it either. :man_shrugging:

cpp/src/spatial/hausdorff.cu
cwharris: I removed this entirely in favor of an `upper_bound`, which is called once per point to determine which to which space that point belongs. The performance of that call scales with the number of spaces, not the number of points.

cwharris: The advantage of the `upper_bound` approach is that it does not require an extra `num_points * sizeof(int32_t)` bytes of memory. Since this O(n^2) approach's advantage is reduced memory usage, and not necessarily performance, I thought it appropriate to reduce the memory usage further. For 32768 points divided up in to 16384 spaces, the performance is comparable (presumably due to the overhead of the additional memory allocation and two kernel launches required by the previous implementation).

cpp/src/spatial/hausdorff.cu
cwharris: Don't need them. Leftover from testing. Thanks for catching it. :smiley:

cwharris:
Latest benchmarks.
```
---------------------------------------------------------------------------------------------------------------
Benchmark                                                     Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------------------------------
HausdorffBenchmark/hausdorff/1024/1024/manual_time        0.801 ms        0.818 ms          869 items_per_second=1.30675G/s
HausdorffBenchmark/hausdorff/4096/1024/manual_time        0.803 ms        0.820 ms          862 items_per_second=1.30263G/s
HausdorffBenchmark/hausdorff/16384/1024/manual_time       0.804 ms        0.821 ms          854 items_per_second=1.30151G/s
HausdorffBenchmark/hausdorff/1024/4096/manual_time         10.8 ms         10.8 ms           63 items_per_second=1.54981G/s
HausdorffBenchmark/hausdorff/4096/4096/manual_time         12.3 ms         12.3 ms           56 items_per_second=1.36443G/s
HausdorffBenchmark/hausdorff/16384/4096/manual_time        12.3 ms         12.3 ms           56 items_per_second=1.36471G/s
HausdorffBenchmark/hausdorff/1024/32768/manual_time         728 ms          728 ms            1 items_per_second=1.475G/s
HausdorffBenchmark/hausdorff/4096/32768/manual_time         790 ms          790 ms            1 items_per_second=1.35852G/s
HausdorffBenchmark/hausdorff/16384/32768/manual_time        770 ms          770 ms            1 items_per_second=1.39406G/s
```

Previously:
```
HausdorffBenchmark/hausdorff/8192/32768/manual_time       1351 ms         1351 ms            1
```

cwharris:
rerun tests

cpp/src/spatial/hausdorff.cu
harrism: Nice. I love upper_bound/lower_bound.

cwharris:
Had to force push because I rebased due to the tests failing (conda picking up the wrong version of cudf).

PR #250, opened 2020-07-01
[REVIEW] Cartesian Product Iterator

cwharris:
Factored out the Cartesian product iterator used by the Hausdorff distance implementation so that it can be used in st distance, and other n^2 distance metrics. Also added the ability to iterate the Cartesian product of two inputs, rather than a single input against itself. One of the goals was to avoid impacting Hausdorff performance. Instead of reaching that goal, Hausdorff performance was increased by 40%. :tada:

```
---------------------------------------------------------------------------------------------------------------
Benchmark                                                     Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------------------------------------------------
HausdorffBenchmark/hausdorff/1024/1024/manual_time        0.681 ms        0.698 ms         1022 items_per_second=1.53716G/s
HausdorffBenchmark/hausdorff/4096/1024/manual_time        0.679 ms        0.696 ms         1024 items_per_second=1.54078G/s
HausdorffBenchmark/hausdorff/16384/1024/manual_time       0.681 ms        0.698 ms         1015 items_per_second=1.53786G/s
HausdorffBenchmark/hausdorff/1024/4096/manual_time         7.33 ms         7.34 ms           95 items_per_second=2.28901G/s
HausdorffBenchmark/hausdorff/4096/4096/manual_time         9.98 ms         10.0 ms           67 items_per_second=1.67971G/s
HausdorffBenchmark/hausdorff/16384/4096/manual_time        9.97 ms         9.99 ms           68 items_per_second=1.68188G/s
HausdorffBenchmark/hausdorff/1024/32768/manual_time         531 ms          531 ms            1 items_per_second=2.02367G/s
HausdorffBenchmark/hausdorff/4096/32768/manual_time         524 ms          524 ms            1 items_per_second=2.04759G/s
HausdorffBenchmark/hausdorff/16384/32768/manual_time        538 ms          539 ms            1 items_per_second=1.99397G/s
```

Issue #231, opened 2020-05-28
[FEA] Support for ST_Distance and ST_contains 


jonathanflack:
We need to quickly calculate the distance between two points in geospatial for high volume network traffic where the geoIP data is known.  Additionally looking to reduce result data according to arbitrary selections using standard selections.  Optimized for geoPoint would be ideal.

ST_contains for arbitrary lasso, rectangle or circular selections in geospatial. Should return boolean.  See OpenGIS Simple Features Implementation Specification for SQL 1.1. s2.1.1.2 // s2.1.13.3 - same as within(geometry B, geometry A)

ST_distance returns the minimum 2D Cartesian (planar) distance between two geometries, in projected units (spatial ref units).  See OpenGIS Simple Features Implementation Specification for SQL 1.1.

These are standard geospatial operators, so no alternatives have been considered.

No additional context.

https://github.com/rapidsai/cuspatial/issues/231#tasklist-block-1c9f6f44-5086-4546-87bc-064bcee1b229

thomcom:
Thanks a lot for filing an issue! I'd love to implement both of these features in the near future. Our team will also be talking about it at our next meeting.

jonathanflack:
Thanks!  This is very important for us in cyber, as we pivot a lot of analysis around geo dimensions.  This would help accelerate that workflow dramatically, as well as any inference based on analytic results based on this refinement.  It also offers an opportunity to potentially enrich network data with geospatial attributes at line rate for us during ETL.

Jonathan Flack, Principal Systems Architect
BROADBRIDGE NETWORKS
www.broadbridgenetworks.com • jf@broadbridgenetworks.net
[m] +1 310-359-3510

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

harrism:
Hi @jonathanflack, a series of questions:

1. Is this the ST_Distance you mean? https://postgis.net/docs/ST_Distance.html ST_Contains: https://postgis.net/docs/ST_Distance.html

2. Are ST_Distance and ST_contains independently useful to you? Or do you need them both before you can start?  (This is a prioritization question, so we can split up the work).
https://postgis.net/docs/ST_Distance.html

3. For ST_Distance, I see that it supports "geometry" and "geography". Is one more important than the other? 

4. For ST_Contains, there are various geometry combinations to support (we already have parallel point-in-polygon APIs). Do you have specific priorities?

Thanks!
Mark

jonathanflack:
[...]

PR #251, opened 2020-07-07
[REVIEW] Directed Polygon Point Distance

cwharris:
Contributes to #231. Depends on #250.


GPUtester:

cpp/include/cuspatial/polygon_distance.hpp
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+#include <cudf/types.hpp>
+
+#include <memory>
+
+namespace cuspatial {
+
+/**
+ * @brief calculates minimum segment-point distance between all shapes
```
harrism, line 26:
```suggestion
 * @brief calculates minimum segment-to-point distance between all shapes
```

harrism: And should it be "polygons" rather than "shapes", since that is specifically what this function supports? And "Euclidean distance" instead of just "distance"?

cpp/src/spatial/polygon_distance.cu
```diff
@@ -0,0 +1,161 @@
...
+
+namespace cuspatial {
+namespace detail {
+namespace {
+
+/**
+ * @brief Calculates segment-point distance
+ *
+ * Given a `cartesian_product_group_index` and two columns representing `x` and `y` coordinates,
+ * calculates the segment-point distance of a line segment in group `a` and a point in group `b`.
```
harrism, line 40:
```suggestion
 * calculates the segment-to-point distance of a line segment in group `a` and a point in group `b`.
```

cpp/src/spatial/polygon_distance.cu
```diff
@@ -0,0 +1,161 @@
...
+#include <memory>
+
+using size_type = cudf::size_type;
+
+namespace cuspatial {
+namespace detail {
+namespace {
+
+/**
+ * @brief Calculates segment-point distance
```
harrism, line 37:
```suggestion
 * @brief Calculates segment-to-point distance
```

cpp/src/spatial/polygon_distance.cu
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * Given a `cartesian_product_group_index` and two columns representing `x` and `y` coordinates,
+ * calculates the segment-point distance of a line segment in group `a` and a point in group `b`.
+ *
+ * If group `a` contains two or more points, calculate segment-point distance.
```
harrism, line 42:
```suggestion
 * If group `a` contains two or more points, calculate segment-to-point distance.
```

cpp/src/spatial/polygon_distance.cu
```diff
@@ -0,0 +1,161 @@
...
+ * If group `a` contains only a single point, calculate point-point distance.
```
harrism, line 43:
```suggestion
 * If group `a` contains only a single point, calculate point-to-point distance.
```

cpp/src/spatial/polygon_distance.cu
```diff
@@ -0,0 +1,161 @@
...
+    auto const edge_y   = ys.element<T>(a_idx_1) - origin_y;
+    auto const point_x  = xs.element<T>(b_idx_0) - origin_x;
+    auto const point_y  = ys.element<T>(b_idx_0) - origin_y;
+
+    auto const magnitude = edge_x * edge_x + edge_y * edge_y;
+    auto const travel    = point_x * edge_x + point_y * edge_y;
+
+    // if point is projected within line segment bounds, use segment-point distance.
+    // if point is projected outside line segment bounds, use point-point distance.
+    if (0 < travel && travel < magnitude) {  // 0 < travel < edge_length
```
harrism, line 68:
```suggestion
    if (0 < travel && travel < magnitude) {
```

cpp/src/spatial/polygon_distance.cu
```diff
@@ -0,0 +1,161 @@
...
+                          thrust::equal_to<thrust::pair<int32_t, int32_t>>(),
+                          thrust::minimum<T>());
+
+    return result;
+  }
+};
+
+}  // namespace
+}  // namespace detail
+
```
harrism, line 141: Typically we have a detail version of API functions that take a stream, like we do in libcudf, don't we?

cpp/tests/spatial/polygon_distance_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+TYPED_TEST(DirectedPolygonDistanceUnsupportedTypesTest, InvalidTypeTest)
+{
+  using T = TypeParam;
+
+  auto x             = fixed_width_column_wrapper<T>({0, 0});
+  auto y             = fixed_width_column_wrapper<T>({0, 0});
+  auto space_offsets = fixed_width_column_wrapper<size_type>({0});
+
+  EXPECT_THROW(cuspatial::directed_polygon_distance(x, y, space_offsets), cuspatial::logic_error);
+}
```
harrism, line 152: I know this is the perennial discussion, but we should have some tests of non-trivial size (large enough to launch multiple thread blocks).   You could generate a thousand squares on a regular grid (for example) which would make it easy to generate a reference solution analytically.

Or maybe we should look at having gtests run a Python script to generate input and reference data using another library (if that's easier).

cpp/include/cuspatial/polygon_distance.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * between points i and j.
+ *
+ *
+ * @param[in] xs: x component of points
+ * @param[in] ys: y component of points
+ * @param[in] offsets: number of points in each space
+ * @param[in] mr: Device memory resource used to allocate the returned memory
+ * @return std::unique_ptr<cudf::column>
+ */
+std::unique_ptr<cudf::column> directed_polygon_distance(
```
harrism, line 39: Why is this a directed distance? Suggest documenting that.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -74,26 +78,92 @@ def directed_hausdorff_distance(xs, ys, points_per_space):
...
+        column of y-coordinates
+    offsets
+        number of points in each space
+
+    Returns
+    -------
+    result : cudf.DataFrame
+        The pairwise directed distance matrix with one row and one column per
+        input polygon; the value at row `i`, column `j` represents the
+        polygon-point distance from polygon i to any vertex in polygon j.
```
harrism, line 119: ?

```suggestion
        minimum polygon-point distance from polygon i to any vertex in polygon j.
```

harrism: From which point in polygon i is the distance measured? Is it really "the minimum distance from any vertex in i to any vertex in j"?

If so, why would the distance from i to j be different from the distance from j to i?

cpp/include/cuspatial/polygon_distance.hpp
cwharris: The function supports individual points and/or segments, as well as polygons. Euclidean distance sounds good to me. :+1:

cpp/src/spatial/polygon_distance.cu
cwharris: Hah. I was originally using `travel * (magnitude - travel) >= 0`, which is why the comment was there. Thanks! :laughing:

cpp/src/spatial/polygon_distance.cu
cwharris: Probably. I'll fix this PR and then go take ensure it's being done elsewhere in a followup (if needed).

cpp/tests/spatial/polygon_distance_test.cpp
cwharris: Opened an issue for good measure. I'll think on how we can approach this.

https://github.com/rapidsai/cuspatial/issues/259

python/cuspatial/cuspatial/core/gis.py
cwharris: If the question is "why would..." the answer is: because it was more computationally efficient and required less redundant reads.
If the question had been "why _should_..." the answer is not so simple.

That said, this is the pattern that Hausdorff follows right now, and it's because its more efficient _on the computation side_, but might be less efficient on the consumers end, depending on how it's used. we should probably approach all/most distance metric APIs similarly, and Hausdorff set the precedence.

This requires further discussion.

harrism: No, the question was another way of asking "why is this a directed distance?". And the reason is because it's the distance from vertices of A to edges of B, not from edges of A to edges of B. If it were the latter, it would be a symmetric distance. This is just asking for clarity in the docs of the geometric reason for having two different distances per pair.

harrism:
@cwharris Should we push this to 0.16?

harrism:
Moving to 0.16.

cwharris:
I had intended to get this in to 0.15, but due to concerns about the API doing all-pairs comparisons for directed distance, I’m adjusting it to accept a pair of offsets and compute symmetric distance per pair. This requires a different iterator (which will be reusable for future Cartesian products of pairs of shapes/groups), and will take a little more time to implement. I’m planning on getting a draft of that API up today, and I’d like to get feedback not just for polygon distance, but for any pair-wise “ranged” comparisons.

isVoid:
I believe this is supercedes by https://github.com/rapidsai/cuspatial/pull/988

PR #252, opened 2020-07-07
[REVIEW] Fix CI style check failures

trxcllnt:
* Pin flake8, black, isort versions
* Update `ci/checks/style.sh` for new isort version

Similar to changes made in https://github.com/rapidsai/cudf/pull/5643

### DIFF
--- a/conda/environments/cuspatial_dev_cuda10.0.yml
+++ b/conda/environments/cuspatial_dev_cuda10.0.yml
@@ -6,7 +6,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black=3.7.7
+  - black=19.10
   - clang=8.0.1
   - clang-tools=8.0.1
   - cudf=0.14.*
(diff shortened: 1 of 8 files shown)


GPUtester:

conda/environments/cuspatial_dev_cuda10.0.yml
```diff
@@ -6,9 +6,12 @@ channels:
...
 dependencies:
+  - black=19.10
   - clang=8.0.1
   - clang-tools=8.0.1
   - cudf=0.14.*
   - cudatoolkit=10.0
+  - flake8=3.7.7
   - gdal>=3.0.2
   - geopandas=0.7.0
+  - isort=5.0.4
```
kkraus14, line 17: In the cudf update they didn't pin the versions of `black`, `flake8`, and `isort`. Should we update cudf similarly to this or unpin the versions of these?

I'd be happier unpinning the versions of these so we catch when there's updates to packages like these locally.

trxcllnt: @kkraus14 they weren't, but I suspect they should've been since the .precommit.yml versions are pinned. I agree with @jrhemstad's comment in https://github.com/rapidsai/cudf/issues/5647, we should pin to specific versions to avoid those projects breaking our local dev and CI.

PR #296, opened 2020-09-22
[REVIEW] Updates for RMM being header only

dantegd:
Fix for RMM being header only, also removes cmake option `CMAKE_CXX11_ABI` following other RAPIDS repos.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -53,15 +53,6 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     # Suppress parentheses warning which causes gmock to fail
     set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xcompiler -Wno-parentheses")
 
-    option(CMAKE_CXX11_ABI "Enable the GLIBCXX11 ABI" ON)
-    if(CMAKE_CXX11_ABI)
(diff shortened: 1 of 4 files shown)


trxcllnt (approved):

kkraus14 (approved):

cpp/CMakeLists.txt
```diff
@@ -338,7 +319,7 @@ endif(USE_NVTX)
 ###################################################################################################
 # - link libraries --------------------------------------------------------------------------------
 
-target_link_libraries(cuspatial cudf rmm cudart cuda cusparse nvrtc GDAL::GDAL)
+target_link_libraries(cuspatial cudf cudart cuda cusparse nvrtc GDAL::GDAL)
```
harrism, line 322: I think that you still want header-only libraries in `target_link_libraries`. CMake knows it's an interface only, and won't try to link librmm.

harrism (changes requested):

cpp/CMakeLists.txt
kkraus14: we haven't built an interface `rmm` target within the CMake here and we aren't doing a `find_package` or `add_subdirectory` here so there's no `rmm` target to use

harrism (approved):

Issue #315, opened 2020-10-26
[BUG] Remove deprecated from_gpu_matrix from cuSpatial `core/gis.py`


taureandyernv:
**Describe the bug**
Borrowed from @beckernick at https://github.com/rapidsai/cugraph/issues/1191 

`cudf.DataFrame.from_gpu_matrix` is deprecated and is removed in 0.17. We use this in two spots in the cuspatial codebase. We'll want to update these to directly use the DataFrame constructor instead of this method to accept device arrays.

https://github.com/rapidsai/cuspatial/blob/a06630547dbcdd710a590e936a8ba0bb589efd2d/python/cuspatial/cuspatial/core/gis.py#L94

https://github.com/rapidsai/cuspatial/blob/a06630547dbcdd710a590e936a8ba0bb589efd2d/python/cuspatial/cuspatial/core/gis.py#L252

taureandyernv:
Fixed in PR #332.  Thanks @kkraus14

PR #322, opened 2020-11-11
[REVIEW] Fix build issues related to libcudf/cudf changes

kkraus14:
Due to libcudf splitting into multiple shared libraries to fix debug builds, libcuspatial was left in a broken state. We don't fully understand why, but it seems like some combination of the linker pruning the `libcudf.so` dependency since it finds no symbols it uses in it BEFORE it adds `DT_NEEDED` entries from `libcudf.so` where it does find symbols.

Added some linker flags to temporarily fix the build while we continue investigating a more proper solution from the libcudf side.

Also fixed a minor gcc9 issue related to a copy elision.

Also fixed usage of a removed cudf Python API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - PR #294 Fix include of deprecated RMM header file.
 - PR #296 Updates for RMM being header only.
 - PR #298 Fix Python docs to render first argument of each public function.
+- PR #322 Fix build issues related to libcudf split build changes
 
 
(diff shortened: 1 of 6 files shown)


GPUtester:

harrism (approved):

cwharris (approved):

trxcllnt (approved):

taureandyernv:
for tracking, fixes #315

PR #353, opened 2021-02-12
Update make_counting_transform_iterator namespace in tests

trxcllnt:
Update location of `make_counting_transform_iterator` after https://github.com/rapidsai/cudf/pull/7306.

### DIFF
--- a/cpp/tests/trajectory/trajectory_utilities.cuh
+++ b/cpp/tests/trajectory/trajectory_utilities.cuh
@@ -47,8 +47,8 @@ std::unique_ptr<cudf::table> make_test_trajectories_table(
 
   auto rand_float = cudf::test::UniformRandomGenerator<T>{};
   auto ids_iter =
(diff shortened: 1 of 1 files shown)


harrism (commented):
@codereport see here. I don't like it that cuspatial now depends on cudf detail API. We should either promote this utility to a public one like it used to be in cudf::test, or reimplement it in cuspatial.

BTW, rapids-compose makes it easy to test impact of lib cudf changes on other libs.

codereport:
> @codereport see here. I don't like it that cuspatial now depends on cudf detail API. We should either promote this utility to a public one like it used to be in cudf::test, or reimplement it in cuspatial.
> 
> BTW, rapids-compose makes it easy to test impact of lib cudf changes on other libs.

I will open up a PR to add this to `Thrust`. That seems like the best solution. For now we can just have duplicates.

thomcom (approved):

trxcllnt:
@gpucibot merge

PR #362, opened 2021-02-25
Improve performance of quadtree point-to-polyline join 

trxcllnt:
Rewrites `quadtree_point_to_nearest_polyline` via thrust in the style of `quadtree_point_in_polygon` to get a 5x speed boost.

Benchmarked locally with NYC taxi 169M float32 points (RMM pool mode enabled):

```
quadtree_point_to_nearest_polyline (before)
1min ± 10.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

quadtree_point_to_nearest_polyline (after)
9.49 s ± 15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Timings for every routine in the benchmark after these changes:
```
quadtree_on_points
66.2 ms ± 64.3 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

polygon_bounding_boxes
322 µs ± 375 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

join_quadtree_and_bounding_boxes
11.4 ms ± 8.19 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

quadtree_point_in_polygon
5.48 s ± 54.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

polyline_bounding_boxes
258 µs ± 2.37 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

join_quadtree_and_bounding_boxes
11.5 ms ± 17.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

quadtree_point_to_nearest_polyline
9.49 s ± 15 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/cpp/src/join/quadtree_point_to_nearest_polyline.cu
+++ b/cpp/src/join/quadtree_point_to_nearest_polyline.cu
@@ -15,11 +15,11 @@
  */
 
 #include <indexing/construction/detail/utilities.cuh>
+#include <utility/point_to_nearest_polyline.cuh>
 
 #include <cuspatial/error.hpp>
 #include <cuspatial/spatial_join.hpp>
 
-#include <cudf/column/column.hpp>
 #include <cudf/column/column_device_view.cuh>
 #include <cudf/column/column_factories.hpp>
 #include <cudf/column/column_view.hpp>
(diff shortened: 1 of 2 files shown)


harrism (approved):

thomcom:
Mind iterating on this one's passing status when you get the chance @trxcllnt ?

trxcllnt:
@thomcom yeah for sure. We might also want to wait on this PR because, while it's faster, it requires allocating much more temporary memory than the current implementation. I _think_ it's possible to apply some of the techniques in the hausdorff iterator to eliminate that extra memory, but I'd need to explore it in depth.

trxcllnt:
@harrism I re-requested your review for the new _even better_ reimplementation of `quadtree_point_to_nearest_polyline.cu`

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -145,74 +157,156 @@ struct compute_quadtree_point_to_nearest_polyline {
...
+      make_zip_iterator(
+        quad_poly_indices.begin(), quad_point_offsets.begin(), quad_point_lengths.begin()));
+
+    // Sort the quad/polyline index pairs by quadrant so that the `point_indices` are
+    // produced/reduced in ascending order below.
+    thrust::stable_sort_by_key(
+      rmm::exec_policy(stream),
+      quad_point_offsets.begin(),
+      quad_point_offsets.end(),
+      make_zip_iterator(quad_point_lengths.begin(), quad_poly_indices.begin()));
```
trxcllnt, line 192: These intermediate allocations + sorting could be avoided if the output of `cuspatial::join_quadtree_and_bounding_boxes` were sorted by the `quad_offset` column.

trxcllnt:
Looks like cuDF headers now require C++17/gcc v9 now.

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

trxcllnt:
rerun tests

harrism:
> @harrism I re-requested your review for the new version of `quadtree_point_to_nearest_polyline.cu`

I totally missed this. I'm sorry.

cpp/tests/join/point_in_polygon_test_small.cpp
```diff
@@ -172,8 +172,8 @@ TYPED_TEST(PIPRefineTestSmall, TestSmall)
   expect_tables_equivalent(
     cudf::table_view{
       {fixed_width_column_wrapper<uint32_t>(
-         {0, 3, 1, 1, 1, 1, 1, 1, 1, 1, 1, 3, 3, 3, 3, 3, 3, 3, 3}),
+         {3, 3, 3, 3, 3, 3, 3, 3, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 3}),
```
harrism, line 175: How was the data for the tests generated? These could be very hard to update / maintain...

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -30,101 +30,118 @@
...
-          nearest_poly_id  = poly_idx;
-        }
-      }
-    }
-
-    out_point_index.element<uint32_t>(point_pos) = point_pos;
-    out_poly_index.element<uint32_t>(point_pos)  = nearest_poly_id;
-    out_distance.element<T>(point_pos)           = sqrt(distance_squared);
-  }
+  // Calculate the position in "point_offsets" that `global_index` falls between.
```
harrism, line 50: This function is very similar to `compute_poly_and_point_indices` functor in quadtree_point_in_polygon. Can you factor out a common utility?

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -30,101 +30,118 @@
...
+inline __device__ thrust::tuple<uint32_t, uint32_t> get_transposed_point_and_pair_index(
+  uint32_t const global_index,
+  uint32_t const *point_offsets,
+  uint32_t const *point_offsets_end,
+  QuadOffsetsIter quad_offsets,
+  QuadOffsetsIter quad_offsets_end,
+  QuadLengthsIter quad_lengths)
+{
+  uint32_t quad_poly_index{0}, local_point_offset{0};
+  thrust::tie(quad_poly_index, local_point_offset) =
```
harrism, line 94: Structured binding?

harrism: Or at least `std::tie`?

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -30,101 +30,118 @@
...
+    get_quad_poly_and_local_point_indices(global_index, point_offsets, point_offsets_end);
+
+  uint32_t local_poly_index{0}, num_polys_in_quad{0};
+  thrust::tie(local_poly_index, num_polys_in_quad) =
```
harrism, line 98: same

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -146,74 +163,178 @@ struct compute_quadtree_point_to_nearest_polyline {
...
+        thrust::make_transform_iterator(all_point_poly_indices_and_distances,
+                                        [] __device__(auto const &x) { return thrust::get<0>(x); });
+
+      // Allocate vectors for the distances min reduction
+      rmm::device_uvector<uint32_t> point_idxs(point_x.size(), stream);
+      rmm::device_uvector<uint32_t> poly_idxs(point_x.size(), stream);
+      rmm::device_uvector<T> distances(point_x.size(), stream);
+
+      // Fill distances with 0
+      thrust::fill(rmm::exec_policy(stream), distances.begin(), distances.end(), T{0});
```
harrism, line 270: First, does this  really need to be filled with zero? Doesn't the reduce_by_key write to every element of the output values? 

If it does need to be initialized, like all Thrust APIs that are not in the `thrust::async` namespace, this will synchronize. We have a utility in libcudf, but  it is a detail API... So here I would replace this with a `cudaMemsetAsync()` on `stream`.  Since it is only setting to 0, you can use a memset.

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -146,74 +163,178 @@ struct compute_quadtree_point_to_nearest_polyline {
...
+                               auto const &i_rhs = thrust::get<1>(rhs);
+                               return i_lhs < i_rhs ? lhs : rhs;
+                             }
+                             // Otherwise choose poly with smallest distance
+                             return d_lhs < d_rhs ? lhs : rhs;
+                           })
+                           .first);
+
+      point_idxs.resize(num_distances, stream);
+      point_idxs.shrink_to_fit(stream);
```
harrism, line 309: shrink_to_fit does a copy... Just something to be aware of in terms of performance. Potential optimization if we ever implement efficient virtual memory based realloc in RMM...  Or, if we find it's a problem and the memory overhead is not, could elide this.

harrism (changes requested):
Looks great. Performance looks great. Several comments, but only one absolutely required change (missing stream sync).

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -111,63 +111,74 @@ struct compute_quadtree_point_in_polygon {
...
+                             quad_lengths_iter,
+                             quad_lengths_iter + num_poly_quad_pairs,
+                             local_point_offsets.begin() + 1);
+
+      // Ensure local point offsets starts at 0
+      uint32_t init{0};
+      local_point_offsets.set_element_async(0, init, stream);
+
+      // The last element is the total number of points to test against any polygon.
+      auto num_total_points = local_point_offsets.back_element(stream);
```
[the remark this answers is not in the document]
trxcllnt: Doesn't `element()` synchronize the stream [here](https://github.com/rapidsai/rmm/blob/branch-0.20/include/rmm/device_uvector.hpp#L273) for us?

cpp/src/join/quadtree_point_to_nearest_polyline.cu
trxcllnt: Yep, this is a space/time trade. `num_distances` will be smaller than `point_idxs` if all the polylines are in the same spatial quadrant. `point_idxs` is large enough to contain indices for every point in the quadtree, so shrinking here allows us to reuse this memory for the output columns.

cpp/tests/join/point_in_polygon_test_small.cpp
trxcllnt: This was part of the original PR, I assume generated manually.

I changed `cuspatial::join_quadtree_and_bboxes` to sort its output because it's needed by point-to-nearest-polyline, and was thinking predictable sorted output would be more useful for end-users too.

trxcllnt:
rerun tests

cpp/src/join/quadtree_point_to_nearest_polyline.cu
trxcllnt: Done in https://github.com/rapidsai/cuspatial/pull/362/commits/1bc89b4eab9d7033b6177097b45981939e794865.

cpp/src/join/quadtree_point_to_nearest_polyline.cu
```diff
@@ -146,74 +144,171 @@ struct compute_quadtree_point_to_nearest_polyline {
...
+        thrust::make_transform_iterator(all_point_poly_indices_and_distances,
+                                        [] __device__(auto const &x) { return thrust::get<0>(x); });
+
+      // Allocate vectors for the distances min reduction
+      rmm::device_uvector<uint32_t> point_idxs(point_x.size(), stream);
+      rmm::device_uvector<uint32_t> poly_idxs(point_x.size(), stream);
+      rmm::device_uvector<T> distances(point_x.size(), stream);
+
+      // Fill distances with 0
+      CUDA_TRY(cudaMemsetAsync(distances.data(), 0, distances.size(), stream.value()));
```
harrism, line 251: Memset sets bytes, not words, so you need to specify the size in bytes. Otherwise unless T is a single-byte type, you are only setting part of the array...

```suggestion
      CUDA_TRY(cudaMemsetAsync(distances.data(), 0, distances.size() * sizeof(T), stream.value()));
```

harrism (changes requested):
Just a couple more fixes.

harrism (approved):

trxcllnt:
@gpucibot merge

cwharris:
I had questions regarding how the tests were updated due to the new sortedness of the outputs, and @trxcllnt answered those offline. lgtm.

PR #367, opened 2021-03-10
Eliminate literals passed to device_uvector::set_element_async

harrism:
After rapidsai/rmm#725 is merged, this PR updates cuspatial to eliminate passing literal values to `device_uvector::set_element_async`.

### DIFF
--- a/cpp/src/join/detail/traversal.cuh
+++ b/cpp/src/join/detail/traversal.cuh
@@ -72,7 +72,8 @@ descend_quadtree(LengthsIter counts,
   thrust::inclusive_scan(
     rmm::exec_policy(stream), parent_counts, parent_counts + num_quads, parent_offsets.begin() + 1);
 
-  parent_offsets.set_element_async(0, 0, stream);
+  uint32_t init{0};
+  parent_offsets.set_element_async(0, init, stream);
 
   auto num_children = parent_offsets.back_element(stream);  // synchronizes stream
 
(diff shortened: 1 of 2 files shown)


cpp/src/join/detail/traversal.cuh
```diff
@@ -72,7 +72,8 @@ descend_quadtree(LengthsIter counts,
   thrust::inclusive_scan(
     rmm::exec_policy(stream), parent_counts, parent_counts + num_quads, parent_offsets.begin() + 1);
 
-  parent_offsets.set_element_async(0, 0, stream);
+  uint32_t init{0};
+  parent_offsets.set_element_async(0, init, stream);
```
thomcom, line 76: Why eliminate one `0` literal and not the other?

harrism: Great question. Only the value parameter (second parameter) is pass-by-reference, so that's the only one we are disallowing to be a literal. The first parameter, the index, is pass-by-value so it's perfectly safe to use a literal since it will be copied rather than referenced, and so we don't have to worry about the lifetime of the temporary inside `set_element_async`.

trxcllnt (approved):

cwharris (approved):

cwharris:
@gpucibot merge

cpp/src/join/detail/traversal.cuh
thomcom: That answers my question, yes. But why have one argument be pass by reference and one pass by value? They're both just `float` or whatever. The `0` is only going to be used by the callee, and the `init` is going to be passed around and modified by others but has to be initialized here? Thanks!

PR #378, opened 2021-04-09
Fix OOM in directed_hausdorff_distances

trxcllnt:
There were two issues causing the OOM failures we've seen in hausdorff:

1. `num_cartesian` overflowed when we squared the `int32_t num_points`
2. `thrust::inclusive_scan_by_key` seems to require the number of input elements is slightly less than `INT_MAX`

This PR fixes the above, updates counters to be uint32/uint64, and makes as many things as possible const and const-ref.


cwharris (approved):
nice find, great fix. thanks!

cwharris (approved):

thomcom (approved):
thanks!

trxcllnt:
@gpucibot merge

PR #382, opened 2021-04-13
Update cuSpatial to use C++17 

trxcllnt:
cuDF 0.20 headers [now use C++17 features](https://github.com/rapidsai/cudf/pull/7931), so we need to update to C++17 also.

We will also need to update CI images to use gcc-9, as C++17 support is incomplete in gcc-7 and gcc-8.


kkraus14 (approved):

cwharris (approved):

ajschmidt8 (approved):
:fire:

trxcllnt:
@gpucibot merge

PR #392, opened 2021-05-03
fix -Werror=sign-compare causing debug builds to fail

trxcllnt:
Fixes this error that curiously only shows up when `CMAKE_BUILD_TYPE=Debug`:
```
cpp/src/spatial/hausdorff.cu:81:18: error: comparison of integer expressions of different signedness: 'const long unsigned int' and 'int' [-Werror=sign-compare]
cc1plus: all warnings being treated as errors
```


harrism:
The reason, as @robertmaynard explained to me, is that when compiling with -O1 and above gcc is able to deduce at compile time that the LHS is always under int::max and therefore the comparison is pruned out. In debug mode we don't pass any -O flag so it defaults to -O0 so there is no pruning and the code that causes the warning remains...

harrism (approved):

harrism:
@gpucibot merge

PR #399, opened 2021-05-13
Import `_version` in `__init__.py` so that `versioneer` can be used.

### LINKED ISSUE #283: [FEA] Provide version number for Python bindings
I'm using the Docker container for RAPIDS 0.15 and running the cuSpatial examples.
I noticed there doesn't seem to be a way to determine the version. I know this is related to the globally installed version of RAPIDS at least in this case but it might be clearer if there was one for cuSpatial itself.

A `_version` module as in the `cudf` package

**Describe alternatives you've considered**
Documentation about how to get the version number from other required components, e.g. `cudf`

thomcom:
Fixes #283

### DIFF
--- a/python/cuspatial/cuspatial/__init__.py
+++ b/python/cuspatial/cuspatial/__init__.py
@@ -1,3 +1,4 @@
+from cuspatial._version import get_versions
 from .core import interpolate
 from .core.gis import (
     directed_hausdorff_distance,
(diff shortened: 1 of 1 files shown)


python/cuspatial/cuspatial/__init__.py
```diff
@@ -1,3 +1,4 @@
+from cuspatial._version import get_versions
```
kkraus14, line 1: Can this be `from ._version import get_versions` just to be consistent with the other imports?

thomcom: Yes of course. Done.

kkraus14 (approved):

kkraus14:
@gpucibot merge

PR #403, opened 2021-05-20
Move rmm::device_buffer instead of copying

harrism:
Simple PR that adds a `std::move` around passing a device_buffer into a column factory. This will be necessary once rapidsai/rmm/#775 is merged.


harrism:
It's a little late but this is a trivial fix that will help avoid breakages later.

trxcllnt (approved):

harrism:
Actually it's not late. cuSpatial burn down isn't until next week. :)

cwharris (approved):

harrism:
@gpucibot merge

PR #427, opened 2021-06-26
Fix build from changes in rapidsai/cudf#8528

cwharris:
rapidsai/cudf#8528 introduced some changes which broke the cuspatial build. ~specifically `cudf.DataFrame._from_data` was change to a `@classmethod`~ cudf no longer accidentally treats `_from_data` as an instance method, which is bad behavior cuspatial was depending on. To fix this, adjustments need to be made to `cuspatial.GeoDataFrame` and some associated utilities.

PR is WIP, but so far fixes all but one failing test.

### DIFF
--- a/python/cuspatial/cuspatial/geometry/geodataframe.py
+++ b/python/cuspatial/cuspatial/geometry/geodataframe.py
@@ -90,9 +90,30 @@ class GeoDataFrame(cudf.DataFrame):
                 result.obj.drop(col, axis=1, inplace=True)
         return result
 
+    # @classmethod
+    # def _from_data(cls, data, index = None, columns = None, ):
+    #     return cudf.DataFrame._from_data(data, index, columns)
+
+    @classmethod
+    def _from_data(
+        cls,
+        data,
+        index = None,
+        columns = None,
+    ):
+        out = cls.__new__(cls)
+        out._data = data
+        if index is None:
+            index = cudf.Index(range(data.nrows))
+        out._index = index
+        if columns is not None:
+            out.columns = columns
+        return out
+
 
 class _GeoSeriesUtility:
-    def _from_data(self, new_data, name=None, index=False):
+    @classmethod
+    def _from_data(cls, new_data, name=None, index=False):
         new_column = new_data.columns[0]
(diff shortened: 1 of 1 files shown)


python/cuspatial/cuspatial/geometry/geodataframe.py
```diff
@@ -90,9 +90,30 @@ def groupby(self, *args, **kwargs):
...
+        columns = None,
+    ):
+        out = cls.__new__(cls)
+        out._data = data
+        if index is None:
+            index = cudf.Index(range(data.nrows))
+        out._index = index
+        if columns is not None:
+            out.columns = columns
+        return out
```
shwina, line 111: This seems identical to `DataFrame._from_data`, so maybe we just not override it at all?

python/cuspatial/cuspatial/geometry/geodataframe.py
```diff
@@ -90,9 +90,30 @@ def groupby(self, *args, **kwargs):
...
+
 
 class _GeoSeriesUtility:
-    def _from_data(self, new_data, name=None, index=False):
+    @classmethod
+    def _from_data(cls, new_data, name=None, index=False):
```
shwina, line 116: 👍

python/cuspatial/cuspatial/geometry/geodataframe.py
vyasr: Looking at the constructor of this class, it seems like this method probably needs some logic like 
```
if not isinstance(next(iter(data.values())), GeoColumn):
    # Convert cudf Columns to GeoColumns in `data`
return super()._from_data(data, index, columns)
```
unless this method is never called with a dict of `GeoColumn`s.

vyasr (commented):
I was very confused by the original description of this PR since `_from_data` was always a `classmethod` :) I see the edit, but perhaps the change you're referring to happened on a different PR of mine? AFAICT the one you're linking to didn't modify the relevant APIs at all.

vyasr:
I'm aiming to post a fix for this issue tomorrow.

PR #428, opened 2021-06-30
 Fix `assert_eq` related imports

galipremsagar:
Recently there has been a change of location of where the testing utils of cudf reside. This PR is adapting to those changes.

reference: rapidsai/cudf#7495

PR #430, opened 2021-07-02
Update CMake, fix testing use of assert_eq, and correct metadata copying

vyasr:
This PR contains three distinct changes required to get cuspatial builds working and tests passing again:
1. RMM switched to rapids-cmake (rapidsai/rmm#800), which requires CMake 3.20.1, so this PR includes the required updates for that.
2. The Arrow upgrade in cudf also moved the location of testing utilities (rapidsai/cudf#7495). Long term cuspatial needs to move away from use of the testing utilities, which are not part of cudf's public API, but we are currently blocked by rapidsai/cudf#8646, so this PR just imports the internal `assert_eq` method as a stopgap to get tests passing.
3. The changes in rapidsai/cudf#8373 altered the way that metadata was propagated to libcudf outputs from previously existing cuDF Python objects. The new code paths require cuspatial to override metadata copying at the GeoDataFrame rather than the GeoColumn level in order to ensure that information about column types is lost in the libcudf round trip and the metadata copying functions are now called on the output DataFrame rather than the input one.

This PR supersedes #427, #428, and #429, all of which can now be closed.

### DIFF
--- a/conda/environments/cuspatial_dev_cuda11.0.yml
+++ b/conda/environments/cuspatial_dev_cuda11.0.yml
@@ -11,7 +11,7 @@ dependencies:
   - cudatoolkit=11.0
   - gdal>=3.0.2
   - geopandas>=0.9.0,<0.10.0a0
-  - cmake>=3.18
+  - cmake>=3.20.1
   - cython>=0.29,<0.30
   - gtest
   - gmock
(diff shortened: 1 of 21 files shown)


ajschmidt8 (approved):

cwharris (approved):
Thanks! :tada:

cwharris:
@gpucibot merge

PR #436, opened 2021-07-26
Update version and std compilation in Cython

vyasr:
Updates cuspatial to 21.10 and update Cython compilation to use C++17 instead of C++14.


vyasr:
Closing since this was replaced by #438

PR #468, opened 2021-11-22
Update DEFAULT_CUDA_VER in ci/cpu/prebuild.sh

Ethyling:
Now that we only do `11.5` builds for RAPIDS, the `DEFAULT_CUDA_VER` variable in `ci/cpu/prebuild.sh` should be set to `11.5` so that the rest of the logic in the file works correctly.


ajschmidt8 (approved):

Ethyling:
@gpucibot merge

PR #490, opened 2022-02-15
Fix documentation of return type of quadtree_point_in_polygon

### LINKED ISSUE #489: [DOC] The column order returned from `quadtree_point_in_polygon` does not match what is described in the API doc.
## Report incorrect documentation
[The doc](https://github.com/rapidsai/cuspatial/blob/90be36d1348adef236ffd14abf1829820a46040b/cpp/include/cuspatial/spatial_join.hpp#L94) tells `quadtree_point_in_polygon` will return the (point_offset, polygon_offset) pair, but [the implementation](https://github.com/rapidsai/cuspatial/blob/90be36d1348adef236ffd14abf1829820a46040b/cpp/src/join/quadtree_point_in_polygon.cu#L169) actually returns (polygon_offset, point_offset) pair.

harrism:
Fixes #489 by swapping the order of the documented return values.


zhangjianting (approved):

harrism:
@gpucibot merge

Issue #492, opened 2022-02-23
inconsistent result with cuspatial.point_in_polygon


hakantekbas:
Hi;

I encountered strange result when I run notebook below.

https://github.com/rapidsai/cuspatial/blob/branch-22.04/notebooks/nyc_taxi_years_correlation.ipynb 

To show inconsistency, I did same calculation in 2 different ways. First one is pickups, second one is pickups2. Last column in pickups is calculated again in first column of pickups2. Strange part is to get different results in two calculations. Sum of true values in last column of pickups should be equal to first column of pickups2. In each calculation, sum of true values in last column is nearly the number rows. It is obvious that there is something wrong here. Can you help me to figure out?

```
start = 0
end = 4
pickups = cuspatial.point_in_polygon(taxi2016['pickup_longitude'] , taxi2016['pickup_latitude'], taxi_zones[0][start:end], taxi_zones[1], taxi_zones[2]['x'], taxi_zones[2]['y'])
pickups2 = cuspatial.point_in_polygon(taxi2016['pickup_longitude'] , taxi2016['pickup_latitude'], taxi_zones[0][3:6], taxi_zones[1], taxi_zones[2]['x'], taxi_zones[2]['y']) 
```
  
pickups 
```
            0      1      2     3
0         False  False  False  True
1         False  False  False  True
2         False  False  False  True
3         False  False  False  True
4         False  False  False  True
...         ...    ...    ...   ...
10906853  False  False  False  True
10906854  False  False  False  True
10906855  False  False  False  True
10906856  False  False  False  True
10906857  False  False  False  True
```
count of true values in pickups
```
0         605
1           1
2          32
3    10727122
dtype: uint64
```

pickups2
```
              3      4     5
0         False  False  True
1         False  False  True
2         False  False  True
3         False  False  True
4         False  False  True
...         ...    ...   ...
10906853  False  False  True
10906854  False  False  True
10906855  False  False  True
10906856  False  False  True
10906857  False  False  True
```
count of true values in pickups2

```
3       34894
4           3
5    10692225
dtype: uint64
```

hakantekbas:
I want to give one more detail. This method only supports to accept 31 polygons at one time. If I create seperate shape file for each 31 polygons, it is running 10 times faster and results are correct.

```
for i in range(len(pip_iterations)-1):
    timeStart = time.time()
    start = pip_iterations[i]
    end = pip_iterations[i+1]

[...]

PR #497, opened 2022-03-16
Improve point_in_polygon documentation regarding poly_ring_offsets

harrism:
Partially addresses #492.

Adds notes to the Python and C++ documentation explaining that behavior is undefined if there are rings in `poly_ring_offsets` that are not referenced by `poly_offsets`

### DIFF
--- a/python/cuspatial/cuspatial/core/gis.py
+++ b/python/cuspatial/cuspatial/core/gis.py
@@ -217,9 +217,9 @@ def point_in_polygon(
 
     note
     poly_ring_offsets must contain only the rings that make up the polygons
-    indexed by poly_offsets. If there are rings in poly_ring_offsets that are not
-    part of the polygons in poly_offsets, results are likely to be incorrect and
-    behavior is undefined.
+    indexed by poly_offsets. If there are rings in poly_ring_offsets that
+    are not part of the polygons in poly_offsets, results are likely to be
+    incorrect and behavior is undefined.
 
     Returns
     -------
--- a/python/cuspatial/cuspatial/core/gis.py
+++ b/python/cuspatial/cuspatial/core/gis.py
@@ -215,6 +215,11 @@ def point_in_polygon(
     input Series x and y will not be index aligned, but computed as
     sequential arrays.
 
+    note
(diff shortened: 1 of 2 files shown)


harrism:
Should probably also fix the taxi notebook but I need to get it running locally before I can test any fixes.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -215,6 +215,11 @@ def point_in_polygon(
     input Series x and y will not be index aligned, but computed as
     sequential arrays.
 
+    note
+    poly_ring_offsets must contain only the rings that make up the polygons indexed by 
+    poly_offsets. If there are rings in poly_ring_offsets that are not part of the polygons in
+    poly_offsets, results are likely to be incorrect and behavior is undefined.
```
harrism, line 221:
```suggestion
    poly_ring_offsets must contain only the rings that make up the polygons
    indexed by poly_offsets. If there are rings in poly_ring_offsets that are not
    part of the polygons in poly_offsets, results are likely to be incorrect and
    behavior is undefined.
```

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -215,6 +215,12 @@ def point_in_polygon(
     input Series x and y will not be index aligned, but computed as
     sequential arrays.
 
+    note
+    poly_ring_offsets must contain only the rings that make up the polygons
+    indexed by poly_offsets. If there are rings in poly_ring_offsets that are not
+    part of the polygons in poly_offsets, results are likely to be incorrect and
+    behavior is undefined.
```
harrism, line 222:
```suggestion
    indexed by poly_offsets. If there are rings in poly_ring_offsets that
    are not part of the polygons in poly_offsets, results are likely to be
    incorrect and behavior is undefined.
```

cwharris (approved):

thomcom (approved):

isVoid:
Might be out of scope, would it be helpful to add an example of input to help demonstrate the use of the API?

harrism:
@isVoid Yes. However I'm going to merge this and add your note to the still-open issue.  This way at least we get some improvement in the current release.

PR #504, opened 2022-03-25
Support `pairwise_polygon_distance`

### LINKED ISSUE #503: [FEA] Add Clang-format to Pre-commit Hooks
As a follow up to #177 , it would be great to setup clang-format to apply automatic formatting prior to every change in cuspatial c++ files.

isVoid:
This PR contributes to #231, and should supercede #251 .


harrism:
Suggest separating clang-format to a standalone PR so that it gets its own line item in release notes and commit history.

zhangjianting:
Just a comment from application perspective: while Hausdorff distance matrix can be used as the input of clustering algorithms to group trajectories, it would be a little difficult to figure out a practical geospatial application that requires pair-wise distance matrix for a set of polygons (clustering/grouping polygons based purely on distance does not make a good geospatial application). 
It could be more useful to compute distances for a vector of polygon pairs, i.e., vectorized computation of polygon-polygon distances, even though it is still much less frequently used than point-to-point, point-to-polyline and polyline-to-polyline distance.

zhangjianting:
For test, I would suggest verify polygon-polygon distance result by comparing against libgeos result, as that would be the what geospaital community would expect.

isVoid:
@zhangjianting out of scope of this PR. For polyline-polyline distance, do you consider the API should compute the matrix from a set of polylines or vectorize from a pair of polylines?

robertmaynard (approved):
CMake changes LGTM

isVoid:
closing in favor of #1074

PR #510, opened 2022-03-31
C++ pairwise linestring distance

isVoid:
This PR contributes to #231 , computes distances between pairs of linestrings.


harrism:
What is the difference between a linestring and a polyline?

isVoid:
> What is the difference between a linestring and a polyline?

I believe they are the same. Both refer to a collection of connected line segments. The term `Linestring` is used in OGC SFS and is adopted here - we can change that if needed.

harrism (changes requested):
Partial review. More tomorrow.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,356 @@
...
+
+/**
+ * @brief Computes shortest distance between @p C and segment @p A @p B
+ */
+template <typename T>
+double __device__ point_to_segment_distance(coord_2d<T> const& C,
+                                            coord_2d<T> const& A,
+                                            coord_2d<T> const& B)
+{
+  double L_squared = (A.x - B.x) * (A.x - B.x) + (A.y - B.y) * (A.y - B.y);
```
harrism, line 67: Let's add vec_2d operators (+, -, dot, scalar *) and make this code more geometrically intuitive.

Also, change this function to `point_to_segment_distance_squared(...)`...

```
template <typename T>
double __device__ point_to_segment_distance_squared(coord_2d<T> const& c,
                                            coord_2d<T> const& a,
                                            coord_2d<T> const& b)
{
  auto ab = b - a;
  auto ac = c - a;
  auto l_squared = dot(ab, ab); // could also use operator*(vec, vec) for dot
  if (L_squared == 0) { return dot(ac, ac); }
  auto r = dot(ac, ab) / L_squared;
  if (r <= 0 or r >= 1) {
    auto bc = c - b;
    return std::min(dot(ac, ac), dot(bc, bc));
  }
  auto p = a + r * ab;
  auto pc = c - p;
  return dot(pc, pc);
}
```

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,356 @@
...
+ * @brief Computes shortest distance between two segments that doesn't intersect.
+ */
+template <typename T>
+double __device__ segment_distance_no_intersect_or_collinear(coord_2d<T> const& A,
+                                                             coord_2d<T> const& B,
+                                                             coord_2d<T> const& C,
+                                                             coord_2d<T> const& D)
+{
+  return std::min(std::min(point_to_segment_distance(A, C, D), point_to_segment_distance(B, C, D)),
+                  std::min(point_to_segment_distance(C, A, B), point_to_segment_distance(D, A, B)));
```
harrism, line 88: This is overkill. The minimum distance corresponds to the minimum squared distance, which is cheaper to compute. So find the minimum squared distance, and then take the square root. This will save at least 3 sqrt/hypot calls.
```suggestion
  auto dist_sqr = std::min(std::min(point_to_segment_distance_squared(A, C, D), point_to_segment_distance_squared(B, C, D)),
                  std::min(point_to_segment_distance_squared(C, A, B), point_to_segment_distance_squared(D, A, B)));
  return sqrt(dist_sqr);
```

cpp/tests/spatial/linestring_distance_test.cpp
```diff
@@ -0,0 +1,399 @@
...
+  auto got = pairwise_linestring_distance(column_view(linestring1_offsets),
+                                          linestring1_points_x,
+                                          linestring1_points_y,
+                                          column_view(linestring2_offsets),
+                                          linestring2_points_x,
+                                          linestring2_points_y);
+  expect_columns_equivalent(expected, *got, verbosity);
+}
+
+TYPED_TEST(PairwiseLinestringDistanceTest, OnePairGeolife)
```
harrism, line 289: What is Geolife?

cpp/tests/spatial/linestring_distance_test.cpp
```diff
@@ -0,0 +1,399 @@
...
+template <typename T>
+struct PairwiseLinestringDistanceTest : public BaseFixture {
+};
+
+// float and double are logically the same but would require separate tests due to precision.
+using TestTypes = FloatingPointTypes;
+TYPED_TEST_CASE(PairwiseLinestringDistanceTest, TestTypes);
+
+constexpr cudf::test::debug_output_level verbosity{cudf::test::debug_output_level::ALL_ERRORS};
+
```
harrism, line 43: One degenerate test I would like to see is one that exercises this:
```
auto r_denom = cross(ab, cd);
  if (r_denom == 0) {  // should we be using an approximate equals or < epsilon to avoid division by a very small number?
    // Segments parallel or collinear
    return segment_distance_no_intersect_or_collinear(A, B, C, D);
  }
  auto r_numer = cross(ac, cd); // note I changed CA in your code to AC here for consistency
  auto r = r_numer / r_denom;
```

Specifically, the case where `r_denom` is very small but non-zero (possibly denormalized).  So the test could be using two segments that are very nearly but not quite coincident and non-parallel.  If the denominator ends up very small, then you could end up with a very large r. This should be OK, but it's good to test to ensure it is robust.

harrism (changes requested):
I think we can make this more efficient, more geometrically intuitive / self-documenting, and more maintainable by adding and using some vector math functions.  Also by using the property that square distances can often be used for comparisons until you need a final distance, and only then do you need to take the sqrt.

vyasr:
> I think we can make this more efficient, more geometrically intuitive / self-documenting, and more maintainable by adding and using some vector math functions. Also by using the property that square distances can often be used for comparisons until you need a final distance, and only then do you need to take the sqrt.

I don't know enough about cuspatial to know whether this is a good idea, but have we considered making use of raft? I know it has some primitives for e.g. spatial distance calculations that cuspatial might be abel to reuse. I don't know if that's a larger dependency that cuspatial wants to have, though.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,356 @@
...
+    true,
+    [offsets     = linestring_offsets.begin(),
+     num_offsets = static_cast<cudf::size_type>(linestring_offsets.size()),
+     num_points  = static_cast<cudf::size_type>(
+       linestring_points_x.size())] __device__(bool prev, cudf::size_type i) {
+      cudf::size_type begin = offsets[i];
+      cudf::size_type end   = i == num_offsets ? num_points : offsets[i + 1];
+      return prev && (end - begin);
+    });
+}
```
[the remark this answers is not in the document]
isVoid: I agree - now that I think of it, this is just a one-pass over the input index array and the lambda is so simple. It's not apparent that verifying on GPU will be faster than doing it on CPU.

cpp/include/cuspatial/distances/linestring_distance.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * @return A column of shortest distances between the pair of linestrings
+ *
+ * @throw cuspatial::logic_error if `linestring1_offsets.size() != linestring2_offsets.size()`
+ * @throw cuspatial::logic_error if size mismatch between the x, y components of the linestring
+ * points.
+ * @throw cuspatial::logic_error if any of the point arrays have mismatch types.
+ * @throw cuspatial::logic_error if any linestring has less than 2 end points.
+ *
+ */
+std::unique_ptr<cudf::column> pairwise_linestring_distance(
```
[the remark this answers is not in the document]
isVoid: The limit is that the largest number of points both set can have in total is less than `std::numeric_limit<size_type>::max()`. This is more restrictive than having separate set of point vectors.
cc. @zhangjianting with usage question.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,356 @@
...
+  double r_denom = (B.x - A.x) * (D.y - C.y) - (B.y - A.y) * (D.x - C.x);
+  double r_numer = (A.y - C.y) * (D.x - C.x) - (A.x - C.x) * (D.y - C.y);
+  if (r_denom == 0) {
+    // Segments parallel or collinear
+    return segment_distance_no_intersect_or_collinear(A, B, C, D);
+  }
+  double r = r_numer / r_denom;
+  double s = ((A.y - C.y) * (B.x - A.x) - (A.x - C.x) * (B.y - A.y)) /
+             ((B.x - A.x) * (D.y - C.y) - (B.y - A.y) * (D.x - C.x));
+  if (r >= 0 and r <= 1 and s >= 0 and s <= 1) { return 0.0; }
```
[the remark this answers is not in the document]
isVoid: > if (r_denom == 0) { // should we be using an approximate equals or < epsilon to avoid division by a very small number?

How do you usually pick epsilon when the type is templated?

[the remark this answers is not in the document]
isVoid: > auto s = cross(ab, ac) / cross(ab, cd);

I noticed in the other branch that `cross(ac, cd)` is recomputed here.

cpp/src/spatial/linestring_distance.cu
isVoid: > auto l_squared = dot(ab, ab); // could also use operator*(vec, vec) for dot

I think dot is better - you need an extra line to add the components with `vec*`.

harrism:
I'm trying to eliminate the cuDF dependency and make libcuspatial lighter weight. I'd rather not add a heavy library like raft. Besides, if there's a spatial distance computation, logically it makes sense to put that in cuSpatial, then other rapids libraries can use it (especially once it is header-only).

cpp/include/cuspatial/types.hpp
```diff
@@ -36,7 +36,7 @@ struct location_3d {
  * @tparam T the base type for the coordinates
  */
 template <typename T>
-struct coord_2d {
+struct vec_2d {
```
harrism, line 39: I didn't mean you should rename this. This renaming is part of #477 which is about to merge.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -38,6 +38,48 @@
 namespace cuspatial {
 namespace {
 
+template <typename T>
+vec_2d<T> __device__ operator+(vec_2d<T> const& a, vec_2d<T> const& b)
```
harrism, line 42: Let's move vec_2d (and its derived types) to a vec_2d.hpp and put these operators in that header. Also make these `__host__ __device__` (and use a macro for that so it's portable).

harrism: Also need to document these functions.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -38,6 +38,48 @@
...
+}
+
+template <typename T>
+vec_2d<T> __device__ operator-(vec_2d<T> const& a, vec_2d<T> const& b)
+{
+  return vec_2d<T>{a.x - b.x, a.y - b.y};
+}
+
+template <typename T>
+vec_2d<T> __device__ operator*(vec_2d<T> const& a, vec_2d<T> const& b)
```
harrism, line 54: Is componentwise multiplication an operator we should have?

isVoid: It's not used in this PR. Can be removed and reintroduced later if needed.

isVoid: Not in the context of this PR. Removed.

cpp/tests/spatial/linestring_distance_test.cpp
isVoid: Test added!

cpp/tests/spatial/linestring_distance_test.cpp
isVoid: Test renamed.

vyasr:
@harrism I'm on board with making cuspatial more lightweight, I'd just like to avoid duplicating work if possible. Perhaps we should sync with the raft team now to figure out if there is overlap between raft and cuspatial, and if so, how best to coordinate? I assume the shared logic (if it exists) would be at a lower level since the cuspatial APIs are based on geospatial coordinates while raft is probably using raw numbers representing dimensionless quantities.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -101,19 +101,22 @@ double __device__ segment_distance_no_intersect_or_collinear(vec_2d<T> const& A,
...
     // Segments parallel or collinear
-    return segment_distance_no_intersect_or_collinear(A, B, C, D);
+    return segment_distance_no_intersect_or_collinear(a, b, c, d);
   }
-  double r = r_numer / r_denom;
-  double s = ((A.y - C.y) * (B.x - A.x) - (A.x - C.x) * (B.y - A.y)) /
-             ((B.x - A.x) * (D.y - C.y) - (B.y - A.y) * (D.x - C.x));
+  double r_numer = det(ac, cd);
+  double r       = r_numer / denom;
+  double s       = det(ac, ab) / denom;
```
harrism: Use auto, not double...

isVoid: Somehow when I look at the branch this line is already updated with `auto`.

harrism: I probably clicked on a specific commit. Sorry.

cpp/include/cuspatial/distances/linestring_distance.hpp
```diff
@@ -0,0 +1,109 @@
...
+ * (0, 2) -> (0, 3) -> (1, 3)
+ *
+ * Both linestrings contains degenerate segments, their distance is 1.
+ *
+ * Third pair:
+ *
+ * (0, 0) -> (2, 2) -> (-2, 0)
+ * (2, 0) -> (0, 2)
+ *
+ * These linestrings intersects, their distance is 0
```
harrism, line 62:
```suggestion
 * These linestrings intersect, so their distance is 0.
```

cpp/include/cuspatial/distances/linestring_distance.hpp
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Forth pair:
+ *
+ * (2, 2) -> (-2, -2)
+ * (1, 1) -> (5, 5) -> (10, 0)
+ *
+ * These linestrings contain collinear and overlapping sections,
```
harrism, line 69:
```suggestion
 * These linestrings contain colinear and overlapping sections, so
```

cpp/include/cuspatial/utility/vec_2d.cuh
```diff
@@ -0,0 +1,60 @@
+#pragma once
+#include <cuspatial/types.hpp>
+
+namespace cuspatial {
+
```
harrism, line 5: I think the definition of the 2D vector types should be moved into this header as well. (types.hpp might go away...)

Also since this header can be used from host source files, it should use the .hpp extension.

cpp/tests/spatial/linestring_distance_test.cpp
```diff
@@ -0,0 +1,471 @@
...
+                                          linestring2_points_x,
+                                          linestring2_points_y);
+  expect_columns_equivalent(expected, *got, verbosity);
+}
+
+TEST_F(PairwiseLinestringDistanceTestUntyped, OnePairDeterminantDoublePrecisionDenormalized)
+{
+  // Vector ab: (1e-155, 2e-155)
+  // Vector cd: (2e-155, 1e-155)
+  // deternminant of matrix [a, b] = -3e-310, a denormalized number
```
harrism, line 275:
```suggestion
  // determinant of matrix [a, b] = -3e-310, a denormalized number
```

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * @note The last endpoint of the linestring is not included in the offset array, thus
+ * @p num_points is returned.
+ */
+template <typename OffsetIterator>
+inline cudf::size_type __device__
+endpoint_index_of_linestring(cudf::size_type const& linestring_idx,
+                             OffsetIterator const& linestring_offsets_begin,
+                             cudf::size_type const& num_linestrings,
+                             cudf::size_type const& num_points)
```
harrism, line 52: I think we should avoid explicitly using `cudf::size_type` here. Perhaps template this function on the index type.

harrism (changes requested):
Some doc fixes and one other thing to consider.

cpp/include/cuspatial/distances/linestring_distance.hpp
isVoid: I've always remembered `colinear` with 2 `l`s. Updating the method name in `linestring_distance.cu` too.

cpp/include/cuspatial/distances/linestring_distance.hpp
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * The shortest distance between the two linstrings are the distance
+ * from point (1, 1) to segment (0, 1) -> (1, 0), which is sqrt(2)/2.
+ *
+ * Second pair:
+ *
+ * (0, 0) -> (0, 1)
+ * (0, 2) -> (0, 3) -> (1, 3)
+ *
+ * Both linestrings contains degenerate segments, their distance is 1.
```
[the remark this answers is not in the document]
isVoid: I viewed degenerate lines as lines infinite slopes. So `(0,0)->(0,1)` and (0,2)->(0,3)`?

cpp/include/cuspatial/distances/linestring_distance.hpp
isVoid: Actually both spelling works: https://www.merriam-webster.com/dictionary/colinear

cpp/include/cuspatial/distances/linestring_distance.hpp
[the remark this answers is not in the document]
harrism: Those are vertical line segments in the 2D plane. Vertical lines are not degenerate AFAIK. :)  Degeneracy is not an exact term. Any line segment is a degenerate something  else (circle, triangle, etc.).  A point can be considered a degenerate line segment. But even with two points (degenerate line  segments) we can  find the distance. In this example I don't think we should call any of these segments degenerate. And the fact that they have +inf slope doesn't impact the distance calculation does it?

cpp/include/cuspatial/distances/linestring_distance.hpp
harrism: Yeah TIL they both work. Pick one. I have always liked colinear because it's the prefix co plus the root linear. It matches coplanar. But it seems many people use collinear. English is tricky that way.

harrism (approved):

cpp/include/cuspatial/distances/linestring_distance.hpp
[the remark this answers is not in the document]
isVoid: Got it - thanks for clarifying!

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,334 @@
...
+  vec_2d<T> A{linestring1_points_xs_begin[p1_idx], linestring1_points_ys_begin[p1_idx]};
+  vec_2d<T> B{linestring1_points_xs_begin[p1_idx + 1], linestring1_points_ys_begin[p1_idx + 1]};
+
+  T min_distance = std::numeric_limits<T>::max();
+  for (cudf::size_type p2_idx = ls2_start; p2_idx < ls2_end; p2_idx++) {
+    vec_2d<T> C{linestring2_points_xs_begin[p2_idx], linestring2_points_ys_begin[p2_idx]};
+    vec_2d<T> D{linestring2_points_xs_begin[p2_idx + 1], linestring2_points_ys_begin[p2_idx + 1]};
+    min_distance = std::min(min_distance, segment_distance(A, B, C, D));
+  }
+  atomicMin(distances + linestring_idx, static_cast<T>(min_distance));
```
trxcllnt, line 214: @harrism mentioned this [here](https://github.com/rapidsai/cuspatial/pull/510#pullrequestreview-954178500), but we should be able to remove the `std::sqrt()` from `segment_distance`, comparing squared distances and taking the root of the min squared distance at the end (I recall seeing a nice performance improvement when we did this in [`point_to_poly_line_distance`](https://github.com/rapidsai/cuspatial/blob/1986a18e529888f9fe8df7c120924fe52cb1f06e/cpp/src/utility/point_to_nearest_polyline.cuh#L57-L60)):

```suggestion
  atomicMin(distances + linestring_idx, static_cast<T>(std::sqrt(min_distance)));
```

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,334 @@
...
+T __device__ segment_distance_no_intersect_or_colinear(vec_2d<T> const& a,
+                                                       vec_2d<T> const& b,
+                                                       vec_2d<T> const& c,
+                                                       vec_2d<T> const& d)
+{
+  auto dist_sqr = std::min(std::min(point_to_segment_distance_squared(a, c, d),
+                                    point_to_segment_distance_squared(b, c, d)),
+                           std::min(point_to_segment_distance_squared(c, a, b),
+                                    point_to_segment_distance_squared(d, a, b)));
+  return std::sqrt(dist_sqr);
```
trxcllnt, line 93: see https://github.com/rapidsai/cuspatial/pull/510/files#r865488840
```suggestion
  return dist_sqr;
```

trxcllnt (changes requested):
Left a small suggestion to reduce the number of `sqrt()` calls.

cpp/src/spatial/linestring_distance.cu
harrism: Good catch! I thought @isVoid had already made this change, I didn't look close enough.

isVoid: I've renamed the variable (and method) with `squared_`.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * @brief Computes shortest distance between @p C and segment @p A @p B
+ */
+template <typename T>
+T __device__ point_to_segment_distance_squared(vec_2d<T> const& c,
+                                               vec_2d<T> const& a,
+                                               vec_2d<T> const& b)
+{
+  vec_2d<T> ab   = b - a;
```
vyasr, line 68:
```suggestion
  auto ab   = b - a;
```

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,335 @@
...
+  auto ac        = c - a;
+  auto bc        = c - b;
+  auto l_squared = dot(ab, ab);
+  if (l_squared == 0) { return dot(ac, ac); }
+  auto r = dot(ac, ab) / l_squared;
+  if (r <= 0 or r >= 1) { return std::min(dot(ac, ac), dot(bc, bc)); }
+  auto p  = a + r * ab;
```
vyasr, line 75: Minor nit, but it's always nice to delay/avoid expensive ops like division
```suggestion
  auto r = dot(ac, ab);
  if (r <= 0 or r >= l_squared) { return std::min(dot(ac, ac), dot(bc, bc)); }
  auto p  = a + (r / l_squared) * ab;
```

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,335 @@
...
+{
+  auto ab    = b - a;
+  auto ac    = c - a;
+  auto cd    = d - c;
+  auto denom = det(ab, cd);
+
+  if (denom == 0) {
+    // Segments parallel or collinear
+    return segment_distance_no_intersect_or_colinear(a, b, c, d);
+  }
```
vyasr, line 116:
```suggestion
  auto ab    = b - a;
  auto cd    = d - c;
  auto denom = det(ab, cd);

  if (denom == 0) {
    // Segments parallel or collinear
    return segment_distance_no_intersect_or_colinear(a, b, c, d);
  }

  auto ac    = c - a;
```

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,335 @@
...
+  auto r_numer = det(ac, cd);
+  auto r       = r_numer / denom;
+  auto s       = det(ac, ab) / denom;
```
vyasr, line 119: Could always precompute 1/denom and then multiply, but I'm probably microoptimizing here.

cpp/src/spatial/linestring_distance.cu
```diff
@@ -0,0 +1,335 @@
...
+
+    auto distances =
+      make_numeric_column(data_type{type_to_id<T>()}, num_string_pairs, mask_state::UNALLOCATED);
+
+    thrust::fill(rmm::exec_policy(stream),
+                 distances->mutable_view().begin<T>(),
+                 distances->mutable_view().end<T>(),
+                 std::numeric_limits<T>::max());
+
+    std::size_t const threads_per_block = 64;
```
vyasr, line 249:
```suggestion
    std::size_t constexpr threads_per_block = 64;
```

vyasr (commented):
I'm getting notifications on this thread after commenting, and I've been seeing more cuspatial activity lately so I figured I'd give this a brief review to get a sense for some of the code structure. Most of my comments are probably microoptimizations, nothing blocking.

cpp/src/spatial/linestring_distance.cu
isVoid: Out of academic curiosity, are there precision related concerns when replacing division with multiply by reciprocal?

vyasr: That's a great question. Yes, there will be differences. Division is usually implemented using a root finding scheme. The numerical convergence properties of that scheme will result in different results if you do `z = y/x` vs doing `tmp = 1 / x; z = y * x`. The bigger issue will be truncation error resulting from the magnitude of `x`. The difference in numerical error between division and multiplication by reciprocal will probably be dominated by the same kind of thing as doing summations of a sequence: you will lose more precision when you perform operations between numbers of very different scales. In this example, if x is of order unity then `1/x` will be computed precisely and you won't lose much, whereas if `x` is much larger or much smaller than `1` then you'll lose some digits. Which approach is more precise will depend on whether `x` and `y` are of similar magnitude. If `x~y` but not `x~1` (which is probably a realistic expectation) you'll get more precise results from doing `z = y/x` than storing the reciprocal as the intermediate. 

All of this is to say that I don't know enough about what tradeoffs cuspatial wants to/is willing to make. My guess would be that choosing the more precise result may be the better solution unless and until someone really wants to squeeze every last drop of performance out of this code. I don't think we're at the point where we try to optimize that far in cuspatial.

harrism: Here is an amazingly detailed discussion of this for x86. https://stackoverflow.com/a/45899202/749748  This of course will be somewhat different for NVIDIA / SASS. 

Like x86, our GPUs do have an rcp instruction which can be used for approximate division. There are multiple other ways to divide other than `/`, including `__fdividef(x,y)`. There is a bit of discussion of the accuracy of division in the CUDA Programming guide: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#intrinsic-functions

All this said, in the absence of an accuracy test specific to real data and a corresponding benchmark, I think you should stick to the division you have in the code above, since it is guaranteed to be IEEE compliant and has the best chance of matching results of other libraries. 

For the future, the fact that these computed values are only used for comparison to zero and one may provide optimization opportunities. But right now let's just make it robust.

trxcllnt (approved):

cpp/src/spatial/linestring_distance.cu
vyasr: Agreed, this kind of microoptimization really shouldn't happen without a compelling benchmark as well as validation that the potential loss of precision wouldn't be a problem.

harrism: @isvoid looks like you already changed to computing reciprocal. Have you compared accuracy or performance vs. previous? It may not make much difference to accuracy as I said, since this is only used for comparing if the results are between 0 and 1.

isVoid: Short answer is I haven't done any specific test to this yet. For accuracy the current implementation passes all existing tests (including offline geolife tests) the same as previous code.

isVoid: On to performance, with the offline dataset (8800 pairs of linestrings), there is a visible 6% speedup with the multiply by reciprocal approach. The below timing are acquired by timing the python bindings (with different cpp implementation) with `timeit`. Each ran 50 times, the average timings are:
```
0.7526701703201979 for multiply by reciprocal
0.8014753964729607 for division
```

harrism: That's good!

harrism:
I think we just need @vyasr to approve.

cpp/src/spatial/linestring_distance.cu
vyasr: Well then I guess this was a worthwhile optimization to suggest after all!

vyasr (approved):

harrism:
@gpucibot merge

PR #515, opened 2022-04-07
Use rapids-cmake for builds

vyasr:
Using rapids-cmake helps reduce duplication across different RAPIDS projects and makes various configuration aspects simpler.


vyasr:
@robertmaynard I think this PR is mostly OK. The main area where I am very uncertain that I did everything right is with the exporting of dependencies for the creation of the config files, so please check through that part extra carefully. I also have TODOs scattered throughout the changes where I wasn't sure how to proceed, so let me know what you think of those.

cpp/CMakeLists.txt
```diff
@@ -84,21 +72,19 @@ message(STATUS "CUSPATIAL: RMM_LOGGING_LEVEL = '${RMM_LOGGING_LEVEL}'.")
...
-    message(VERBOSE "CUSPATIAL: Conda environment detected, CMAKE_PREFIX_PATH set to: ${CMAKE_PREFIX_PATH}")
-endif()
+rapids_cmake_support_conda_env(conda_env MODIFY_PREFIX_PATH)
 
 ###################################################################################################
 # - compiler options ------------------------------------------------------------------------------
 
+rapids_find_package(
+  CUDAToolkit REQUIRED
+  # TODO: Do we want cuspatial to export CUDA the same way that cudf does? I
```
robertmaynard, line 82: We want to express any first order depdenencies. So if cuspatial is making direct cuda runtime calls or explicitly launching kernels we want this in our export set.

cpp/CMakeLists.txt
```diff
@@ -109,9 +95,13 @@ include(cmake/Modules/ConfigureCUDA.cmake)
 # - dependencies ----------------------------------------------------------------------------------
 
 # find gdal
+# TODO: How do we add this to rapids-export? I don't think we want to
+# rapids-cpm-find this because we're not looking to download and build a local
+# copy, so should we be using one of the rapids_export_* functions to manually
+# add it to the list?
```
robertmaynard, line 101: You want to use `rapids_find_package` which behaves like `find_package` but with depdency tracking.

https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_find_package.html

cpp/CMakeLists.txt
```diff
@@ -233,37 +223,48 @@ set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME cuspatial)
...
 
-write_basic_package_version_file(${CUSPATIAL_BINARY_DIR}/cuspatial-config-version.cmake
-                                 COMPATIBILITY SameMinorVersion)
+################################################################################################
+# - build export -------------------------------------------------------------------------------
 
-export(EXPORT cuspatial-targets
-    FILE ${CUSPATIAL_BINARY_DIR}/cuspatial-targets.cmake
-    NAMESPACE   cuspatial::)
+# TODO: I'm not sure how these two exports should differ here.
```
robertmaynard, line 263: They should differ by the `BUILD` keyword only. This tells rapids-cmake to place the information in the build directory so it can be used to find cuspatial. Without that you would need to first install cuspatial to find it via CMake's `find_package`.

cpp/cmake/thirdparty/CUSPATIAL_GetCUDF.cmake
```diff
@@ -26,15 +38,15 @@ function(find_and_configure_cudf VERSION)
...
+      CPM_ARGS
+      GIT_REPOSITORY https://github.com/rapidsai/cudf.git
+      GIT_TAG branch-${MAJOR_AND_MINOR}
+      GIT_SHALLOW TRUE
+      OPTIONS "BUILD_TESTS OFF" "BUILD_BENCHMARKS OFF"
+      FIND_PACKAGE_ARGUMENTS "COMPONENTS testing"
+    )
 
     # Make sure consumers of cuspatial can see cudf::cudf
     fix_cmake_global_defaults(cudf::cudf)
```
robertmaynard, line 52: We can fully remove `fix_cmake_global_defaults` and instead specify `cudf::cudf` as a `GLOBAL_TARGET` in the `rapids_cpm_find` call.

https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_cpm_find.html

robertmaynard (changes requested):

cpp/CMakeLists.txt
vyasr: Ah right OK so `rapids_cpm_find` will perform dependency tracking _and_ use CPM to download a package if it's not found locally, whereas `rapids_find_package` will perform dependency tracking but _will not_ download and build missing packages? In this case we definitely don't want to build GDAL locally.

robertmaynard: correct!

ajschmidt8 (approved):

cpp/cmake/thirdparty/CUSPATIAL_GetCUDF.cmake
```diff
@@ -14,6 +14,18 @@
 # limitations under the License.
 #=============================================================================
 
+# If a target is installed, found by the `find_package` step of CPMFindPackage,
+# and marked as IMPORTED, make it globally accessible to consumers of our libs.
+function(fix_cmake_global_defaults target)
```
robertmaynard, line 19: This function is not used and can be deleted

robertmaynard (changes requested):
This is looking really nice. Only one issue left

vyasr:
@gpucibot merge

harrism:
Thanks for doing this @vyasr !

PR #521, opened 2022-05-06
Python Bindings for Pairwise Linestring Distance

isVoid:
This PR introduces python bindings for pairwise linestring distance. Depends on #510 and has lots of commits from that PR. Should reduce once #510 merges.


isVoid:
Note - the doc entry of this method will be added after #523 is merged.

python/cuspatial/cuspatial/_lib/linestring_distance.pyx
```diff
@@ -0,0 +1,39 @@
...
+    cdef column_view linestring1_offsets_view = linestring1_offsets.view()
+    cdef column_view linestring1_points_x_view = linestring1_points_x.view()
+    cdef column_view linestring1_points_y_view = linestring1_points_y.view()
+    cdef column_view linestring2_offsets_view = linestring2_offsets.view()
+    cdef column_view linestring2_points_x_view = linestring2_points_x.view()
+    cdef column_view linestring2_points_y_view = linestring2_points_y.view()
+
+    cdef unique_ptr[column] c_result
+    with nogil:
+        c_result = move(cpp_pairwise_linestring_distance(
```
harrism, line 30: Why do you `move` here and in the return line below?

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -335,3 +338,40 @@ def polyline_bounding_boxes(poly_offsets, xs, ys, expansion_radius):
     return DataFrame._from_data(
         *cpp_polyline_bounding_boxes(poly_offsets, xs, ys, expansion_radius)
     )
+
+
+def pairwise_linestring_distance(xs1, ys1, offsets1, xs2, ys2, offsets2):
+    """Compute the distances of pairs of linestrings from two arrays.
```
harrism, line 344: Let's use the same description as in the C++ API.

```suggestion
    """Compute shortest distance between pairs of linestrings (a.k.a. polylines).
```

Please also copy the examples from the C++ API here.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -335,3 +338,40 @@ def polyline_bounding_boxes(poly_offsets, xs, ys, expansion_radius):
     return DataFrame._from_data(
         *cpp_polyline_bounding_boxes(poly_offsets, xs, ys, expansion_radius)
     )
+
+
+def pairwise_linestring_distance(xs1, ys1, offsets1, xs2, ys2, offsets2):
```
harrism, line 343: In the C++ API you have the offsets before the coordinates, which is the opposite of this Python API. It would be nice to keep them consistent.

python/cuspatial/cuspatial/core/gis.py
```diff
@@ -335,3 +338,40 @@ def polyline_bounding_boxes(poly_offsets, xs, ys, expansion_radius):
...
+    """Compute the distances of pairs of linestrings from two arrays.
+
+    Parameters
+    ----------
+    xs1
```
harrism, line 348: Use the parameter descriptions from the C++ API since they should be the same.

harrism (changes requested):
Just some suggestions on expanding the documentation and making it match the C++ API, since the behavior should be the same.

python/cuspatial/cuspatial/_lib/linestring_distance.pyx
isVoid: I believe this relates to how cython translates it into cpp. Without `move` this line is tranlated into:
```cpp
        /* "cuspatial/_lib/linestring_distance.pyx":30
 *     cdef unique_ptr[column] c_result
 *     with nogil:
 *         c_result = cpp_pairwise_linestring_distance(             # <<<<<<<<<<<<<<
 *             linestring1_offsets_view,
 *             linestring1_points_x_view,
 */
        try {
          __pyx_t_2 = cuspatial::pairwise_linestring_distance(__pyx_v_linestring1_offsets_view, __pyx_v_linestring1_points_x_view, __pyx_v_linestring1_points_y_view, __pyx_v_linestring2_offsets_view, __pyx_v_linestring2_points_x_view, __pyx_v_linestring2_points_y_view);
        } catch(...) {
          #ifdef WITH_THREAD
          PyGILState_STATE __pyx_gilstate_save = __Pyx_PyGILState_Ensure();
          #endif
          __Pyx_CppExn2PyErr();
          #ifdef WITH_THREAD
          __Pyx_PyGILState_Release(__pyx_gilstate_save);
          #endif
          __PYX_ERR(0, 30, __pyx_L4_error)
        }
        __pyx_v_c_result = __pyx_t_2;
      }
```
The last line by default copies `__pyx_t_2` into a new variable and will error for `unique_ptr`.

thomcom (approved):
lg2m

vyasr:
@isVoid I'm pretty sure the test failures here came from CI merging in my scikit-build changes from branch-22.06 down into your branch and then attempting to compile the resulting files. Since the new builds require adding the Python files to the CMakeLists.txt, your new module wasn't getting compiled. I merged in branch-22.06 and added the new file, which hopefully should fix the issue.

harrism:
> RuntimeError: cuSpatial failure at: /workspace/.conda-bld/work/cpp/src/spatial/linestring_distance.cu:292: Mismatch number of linestrings in the linestring pair array.

harrism (approved):

harrism:
@gpucibot merge

PR #523, opened 2022-05-11
Update Documentation with Pydata Sphinx Theme, and more

### LINKED ISSUE #459: [DOC] Update cuSpatial documentation to use PyData Sphinx theme
As summarized by @isVoid in https://github.com/rapidsai/cudf/issues/7098 , Python API documentation for RAPIDS projects has historically used the Sphinx Read The Docs theme and served a single large HTML page. This has caused the docs to load slowly as API surfaces have grown and caused users to experience potentially significant latency/delays when searching for functions.

In https://github.com/rapidsai/cudf/pull/8746 , @galipremsagar updated the cuDF documentation to the more responsive PyData Sphinx theme. In https://github.com/rapidsai/cugraph/pull/1793, @BradReesWork made a similar change.

Updating the cuSpatial documentation to the PyData Sphinx theme should improve the user experience for our documentation.

isVoid:
This PR closes #459 . It updates the sphinx theme with `Pydata-sphinx-theme`.

Besides this PR also:

- [x] Fixed all existing warnings emitted during rendering the doc
- [x] Added a new section: `internals.rst` to focus on discussing GeoArrow format. Shuffled around the docstrings for `GeoArrowBuffer` and make background description and API usage relatively independent, but all presented in the same page for shared context.
- [x] Minor improvement to `from_geopandas` docstring.


isVoid:
Front page:
image

docs/source/api_docs/internals.rst
```diff
@@ -0,0 +1,100 @@
...
+The `Point` geometry is the simplest: N points are stored in a length 2*N
+buffer with interleaved x,y coordinates. An optional z buffer of length N
+can be used.
+
+The `Multipoint` geometry is the second simplest - identical to points,
+with the addition of a `multipoints_offsets` buffer. The offsets buffer
+stores N+1 indexes. The first multipoint is specified by 0, which is always
+stored in offsets[0], and offsets[1], which is the length in points of
+the first multipoint geometry. Subsequent multipoints are the prefix-sum of
+the lengths of previous multipoints.
```
harrism, line 44: Invoking prefix sum here implies that the lengths are stored, which I don't believe they are.

```suggestion
A `Multipoint` is a group of points, and is the second simplest GeoArrow 
geometry type. It is identical to points, with the addition of a 
`multipoints_offsets` buffer. The offsets buffer stores N+1 indexes. The
 first multipoint offset is specified by 0, which is always stored in 
 `offsets[0]`. The second offset is stored in `offsets[1]`, and so on.
 The length of each multipoint `i` is the difference between `offsets[i+1]`
 and `offsets[i]`. 
```

docs/source/api_docs/internals.rst
```diff
@@ -0,0 +1,100 @@
...
+Which encodes a GeoPandas Series::
+
+    series = geopandas.Series([
+        LineString((0, 0), (0, 1), (0, 2)),
+        MultiLineString([(1, 0), (1, 1), (1, 2)],
+                        [(2, 0), (2, 1), (2, 2)],
+        )
+        LineString((3, 0), (3, 1), (3, 2)),
+        LineString((4, 0), (4, 1), (4, 2)),
+    ])
```
harrism, line 87: It's unclear how this works. What does the `3` in `"mlines": [1, 3]"` represent? Is that a length or an offset?

docs/source/api_docs/internals.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Polygon geometry includes `mpolygons` for MultiPolygons similar to the
+LineString geometry. Polygons are encoded using the same format as
+Shapefiles, with left-wound external rings and right-wound internal rings.
```
harrism, line 91: We should provide a reference to the ShapeFile format here if we aren't going to define the terms and document the format here.

docs/source/api_docs/spatial_indexing.rst
```diff
@@ -0,0 +1,16 @@
+Spatial Indexing
+----------------
+
+Spatial indexing functions provide blisteringly-fast on-GPU point-in-polygon
+operations.
```
harrism, line 5:
```suggestion
Spatial indexing functions provide GPU-accelerated point-in-polygon
and spatial join operations.
```

docs/source/api_docs/trajectory.rst
```diff
@@ -0,0 +1,14 @@
+Trajectory
+----------
+
+Trajectory functions make it easy to identify and group trajectories from point data.
```
harrism, line 4:
```suggestion
Functions for identifying and grouping trajectories from point data.
```

docs/source/index.rst
```diff
@@ -15,26 +15,35 @@ geometries.
 GeoArrow
 --------
 
-cuSpatial proposes a new GeoArrow format from the fruit of discussions with the GeoPandas team. GeoArrow is a packed columnar data format for the six fundamental geometry types: Point, MultiPoint, Lines, MultiLines, Polygons, and MultiPolygons. MultiGeometry is a possibility that may be implemented in the future. GeoArrow uses packed coordinate and offset columns to define objects, which enables very-fast copy between CPU, GPU, and NIC.
+cuSpatial proposes a new GeoArrow format from the fruit of discussions
+with the GeoPandas team. GeoArrow is a packed columnar data format
```
harrism, line 19: GeoArrow exists independently of cuSpatial. This makes it sound like we invented GeoArrow. I don't think this sentence belongs in our documentation. (CC @thomcom )

```suggestion
GeoArrow is a packed columnar data format
```

docs/source/index.rst
```diff
@@ -15,26 +15,35 @@ geometries.
...
+for the six fundamental geometry types:
+Point, MultiPoint, Lines, MultiLines, Polygons, and MultiPolygons.
+MultiGeometry is a possibility that may be implemented in the future.
+GeoArrow uses packed coordinate and offset columns to define objects,
+which enables very-fast copy between CPU, GPU, and NIC.
```
harrism, line 24:
```suggestion
which enables very fast copies between CPU, GPU, and NIC.
```

python/cuspatial/cuspatial/core/interpolate.py
```diff
@@ -49,6 +49,7 @@ class CubicSpline:
     However, cuSpatial massively outperforms scipy when many splines are fit
```
harrism, line 49:
```suggestion
    However, cuSpatial significantly outperforms scipy when many splines are fit
```

docs/source/api_docs/internals.rst
harrism: Perhaps link to the docs for the function where this is explained.

harrism (changes requested):
Thanks for cleaning up the docs and updating to the new format @isVoid !  I have some edit suggestions for the docs while you are at it.

docs/source/api_docs/geopandas_compatibility.rst
```diff
@@ -0,0 +1,18 @@
...
+-----------------------
+
+We support any geometry format supported by `GeoPandas`. Load geometry information from a `GeoPandas.GeoSeries` or `GeoPandas.GeoDataFrame`.
+
+    >>> gpdf = geopandas.read_file('arbitrary.txt')
+        cugpdf = cuspatial.from_geopandas(gpdf)
+
+or
+
+    >>> cugpdf = cuspatial.GeoDataFrame(gpdf)
```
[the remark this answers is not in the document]
isVoid: I don't see much difference in code - `from_geopandas` is a thin wrapper over the constructor. It dynamically dispatch to series/dataframe constructor depending on the geopandas object type.

docs/source/api_docs/internals.rst
isVoid: It seems like an offset. I think the key difference between `offsets` and `mlines` is that `offsets` is `N+1` in length, because all point in the point array belongs to some linestring. But `mlines` is `2N` format, becuase not all linestrings are multilinestrings, you need to explicitly specify the start and end (linestring) offset of each multilinestring. cc. @thomcom for detail.

ajschmidt8 (approved):

docs/source/api_docs/internals.rst
isVoid: Meanwhile, checkout the updated paragraph to see if that's clear.

docs/source/api_docs/geopandas_compatibility.rst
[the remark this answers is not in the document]
harrism: So is it preferred / faster?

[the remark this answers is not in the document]
isVoid: It depends on the usecase, `from_geopandas` can be used when user wants to convert any geopandas object into cuspatial objects, provided they don't know before hand the object type. On the other hand, if the user knows explicitly that the geopandas object is a geodataframe, then the constructor is preferred. This is similar to cudf's `cudf.from_pandas` and `cudf.DataFrame` ctor. I don't have enought usecase knowledge to know which one will come up more often in practice though.

isVoid:
Thanks to @ajschmidt8 , the documentation is now live for review at https://627d5be4c229560ba620c84d--docs-rapids-ai.netlify.app/api/cuspatial/nightly/ !

docs/source/api_docs/internals.rst
```diff
@@ -38,11 +38,11 @@ can be used.
 
 A `Multipoint` is a group of points, and is the second simplest GeoArrow 
 geometry type. It is identical to points, with the addition of a 
-`multipoints_offsets` buffer. The offsets buffer stores N+1 indices. The
+``multipoints_offsets`` buffer. The offsets buffer stores N+1 indices. The
```
harrism, line 41: Why the double back quotes? Markdown only requires single. Is this a Sphinx thing?

isVoid: This isn't markdown but reStructuredText (we don't like that and cuDF is converting all documentation to markdown). Per [reST doc](https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup), only double backticks is recognized as monospaced fonts and single backtick is application specific. In Sphinx, double backticks are rendered as inline highlighted code:
image
And single backtick are rendered as italics:
image

harrism: > (we don't like that and cuDF is converting all documentation to markdown)

So shouldn't we do the same in cuSpatial?

isVoid: I haven't investigated how many hurdles it takes to do all the conversion but would love to tackle that as a follow up to this. Meanwhile, perhaps the author of https://github.com/rapidsai/cudf/pull/10691 can explain a bit on this :) @shwina

harrism (approved):

harrism:
@thomcom could you please review this PR?

docs/source/index.rst
thomcom: That's fine. I had the first implementation, though GeoArrow diverged from my implementation a little as per the previous discussions. :D :D

docs/source/api_docs/internals.rst
thomcom: Heya, sorry for not being in this convo earlier. I'll be looking into dropping the 'bounding regions' in the `mlines` object to fit GeoArrow format correctly.

thomcom (approved):

harrism:
@gpucibot merge

PR #526, opened 2022-05-13
Refactor `linestring_distance` to header only API

isVoid:
This PR converts `pairwise_linestring_distance` into header only API.


cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,346 @@
...
+ * @brief Computes shortest distance between two segments (ab and cd) that
+ * doesn't intersect.
+ */
+template <typename T>
+T __device__ segment_distance_no_intersect_or_colinear(vec_2d<T> const& a,
+                                                       vec_2d<T> const& b,
+                                                       vec_2d<T> const& c,
+                                                       vec_2d<T> const& d)
+{
+  auto dist_sqr = std::min(std::min(point_to_segment_distance_squared(a, c, d),
```
harrism, line 158: This is definitely an older version.  No deferred sqrt...

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,346 @@
...
+                                          typename Cart2dB::value_type,
+                                          typename std::iterator_traits<OutputIt>::value_type>(),
+                "Inputs and output must be floating point types.");
+
+  static_assert(detail::is_same<T,
+                                typename Cart2dB::value_type,
+                                typename std::iterator_traits<OutputIt>::value_type>(),
+                "Inputs and output must be the same types.");
+
+  static_assert(detail::is_same<cartesian_2d<T>, Cart2dA, Cart2dB>(),
```
harrism, line 314: If this is the case, then you don't need both Cart2dA and Cart2dB -- just use the same type for both inputs.

cpp/tests/experimental/spatial/linestring_distance_test.cu
```diff
@@ -0,0 +1,153 @@
...
+                               a_cart2d.begin(),
+                               a_cart2d.end(),
+                               offset.begin(),
+                               b_begin,
+                               b_end,
+                               distance.begin());
+
+  EXPECT_EQ(distance, expected);
+}
+
```
harrism, line 151: Anytime we have a custom CUDA kernel or significant custom CUDA logic (like a non-trivial operator passed to a Thrust algorithm), I think we should have at least one test that causes the kernel to be invoked with more threads than there are in a single warp. This is especially true when the kernel involves cooperation between threads, as you do in this algorithm's kernel using atomic operations. So I suggest making a test that generates at least ~1000 data points.

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @tparam Cart2dItA iterator type for point array of the first linestring of each pair. Must meet
+ * the requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam Cart2dItB iterator type for point array of the second linestring of each pair. Must meet
+ * the requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIterator iterator type for offset array. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OutputIt iterator type for output array. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam std::iterator_traits<Cart2dItA>::value_type value type of `Cart2dItA`, must be
```
harrism, line 38: Why must it be? Why can't it be `vec_2d`?  And if it must be, then why is it a template parameter? I think you made a similar comment on one of my PRs.

harrism (changes requested):
Something seems off -- the added code is the old version of the implementation. The removed code is the new version.

cpp/include/cuspatial/experimental/linestring_distance.cuh
isVoid: I was under the impression that this is a geometric operation and we can't use this function with geographic types - all the math computations - from cross products to point to segment distance are only well-defined for cartesian coordinates. They won't get proper results if user start using them on geographical coordinates. I had originally believed that this is why we introduce strong types for geometric/geographic types, to prevent user from calling the API with the wrong types. If they want to use it with geographic coordinates, they should perform a projection first.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
isVoid: Unfortunately `Cart2dA` and `Cart2dB` are just two alias for the `value_type` of iterator `Cart2dItA` and `Cart2dItB`, note that they are parameters derived from existing template parameters. The test here is basically asserting that the value type for both iterators should be `cartesian_2d<T>`.

cpp/include/cuspatial/experimental/linestring_distance.cuh
harrism: OK. It may not be well-defined for lonlat_2d, but there may be some other type derived from vec_2d for which it is well-defined...  But I am indeed unsure of this, so we can leave it only defined for cartesian_2d for now.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism: I get that. But you only need one type parameter if they are both required to be the same type, right? See how I did it for haversine.  Single parameter for `Location` type:
https://github.com/rapidsai/cuspatial/blob/1f070d2589b7a4fb6d5299762630d6d4f9b843ae/cpp/include/cuspatial/experimental/haversine.cuh#L71-L81

Then check that both `Location` and the `value_type` of the B iterator are `lonlat_2d`:
https://github.com/rapidsai/cuspatial/blob/1f070d2589b7a4fb6d5299762630d6d4f9b843ae/cpp/include/cuspatial/experimental/detail/haversine.cuh#L70-L73

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,291 @@
...
+constexpr bool is_same()
+{
+  return std::conjunction_v<std::is_same<T, Ts>...>;
+}
+
+template <typename... Ts>
+constexpr bool is_floating_point()
+{
+  return std::conjunction_v<std::is_floating_point<Ts>...>;
+}
```
harrism, line 48: I like these utilities, and they would be useful in all of our APIs (I could have used them in haversine), so perhaps put them in type_utils.hpp?

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,291 @@
...
+  auto ac               = c - a;
+  auto r_numer          = det(ac, cd);
+  auto denom_reciprocal = 1 / denom;
+  auto r                = r_numer * denom_reciprocal;
+  auto s                = det(ac, ab) * denom_reciprocal;
+  if (r >= 0 and r <= 1 and s >= 0 and s <= 1) { return 0.0; }
+  return segment_distance_no_intersect_or_colinear(a, b, c, d);
+}
+
+/**
```
harrism, line 136: This shouldn't be in our public docs.

```suggestion
/*
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,291 @@
...
+  return std::conjunction_v<std::is_same<T, Ts>...>;
+}
+
+template <typename... Ts>
+constexpr bool is_floating_point()
+{
+  return std::conjunction_v<std::is_floating_point<Ts>...>;
+}
+
+/** @brief Get the index that is one-past the end point of linestring at @p linestring_idx
```
harrism, line 50: These detail functions don't need to be in the public docs.

```suggestion
/* @brief Get the index that is one-past the end point of linestring at @p linestring_idx
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,291 @@
...
+                             SizeType const& num_linestrings,
+                             SizeType const& num_points)
+{
+  return (linestring_idx == (num_linestrings - 1)
+            ? (num_points)
+            : *(linestring_offsets_begin + linestring_idx + 1)) -
+         1;
+}
+
+/**
```
harrism, line 68:
```suggestion
/*
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,291 @@
...
+  auto r  = dot(ac, ab);
+  auto bc = c - b;
+  // If the projection of `c` is outside of segment `ab`, compute point-point distance.
+  if (r <= 0 or r >= l_squared) { return std::min(dot(ac, ac), dot(bc, bc)); }
+  auto p  = a + (r / l_squared) * ab;
+  auto pc = c - p;
+  return dot(pc, pc);
+}
+
+/**
```
harrism, line 89:
```suggestion
/*
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,291 @@
...
+                                                       vec_2d<T> const& d)
+{
+  auto dist_sqr = std::min(std::min(point_to_segment_distance_squared(a, c, d),
+                                    point_to_segment_distance_squared(b, c, d)),
+                           std::min(point_to_segment_distance_squared(c, a, b),
+                                    point_to_segment_distance_squared(d, a, b)));
+  return dist_sqr;
+}
+
+/**
```
harrism, line 106:
```suggestion
/*
```

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ * `cuspatial::cartesian_2d`
+ * @tparam std::iterator_traits<Cart2dItB>::value_type value type of `Cart2dItB`, must be
+ * `cuspatial::cartesian_2d`
+ *
+ * @param linestring1_offsets_first begin of range of the offsets to the first linestring of each
```
harrism, line 43:
```suggestion
 * @param linestring1_offsets_first beginning of range of the offsets to the first linestring of each
```

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ * pair
+ * @param linestring1_offsets_last end of range of the offsets to the first linestring of each pair
+ * @param linestring1_points_first begin of range of the point of the first linestring of each pair
```
harrism, line 46:
```suggestion
 * @param linestring1_points_first beginning of range of the point of the first linestring of each pair
```

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ * @param linestring1_points_last end of range of the point of the first linestring of each pair
+ * @param linestring2_offsets_first begin of range of the offsets to the second linestring of each
```
harrism, line 48:
```suggestion
 * @param linestring2_offsets_first beginning of range of the offsets to the second linestring of each
```

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ * pair
+ * @param linestring2_points_first begin of range of the point of the second linestring of each pair
```
harrism, line 50:
```suggestion
 * @param linestring2_points_first beginning of range of the point of the second linestring of each pair
```

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ * @param linestring2_points_last end of range of the point of the second linestring of each pair
+ * @param distances_first begin to output array
```
harrism, line 52:
```suggestion
 * @param distances_first beginning iterator to output
```

cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -0,0 +1,80 @@
...
+ * @param stream Used for device memory operations and kernel launches.
```
harrism, line 53:
```suggestion
 * @param stream The CUDA stream to use for device memory operations and kernel launches.
```

cpp/include/cuspatial/utility/device_atomics.cuh
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace cuspatial {
+namespace detail {
+
+__device__ double atomicMin(double* addr, double val)
+{
+  unsigned long long int* address_as_ull = (unsigned long long int*)addr;
```
harrism, line 22: Should this be a memcpy instead of a C-style cast?

cpp/include/cuspatial/utility/device_atomics.cuh
```diff
@@ -0,0 +1,53 @@
...
+
+    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
+  } while (assumed != old);
+
+  return __longlong_as_double(old);
+}
+
+__device__ float atomicMin(float* addr, float val)
+{
+  unsigned int* address_as_ull = (unsigned int*)addr;
```
harrism, line 38: memcpy?

cpp/scripts/run-cmake-format.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# This script is a wrapper for cmakelang that may be used with pre-commit. The
```
harrism, line 3: separate PR?

harrism (changes requested):

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
isVoid: Since these are derived type traits, perhaps a better place is something like traits.hpp?

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
isVoid: On that note, I realized cuspatial doesn't have public documentation for cpp.

isVoid: If we do this, my vscode editor will lose highlight to all the entire section. This makes reading a bit harder. Can we use the annotation `@internal` in the section, and set doxygen configuration [`INTERNAL_DOC`](https://www.doxygen.nl/manual/config.html#cfg_internal_docs) to `NO` to hide the section? As described here: https://www.doxygen.nl/manual/commands.html#cmdinternal

cpp/scripts/run-cmake-format.sh
isVoid: Oops, accidentally commited this file in this PR. Reveted.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism: Sounds like a good solution.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,288 @@
...
+                                  OffsetIterator linestring2_offsets_first,
+                                  Cart2dItB linestring2_points_first,
+                                  Cart2dItB linestring2_points_last,
+                                  OutputIt distances_first,
+                                  rmm::cuda_stream_view stream)
+{
+  using T = typename std::iterator_traits<Cart2dItA>::value_type::value_type;
+
+  static_assert(detail::is_floating_point<T,
+                                          typename Cart2dB::value_type,
```
harrism, line 245: You are still using `Cart2dB` here, hence the compilation failure...

harrism (approved):
Great work on this!

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
isVoid: Also, in doxygen there's a configuration called `EXCLUDE_PATTERNS`, currently in cudf it includes `*/detail/*` that excludes all APIs if the absolute path contains `detail`.

harrism:
We need one more CPP codeowners review here. @trxcllnt @cwharris @jrhemstad do any of you have a few minutes to review?  Thanks.

cpp/include/cuspatial/utility/device_atomics.cuh
```diff
@@ -0,0 +1,53 @@
...
+  do {
+    assumed = old;
+    old     = atomicCAS(
+      address_as_ull, assumed, __double_as_longlong(std::min(val, __longlong_as_double(assumed))));
+
+    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
+  } while (assumed != old);
+
+  return __longlong_as_double(old);
+}
```
harrism, line 34: 1. Access `addr` as `double` before conversion to avoid strict aliasing violation (the other aliased accesses are using compiler intrinsics so any violations are harmless).
2. Remove c-style cast
3. Don't declare two variables on a single line.
4. Repeat for the other implementation below.

(Thanks @jrhemstad)

```
__device__ double atomicMin(double* addr, double val)
{
  double old_double = *addr;
  unsigned long long int* address_as_ull = reinterpret_cast<unsigned long long int*>(addr);
  unsigned long long int old   = __double_as_longlong(old_double);
  unsigned long long int assumed;

  do {
    assumed = old;
    old     = atomicCAS(
      address_as_ull, assumed, __double_as_longlong(std::min(val, __longlong_as_double(assumed))));

    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
  } while (assumed != old);

  return __longlong_as_double(old);
}
```

harrism (changes requested):
Just some cleanup for the atomics implementation.

harrism (approved):

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+ * to form the globally minimum distance between the linestrings.
+ *
+ * @tparam Cart2dItA Iterator to 2d cartesian coordinates. Must meet requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam Cart2dItB Iterator to 2d cartesian coordinates. Must meet requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIterator Iterator to linestring offsets. Must meet requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OutputIterator Iterator to output distances. Must meet requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
```
harrism, line 148:
```suggestion
 * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible and mutable.
```

harrism:
rerun tests

cpp/include/cuspatial/utility/device_atomics.cuh
```diff
@@ -0,0 +1,89 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace cuspatial {
+namespace detail {
+
+__device__ double atomicMin(double* addr, double val)
```
vyasr, line 20: All four of these operators should probably be implemented as specializations of a single templated function. It'll need ~5 template parameters (the value type, the bit view type, the two conversion functions between value and view types, and the actual operation to perform e.g. std::max), which is kind of gross, but it'll significantly reduce duplication and the pain of adding new operators if you need other atomics in the future.

cpp/include/cuspatial/utility/traits.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace detail {
+
+/**
+ * @internal
+ * @brief returns true if all types are the same.
+ */
+template <typename T, typename... Ts>
+constexpr bool is_same()
+{
+  return std::conjunction_v<std::is_same<T, Ts>...>;
```
vyasr, line 13: Nice to see folds getting some use!

cpp/tests/experimental/spatial/linestring_distance_test.cu
```diff
@@ -0,0 +1,190 @@
...
+TYPED_TEST(PairwiseLinestringDistanceTest, FromSeparateArrayInputs)
+{
+  using T       = TypeParam;
+  using CartVec = std::vector<cartesian_2d<T>>;
+
+  auto a_cart2d = rmm::device_vector<cartesian_2d<T>>{
+    CartVec({{0.0f, 0.0f}, {1.0f, 0.0f}, {2.0f, 0.0f}, {3.0f, 0.0f}, {4.0f, 0.0f}})};
+  auto b_cart2d = rmm::device_vector<cartesian_2d<T>>{
+    CartVec({{0.0f, 1.0f}, {1.0f, 1.0f}, {2.0f, 1.0f}, {3.0f, 1.0f}, {4.0f, 1.0f}})};
+  auto offset = rmm::device_vector<int32_t>{std::vector<int32_t>{std::vector<int32_t>{0}}};
```
vyasr, line 56: Why do we need this double-nested `std::vector` construction?

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+  auto l_squared = dot(ab, ab);
+  if (l_squared == 0) { return dot(ac, ac); }
+  auto r  = dot(ac, ab);
+  auto bc = c - b;
+  // If the projection of `c` is outside of segment `ab`, compute point-point distance.
+  if (r <= 0 or r >= l_squared) { return std::min(dot(ac, ac), dot(bc, bc)); }
+  auto p  = a + (r / l_squared) * ab;
+  auto pc = c - p;
+  return dot(pc, pc);
+}
```
vyasr, line 79: I love how simple these formulas are in 2D!

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+
+/**
+ * @internal
+ * @brief The kernel to compute point to linestring distance
+ *
+ * Each thread of the kernel computes the distance between a segment in a linestring in pair 1
+ * to a linestring in pair 2. For a segment in pair 1, the linestring index is looked up from
+ * the offset array and mapped to the linestring in the pair 2. The segment is then computed
+ * with all segments in the corresponding linestringin pair 2. This forms a local minima of the
```
vyasr, line 137:
```suggestion
 * with all segments in the corresponding linestring in pair 2. This forms a local minima of the
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+ * shortest distance, which is then combined with other segment results via an atomic operation
```
vyasr, line 138: Atomics are really an implementation detail IMO.
```suggestion
 * shortest distance, which is then combined with other segment results via an operation
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+                                  OutputIt distances_first,
+                                  rmm::cuda_stream_view stream)
+{
+  using T = typename std::iterator_traits<Cart2dItA>::value_type::value_type;
+
+  static_assert(
+    detail::is_floating_point<T,
+                              typename std::iterator_traits<Cart2dItB>::value_type::value_type,
+                              typename std::iterator_traits<OutputIt>::value_type>(),
+    "Inputs and output must be floating point types.");
```
vyasr, line 243: I'm not sure of the best wording, but ideally this should clarify that it's referring to the underlying type of the `cartesian_2` to differentiate from the third error check below.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+                                typename std::iterator_traits<Cart2dItB>::value_type::value_type,
+                                typename std::iterator_traits<OutputIt>::value_type>(),
+                "Inputs and output must be the same types.");
+
+  static_assert(detail::is_same<cartesian_2d<T>,
+                                typename std::iterator_traits<Cart2dItA>::value_type,
+                                typename std::iterator_traits<Cart2dItB>::value_type>(),
+                "Inputs must be cuspatial::cartesian_2d");
+
+  auto const num_string_pairs =
```
vyasr, line 255:
```suggestion
  auto const num_linestring_pairs =
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+    thrust::distance(linestring1_offsets_first, linestring1_offsets_last);
+  auto const num_linestring1_points =
+    thrust::distance(linestring1_points_first, linestring1_points_last);
+  auto const num_linestring2_points =
+    thrust::distance(linestring2_points_first, linestring2_points_last);
+
+  thrust::fill(rmm::exec_policy(stream),
+               distances_first,
+               distances_first + num_string_pairs,
```
vyasr, line 264:
```suggestion
               distances_first + num_linestring_pairs,
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+               std::numeric_limits<T>::max());
+
+  std::size_t constexpr threads_per_block = 64;
```
vyasr, line 267: We use `std::size_t` here while using `::size_t` in the kernel. We should probably be consistent.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+  auto const p1_idx = threadIdx.x + blockIdx.x * blockDim.x;
+  size_t const num_linestrings =
+    thrust::distance(linestring1_offsets_begin, linestring1_offsets_end);
+
+  size_t const linestring1_num_points =
+    thrust::distance(linestring1_points_begin, linestring1_points_end);
+  size_t const linestring2_num_points =
+    thrust::distance(linestring2_points_begin, linestring2_points_end);
+
+  if (p1_idx >= linestring1_num_points) { return; }
```
vyasr, line 190: This exit can technically happen a few lines earlier, but it probably doesn't matter.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+  auto ls1_end = endpoint_index_of_linestring(
+    linestring_idx, linestring1_offsets_begin, num_linestrings, linestring1_num_points);
+
+  if (p1_idx == ls1_end) {
+    // Current point is the end point of the line string.
+    return;
+  }
+
+  auto ls2_start = *(linestring2_offsets_begin + linestring_idx);
+  auto ls2_end   = endpoint_index_of_linestring(
```
vyasr, line 207:
```suggestion
  auto const ls2_start = *(linestring2_offsets_begin + linestring_idx);
  auto const ls2_end   = endpoint_index_of_linestring(
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -0,0 +1,284 @@
...
+  if (r <= 0 or r >= l_squared) { return std::min(dot(ac, ac), dot(bc, bc)); }
+  auto p  = a + (r / l_squared) * ab;
+  auto pc = c - p;
+  return dot(pc, pc);
+}
+
+/**
+ * @internal
+ * @brief Computes shortest distance between two segments (ab and cd) that
+ * doesn't intersect.
```
vyasr, line 84:
```suggestion
 * don't intersect.
```

vyasr (changes requested):

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism: The whole comment is implementation details. This is an internal function. Note the `@internal` at the top of the function comment. I think the word atomic should stay here, because algorithmically it doesn't make sense to just say "an operation".

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism:
```suggestion
    "Inputs and output must have floating point value type.");
```

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism: kernel should use the former.

harrism:
rerun tests

cpp/include/cuspatial/utility/device_atomics.cuh
harrism: Let's punt this to a future PR.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism: I think it's cleaner to keep the definitions of `linestring1_num_points` and `linestring2_num_points` together. The compiler can handle this optimization.

cpp/tests/experimental/spatial/linestring_distance_test.cu
harrism: Not needed. Fixed.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
vyasr: Fair enough, let's leave it then.

vyasr:
@gpucibot merge

PR #527, opened 2022-05-16
Fix various issues with CMake exports

vyasr:
This PR contains three fixes to how libcuspatial handles dependencies to enable its inclusion into other projects:
1. GDAL is now exported as a global target so that consumers automatically see the library in addition to inheriting the dependency.
2. cudf is now part of the build and export set for libcuspatial so that consumers can inherit the dependency.
3. cudftestutil is included in the export set and exported as a global target iff cuspatial is being built with tests. This change ensures that when tests are not being built cuspatial can be built against a libcudf that also does not have tests built.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -94,6 +94,7 @@ include(cmake/Modules/ConfigureCUDA.cmake)
 # find gdal
 rapids_find_package(
     GDAL REQUIRED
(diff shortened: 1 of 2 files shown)


harrism (approved):

trxcllnt (approved):

vyasr:
@gpucibot merge

PR #533, opened 2022-05-23
DEBUG

Ethyling:


vyasr:
Because of the arrow issues that we fixed in libcudf by switching to conda compilers unfortunately it looks like we can't really test this with the old compilers without using an upstream libcudf built against a pinned arrow version.

vyasr:
I think we can convert this to a real PR and merge it to get unblocked, then perhaps revisit as we continue. I think there's enough work happening to improve our build systems and CI that it may not be worthwhile to sink too much time into perfecting these setups right now.

vyasr:
I think we can convert this to a real PR and merge it to get unblocked, then perhaps revisit as we continue. I think there's enough work happening to improve our build systems and CI that it may not be worthwhile to sink too much time into perfecting these setups right now.

Issue #525, opened 2022-05-12
[DOC] Documented method and code location mismatches


isVoid:
During #523 , I noticed many methods in cuSpatial are documented in one place, but the actual definition is in another. For example:

**Location of incorrect documentation**
`polyline_bounding_boxes` is under ["spatial indexing" category](https://docs.rapids.ai/api/cuspatial/stable/api.html#cuspatial.polyline_bounding_boxes).  

But in code the methods is in `gis.py`: https://github.com/rapidsai/cuspatial/blob/ba0aed1e75ac37b7ba9723c68234438887f49101/python/cuspatial/cuspatial/core/gis.py#L308

`directed_hausdorff_distance` is under [`trajectories` category](https://docs.rapids.ai/api/cuspatial/stable/api.html#cuspatial.directed_hausdorff_distance), but was exposed in `gis.py` in code:

https://github.com/rapidsai/cuspatial/blob/ba0aed1e75ac37b7ba9723c68234438887f49101/python/cuspatial/cuspatial/core/gis.py#L29

**Describe the problems or issues found in the documentation**
A blind fix to the issue is that the methods should be in the file named after the same category in documentation. However a priori to the question is that we know the categories actually makes sense. Currently we have `spatial indexing`, `gis` and `trajectories` to sort computing related APIs, `internals` to document storage foramts, `geopandas computatbility` and `IO` to document interops. 

The computing related APIs are slightly confusing since `gis` and `trajectories` are two slightly overlapped concepts. In the bigger picture the category of computing methods should relate to the problem we try to solve with cuSpatial as a package. I'd like to get some opinions from the community.

Besides the above, geoArrow package currently is placed within `cuspatial.geometry` module which is eccentric, in two ways. First the name implies geometric operations, but a data structure was defined instead. Second, a well defined data type system is the center piece to a library. Though geoarrow is largely a work in progress definition and cuSpatial will remain it's functional interface for foreseeable future, we should experiment using geoArrow/geodataframe to organize our input, instead of using legacy SoA array format. This is an early thought and I haven't done any verification work yet. The biggest obstacle I can imagine is that the legacy APIs may define the buffer layout differently to geoArrow (`MultiLineString`s being an example). Standardizing these inputs may require significant amount of rework of existing APIs.

thomcom:
[...]

PR #534, opened 2022-05-23
Add Doxygen Documentation for `libcuspatial`

isVoid:
This PR adds doxygen documentation for libcuspatial.

In general, groupings to c++ APIs per our discussion on cuspatial method categories. There are 4 main categories:
- spatial APIs
- trajectory APIs
- spatial indexing
- spatial join

I took liberty on sub-categories and is subject to discussion and change, as part of discussion in #525 .

Many docstring for APIs were missing and are added in this PR. Such as factory methods in `type_utils.hpp`.

Some docstrings broke when being rendered in pages, such as `point_in_polygon`, these were also fixed.


cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -41,6 +41,38 @@ struct vec_2d_to_tuple {
...
+ * to create iterators from data in various format.
+ *
+ * @{
+ * @file
+ *
+ * @copydetails type_factories
+ */
+
+/**
+ * @brief Create an iterator of `vec_2d` from two input interators
```
harrism, line 60:
```suggestion
 * @brief Create an iterator to `vec_2d` data from two input iterators.
 *
 * Interleaves x and y coordinates from separate iterators into a single iterator to x-y coordinates.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +84,69 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
   return thrust::make_transform_iterator(zipped, detail::tuple_to_vec_2d<T, VectorType>());
 }
 
+/**
+ * @brief Create an iterator of `lonlat_2d` from two input interators
+ *
```
harrism, line 89:
```suggestion
 * @brief Create an iterator to `lonlat_2d` data from two input iterators.
 *
 * Interleaves longitude and latitude from separate iterators into a single iterator to lon/lat coordinates.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +84,69 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
...
+ * @tparam FirstIter Iterator type to the first component of `lonlat_2d`. Must meet the requirements
+ * of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam SecondIter Iterator type to the second component of `lonlat_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @param first Iterator of beginning of `lonlat_2d::x`
+ * @param second Iterator of beginning of `lonlat_2d::y`
```
harrism, line 95:
```suggestion
 * @param first Iterator to beginning of `lonlat_2d::x`
 * @param second Iterator to beginning of `lonlat_2d::y`
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +84,69 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
...
 template <typename FirstIter, typename SecondIter>
 auto make_lonlat_iterator(FirstIter first, SecondIter second)
 {
   using T = typename std::iterator_traits<FirstIter>::value_type;
   return make_vec_2d_iterator<lonlat_2d<T>>(first, second);
 }
 
+/**
+ * @brief Create an iterator of `cartesian_2d` from two input interators
+ *
```
harrism, line 112:
```suggestion
 * @brief Create an iterator to `cartesian_2d` data from two input iterators.
 *
 * Interleaves x and y coordinates from separate iterators into a single iterator to x-y coordinates.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +84,69 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
...
+ * @tparam FirstIter Iterator type to the first component of `cartesian_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam SecondIter Iterator type to the second component of `cartesian_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @param first Iterator of beginning of `cartesian_2d::x`
+ * @param second Iterator of beginning of `cartesian_2d::y`
```
harrism, line 118:
```suggestion
 * @param first Iterator to beginning of `cartesian_2d::x`
 * @param second Iterator to beginning of `cartesian_2d::y`
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +84,69 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
...
+ */
 template <typename FirstIter, typename SecondIter>
 auto make_cartesian_2d_iterator(FirstIter first, SecondIter second)
 {
   using T = typename std::iterator_traits<FirstIter>::value_type;
   return make_vec_2d_iterator<cartesian_2d<T>>(first, second);
 }
 
+/**
+ * @brief Create an output iterator to `vec_2d` from two output interators
```
harrism, line 134:
```suggestion
 * @brief Create an output iterator to `vec_2d` data from two output iterators.
 *
 * Creates an output iterator from separate iterators to x and y data to which
 * can be written interleaved x/y data. This allows using two separate arrays of
 * output data with APIs that expect an iterator to structured data.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +84,69 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
...
+ *
+ * @tparam VectorType cuSpatial vector type, must be `vec_2d`, `lonlat_2d` or `cartesian_2d`
+ * @tparam FirstIter Iterator type to the first component of `vec_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ * @tparam SecondIter Iterator type to the second component of `vec_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ * @param first Iterator of beginning of `vec_2d::x`
+ * @param second Iterator of beginning of `vec_2d::y`
```
harrism, line 142:
```suggestion
 * @param first Iterator to beginning of `x` data.
 * @param second Iterator to beginning of `y` data.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -75,18 +156,54 @@ auto make_zipped_vec_2d_output_iterator(FirstIter first, SecondIter second)
                                                 detail::vec_2d_to_tuple<T, VectorType>());
 }
 
+/**
+ * @brief Create an output iterator to `lonlat_2d` from two output interators
+ *
```
harrism, line 161:
```suggestion
 * @brief Create an output iterator to `lonlat_2d` from two output iterators.
 *
 * Creates an output iterator from separate iterators to longitude and latitude data
 * to which can be written interleaved longitude/latitude data. This allows using two 
 * separate arrays of output data with APIs that expect an iterator to interleaved
 * data.
 *
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -75,18 +156,54 @@ auto make_zipped_vec_2d_output_iterator(FirstIter first, SecondIter second)
...
 template <typename FirstIter, typename SecondIter>
 auto make_zipped_lonlat_output_iterator(FirstIter first, SecondIter second)
 {
   using T = typename std::iterator_traits<FirstIter>::value_type;
   return make_zipped_vec_2d_output_iterator<lonlat_2d<T>>(first, second);
 }
 
+/**
+ * @brief Create an output iterator to `cartesian_2d` from two output interators
+ *
```
harrism, line 184:
```suggestion
 * @brief Create an output iterator to `cartesian_2d` from two output iterators.
 *
 * Creates an output iterator from separate iterators to x and y data to which
 * can be written interleaved x/y data. This allows using two separate arrays of
 * output data with APIs that expect an iterator to structured data.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -75,18 +156,54 @@ auto make_zipped_vec_2d_output_iterator(FirstIter first, SecondIter second)
...
+ * @tparam FirstIter Iterator type to the first component of `cartesian_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ * @tparam SecondIter Iterator type to the second component of `cartesian_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ * @param first Iterator of beginning of `cartesian_2d::x`
+ * @param second Iterator of beginning of `cartesian_2d::y`
```
harrism, line 190:
```suggestion
 * @param first Iterator to beginning of `x` data.
 * @param second Iterator of beginning of `y` data.
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -75,18 +156,54 @@ auto make_zipped_vec_2d_output_iterator(FirstIter first, SecondIter second)
...
 
+/**
+ * @brief Create an output iterator to `lonlat_2d` from two output interators
+ *
+ * @tparam FirstIter Iterator type to the first component of `lonlat_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ * @tparam SecondIter Iterator type to the second component of `lonlat_2d`. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ * @param first Iterator of beginning of `lonlat_2d::x`
+ * @param second Iterator of beginning of `lonlat_2d::y`
```
harrism, line 167:
```suggestion
 * @param first Iterator to beginning of longitude data.
 * @param second Iterator of beginning of latitude data.
```

cpp/include/cuspatial/haversine.hpp
```diff
@@ -24,7 +24,9 @@
 namespace cuspatial {
 
 /**
- * brief Compute haversine distances between points in set A to the corresponding points in set B.
+ * @brief Compute haversine distances between points in set A to the corresponding points in set B.
```
harrism, line 27:
```suggestion
 * @brief Compute haversine distances between points in set A and the corresponding points in set B.
```

cpp/include/cuspatial/types.hpp
```diff
@@ -26,6 +26,13 @@
 
 namespace cuspatial {
 
+/**
+ * @addtogroup type_definition_operator
+ * @brief Type declaration for cuspatial
```
harrism, line 31:
```suggestion
 * @brief Type declarations for cuspatial
```

cpp/include/cuspatial/types.hpp
```diff
@@ -26,6 +26,13 @@
...
+ * @{
+ * @file
+ */
+
 /**
  * @brief A timestamp
```
harrism, line 37: I don't think this struct is used anymore. We should consider removing it and this file.

cpp/include/doxygen_groups.h
```diff
@@ -0,0 +1,50 @@
...
+ * @brief Doxygen group definitions
+ */
+
+// This header is only processed by doxygen and does
+// not need to be included in any source file.
+// Below are the main groups that doxygen uses to build
+// the Modules page in the specified order.
+//
+// To add a new API to an existing group, just use the
+// @ingroup tag to the API's doxygen comment.
```
harrism, line 28:
```suggestion
// @ingroup tag in the API's doxygen comment.
```

cpp/include/doxygen_groups.h
```diff
@@ -0,0 +1,50 @@
...
+// Add a new group by first specifying in the hierarchy below.
```
harrism, line 29: Where can we add more documentation for each group? These descriptions are very brief.

harrism (changes requested):
This is awesome, what an improvement!  Thanks very much. I just have minor fixes.

isVoid:
Live review link: https://628cddb48ad16b199ab6794f--docs-rapids-ai.netlify.app/api/libcuspatial/22.06/

cpp/include/doxygen_groups.h
isVoid: Great question, I believe doxygen is very liberal with where the group documentation should be. You can write the docs anywhere inside the group, which includes the lines immediately after the `@defgroup` or anywhere within the group block (between `@{` and `@}`).

Libcudf adds module description by in-lining them after each `@addtogroup` decorator.
Unlike libcudf where the modules are quite well organized with the file structures, libcuspatial is a bit fragmented atm. For example the distances computation API are scattered in several header files. With this in mind, I moved all group descriptions to `doxygen_groups.h`. This provides a nice overview to the structure of the entire code base and reduces large group docs inlined within codes. Let me know if the new organization works better.

cpp/include/cuspatial/types.hpp
isVoid: Would you like to do that in this PR?

harrism (approved):

cpp/include/cuspatial/distances/linestring_distance.hpp
```diff
@@ -78,15 +83,16 @@ namespace cuspatial {
...
- * @param linestring2_points_x x-component of points in the first linestring of each pair.
- * @param linestring2_points_y y-component of points in the first linestring of each pair.
- * @param mr Device memory resource used to allocate the returned column's device memory.
- * @return A column of shortest distances between each pair of linestrings.
+ * @param linestring1_offsets Indices of the first point of the first linestring of each pair
+ * @param linestring1_points_x x-components of points in the first linestring of each pair
+ * @param linestring1_points_y y-component of points in the first linestring of each pair
+ * @param linestring2_offsets Indices of the first point of the second linestring of each pair
+ * @param linestring2_points_x x-component of points in the first linestring of each pair
+ * @param linestring2_points_y y-component of points in the first linestring of each pair
```
vyasr, line 93:
```suggestion
 * @param linestring2_points_x x-component of points in the second linestring of each pair
 * @param linestring2_points_y y-component of points in the second linestring of each pair
```

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -23,6 +23,11 @@
 namespace cuspatial {
 
 namespace detail {
+
+/**
+ * @internal
+ * @brief Helper to convert a tuple of elements into a `vec_2d`
```
vyasr, line 29: This question is out of scope for this PR, but why a tuple and not a pair? Issues with thrust::pair?

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -52,20 +86,77 @@ auto make_vec_2d_iterator(FirstIter first, SecondIter second)
   return thrust::make_transform_iterator(zipped, detail::tuple_to_vec_2d<T, VectorType>());
 }
 
+/**
+ * @brief Create an iterator to `lonlat_2d` data from two input iterators.
+ *
+ * Interleaves longitude and latitude from separate iterators into a single iterator to lon/lat
+ * coordinates.
+ * @tparam FirstIter Iterator type to the first component of `lonlat_2d`. Must meet the requirements
```
vyasr, line 94: It might be nice to state explicitly that the first component is the longitude and the second is the latitude. I could definitely see users defaulting to think that latitude is "first" based on just colloquial usage of "latitude and longitude".

cpp/include/cuspatial/experimental/type_utils.hpp
```diff
@@ -41,6 +50,31 @@ struct vec_2d_to_tuple {
...
+ * @tparam VectorType cuSpatial vector type, must be `vec_2d`, `lonlat_2d` or `cartesian_2d`
+ * @tparam FirstIter Iterator type to the first component of `vec_2d`. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam SecondIter Iterator type to the second component of `vec_2d`. Must meet the requirements
+ * of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @param first Iterator to beginning of `vec_2d::x`
+ * @param second Iterator to beginning of `vec_2d::y`
+ * @return Iterator to `vec_2d`
+ *
+ * @pre `first` and `second` must iterate on same data type.
```
vyasr, line 73: Maybe out of scope for this PR, but if first and second must iterate on the same type why have two template parameters `FirstIter` and `SecondIter`? Am I misunderstanding this condition?

cpp/include/cuspatial/polygon_bounding_box.hpp
```diff
@@ -25,6 +25,8 @@ namespace cuspatial {
 /**
  * @brief Compute minimum bounding boxes for a set of polygons.
  *
+ * @ingroup spatial_relationship
```
vyasr, line 28: `spatial_relationship` isn't my favorite. It feels rather vague, but I assume you all discussed this already in your meetings and I don't immediately have a better suggestion so I'm fine getting this merged and maybe pondering future improvements.

cpp/include/cuspatial/types.hpp
vyasr: If we're going to remove this anyway I wouldn't bother adding docs. The current group name is uninformative will only confuse users, especially if they'll never need to know about it.

cpp/include/cuspatial/utility/vec_2d.hpp
```diff
@@ -1,14 +1,33 @@
...
+ * limitations under the License.
+ */
+
 #pragma once
 #include <cuspatial/types.hpp>
 
 namespace cuspatial {
 
+/**
+ * @addtogroup type_definition_operator
```
vyasr, line 23: What is `type_definition_operator` intended to denote? Why not just `types`? I see that these types are defined in `utility` instead of directly in `types.hpp`, I guess that's because with the header-only API it will be possible for users to never interact directly with these objects? A simpler name like `types` or `coordinate_types` would be more useful.

cpp/include/doxygen_groups.h
```diff
@@ -0,0 +1,113 @@
...
+ *          @brief Distance computation APIs
+ *
+ *          @file linestring_distance.hpp
+ *          @file hausdorff.hpp
+ *          @file haversine.hpp
+ *          @file haversine.cuh
+ *      @}
+ *      @defgroup spatial_relationship Spatial Relationship
+ *      @{
+ *          @brief APIs related to spatial relationship
```
vyasr, line 57: Is there a less vague way to describe what kinds of "relationships" are possible here? Maybe just a couple of examples by adding "such as ..."? Distances? Overlaps? Etc...?

cpp/include/doxygen_groups.h
```diff
@@ -0,0 +1,113 @@
...
+ *
+ *          @file point_in_polygon.hpp
+ *          @file polygon_bounding_box.hpp
+ *          @file polyline_bounding_box.hpp
+ *          @file spatial_window.hpp
+ *      @}
+ *      @defgroup cubic_spline Cubic Spline
```
vyasr, line 64: This name is highly specific. I would prefer a namespace for "splines", or even better "interpolation" so that there would be a natural place to put future alternatives to splines (if and when they were added).

cpp/include/doxygen_groups.h
```diff
@@ -0,0 +1,113 @@
...
+ *      @file point_quadtree.hpp
+ * @}
+ * @defgroup spatial_join Spatial Join
+ * @{
+ *      @brief APIs related to spatial join
+ *      @file spatial_join.hpp
+ * @}
+ * @defgroup cuspatial_types Types
+ * @{
+ *      @defgroup type_definition_operator Definition and Operators
```
vyasr, line 87: I don't think this group is necessary. The types can live in the top level of the Types group. The factories can be in a subgroup.

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -22,6 +22,11 @@
 
 namespace cuspatial {
 
+/**
+ * @addtogroup spatial_join
+ * @{
+ */
+
```
vyasr, line 29: I can't comment on it since it's not part of the diff, but I don't think the return statement on line 55 is going to render the way that you want. It'll just flatten it all into a single line. That may be something that you improve in a future PR, though, instead of trying to fix it here.

vyasr (changes requested):
I have some minor requests for changes and some points for discussion. I'd probably be OK with approving this PR even if you decide not to change any of them, but I think they are at least worth quick discussions before finalizing.

cpp/include/cuspatial/experimental/type_utils.hpp
cwharris: why is a pair preferable to a tuple?

cpp/include/cuspatial/experimental/type_utils.hpp
cwharris: They must iterator on the same value type, but the iterator types may differ.

Consider these two values which could be iterators with the same value type:
```
float* values_a;

auto values_b = thrust::make_transform_iterator(
  thrust::make_counting_iterator(0),
  [](auto idx){ return static_cast<float>(idx);
});

cpp/include/cuspatial/utility/vec_2d.hpp
cwharris: `alignas` is a function that is used to calculate some things for defining aligned types, so `type_definition_operator` seems fine? :shrug: `types` seems conflated and `coordinate_types` seems inappropriate.

cpp/include/doxygen_groups.h
cwharris: We'd be getting ahead of ourselves if we make this more generic. The name seems perfect for this version. Why not change it later when we have more information?

cwharris (approved):

cwharris (approved):

cpp/include/cuspatial/experimental/type_utils.hpp
harrism: Correct. An STL example is the two-input flavor of `transform`. This enables the use of custom or fancy iterators, as well as raw pointers, etc.

cpp/include/cuspatial/polygon_bounding_box.hpp
harrism: We haven't discussed. The names of groups was an item for discussion in this PR.

cpp/include/cuspatial/experimental/type_utils.hpp
vyasr: Preferable is both subjective and circumstantial. As far as subjective arguments, on one hand we're making a 2d vector, which is exactly a pair of values, so it seems intuitive to use a pair. On the other hand I guess you could argue that you might want to generalize this to other dimensionality and returning a tuple across all dimensions would be more consistent. It's also convenient to use `pair.first/second` accessors rather than having to call a method.

As far as circumstantial arguments, the only real important thing that I can see is that in some situations you might care that pairs are standard layout while tuples are not. I don't think that really matters for us here.

I don't particularly think that we need to change this, just thought I'd check.

cpp/include/cuspatial/experimental/type_utils.hpp
vyasr: Ah of course I didn't realize this was referring to the underlying data type, not the iterator type.

cpp/include/cuspatial/polygon_bounding_box.hpp
vyasr: Given the time constraints, I'd say we just leave this as is and revisit in 22.08?

cpp/include/cuspatial/utility/vec_2d.hpp
vyasr: I'm not sure what you're getting at with the `alignas`. That's not a function defined in this namespace, that's just a specifier for the `vec_2d` type. What is being conflated here?

vyasr (approved):

vyasr:
I've addressed my own comments on this PR and I approve. I think we can continue further discussions on the appropriate groups etc in the future, but those shouldn't block merging of this PR.

vyasr:
rerun tests

harrism:
@gpucibot merge

PR #543, opened 2022-05-31
Change build.sh to find C++ library by default and avoid shadowing CMAKE_ARGS

vyasr:
Discussions with the team indicated a general preference for having build.sh default to finding the C++ library by default (and erroring if one is not found) rather than building a new copy of libcuspatial when building cuspatial Python. This PR also avoids shadowing the `CMAKE_ARGS` environment variable in build.sh, which is important for conda-forge compatibility.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -116,9 +116,9 @@ if hasArg tests; then
 fi
 
 # Append `-DFIND_CUSPATIAL_CPP=ON` to CMAKE_ARGS unless a user specified the option.
-#if [[ "${CMAKE_ARGS}" != *"DFIND_CUSPATIAL_CPP"* ]]; then
-#    CMAKE_ARGS="${CMAKE_ARGS} -DFIND_CUSPATIAL_CPP=ON"
-#fi
+if [[ "${CMAKE_ARGS}" != *"DFIND_CUSPATIAL_CPP"* ]]; then
+    CMAKE_ARGS="${CMAKE_ARGS} -DFIND_CUSPATIAL_CPP=ON"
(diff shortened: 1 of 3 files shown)


vyasr:
I know that we're in burn down, so I can push this to 22.08 if we prefer.

vyasr:
rerun tests

harrism (approved):

ajschmidt8 (approved):

vyasr:
The problem here is an issue with scikit-build, specifically that calling `python setup.py build_ext ... -- -DCMAKE_INSTALL_PREFIX=...` doesn't work correctly. During the `build_ext` phase scikit-build actually performs a CMake installation, just from `_skbuild/.../cmake-build`->`_skbuild/.../cmake-install` rather than into any real installation location. Then, if you request a `--inplace` install it will copy those files from `cmake-install` into the source tree, whereas if you do a `python setup.py install - -DCMAKE_INSTALL_PREFIX=...` it will install into the desired install prefix. The problem is that if you specify an install prefix during the `build_ext` phase it will see that value but then raise an error saying that you're installing into an invalid location. I'll follow up with the scikit-build devs to see why that's the case, but until then we probably need to add some extra argument parsing logic.

vyasr:
I'm going to bump this to 22.08, we're blocked by https://github.com/conda-forge/ctng-compiler-activation-feedstock/issues/77 or a suitable alternate resolution if that doesn't move forward..

vyasr:
@gpucibot merge

PR #544, opened 2022-06-01
Fix out of bounds access in spline interpolation

vyasr:
The cubic spline API supports fitting many splines at once. Each spline's coefficients are stored on construction of the `CubicSpline` object, and when users then specify a set of points at which to evaluate that spline they can specify which spline to evaluate for each of the input points. When a user does not specify that information, the call operator defaults to assuming that only a single spline exists and interpolates all the data using that spline by setting all input points to use spline 0. Currently, however, the code for that creating that default list only making it as long as the number of knots in the spline instead of making it as long as the set of input points to evaluate the spline at. This leads to out of bounds memory accesses when that curve is fit.

### DIFF
--- a/python/cuspatial/cuspatial/core/interpolate.py
+++ b/python/cuspatial/cuspatial/core/interpolate.py
@@ -184,7 +184,7 @@ class CubicSpline:
                 self.groups = groups.astype("int32")
             else:
                 self.groups = Series(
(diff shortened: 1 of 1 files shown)


harrism (approved):

harrism:
@gpucibot merge

PR #554, opened 2022-06-03
Remove `itstimestamp` and `types.hpp`

### LINKED ISSUE #540: Remove no-longer-used struct its_timestamp
I don't think this struct is used anymore. We should consider removing it and the containing file (types.hpp).

_Originally posted by @harrism in https://github.com/rapidsai/cuspatial/pull/534#discussion_r880065153_

isVoid:
This PR removes `itstimestamp` and `types.hpp` as they are not used in current code base. closes #540


cpp/include/cuspatial/utility/vec_2d.hpp
```diff
@@ -15,10 +15,14 @@
...
 #pragma once
-#include <cuspatial/types.hpp>
-
 namespace cuspatial {
 
+#ifdef __CUDACC__
+#define CUSPATIAL_HOST_DEVICE __host__ __device__
+#else
+#define CUSPATIAL_HOST_DEVICE
+#endif
```
harrism, line 24: I think this should be put somewhere more central. It's a bit hidden in this file.

harrism (changes requested):
Thanks for the fix. Perhaps we need a cuda_utils.hpp or somewhere like that to put the host/device macro.

cpp/include/cuspatial/utility/vec_2d.hpp
isVoid: Moved.

harrism (approved):

isVoid:
rerun tests

isVoid:
@gpucibot merge

PR #558, opened 2022-06-08
Pairwise Point to Point Distance; Rename Folder `distances` to `distance`

isVoid:
Implements 2D point-point L2 distance. Also renamed folder `distances` into `distance`

Contributes to #231


cpp/include/cuspatial/distances/point_distance.hpp
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @ingroup distance
+ * @brief Compute pairwise point to point cartesian distance
+ *
+ * Computes cartesian distances between points.
+ *
+ * @param points1_x Column of x coordinates to the first point in pair
+ * @param points1_y Column of y coordinates to the first point in pair
+ * @param points2_x Column of x coordinates to the second point in pair
+ * @param points2_y Column of y coordinates to the second point in pair
```
harrism, line 33:
```suggestion
 * @brief Compute pairwise point-to-point Cartesian distance.
 *
 * @param points1_x Column of x coordinates of the first point in each pair
 * @param points1_y Column of y coordinates of the first point in each pair
 * @param points2_x Column of x coordinates of the second point in each pair
 * @param points2_y Column of y coordinates of the second point in each pair
```

cpp/include/cuspatial/distances/point_distance.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * @param stream The CUDA stream to use for device memory operations and kernel launches
+ * @return Column of distances
```
harrism, line 35:
```suggestion
 * @return Column of distances between each pair of input points
```

cpp/include/cuspatial/experimental/detail/point_distance.cuh
```diff
@@ -0,0 +1,66 @@
...
+  using T = typename std::iterator_traits<Cart2dItA>::value_type::value_type;
+
+  static_assert(
+    detail::is_floating_point<T,
+                              typename std::iterator_traits<Cart2dItB>::value_type::value_type,
+                              typename std::iterator_traits<OutputIt>::value_type>(),
+    "Inputs and output must be floating point types.");
+
+  static_assert(detail::is_same<T,
+                                typename std::iterator_traits<Cart2dItB>::value_type::value_type,
```
harrism, line 46: Could remove this term since it's covered by the combination of previous and next assertions...

```suggestion
```

cpp/include/cuspatial/experimental/point_distance.cuh
```diff
@@ -0,0 +1,61 @@
...
+
+#include <rmm/cuda_stream_view.hpp>
+
+namespace cuspatial {
+
+/**
+ * @ingroup distance
+ * @copybrief cuspatial::pairwise_point_distance
+ *
+ * Computes cartesian distances between points.
```
harrism, line 27: Redundant with brief. 
```suggestion
```

cpp/include/cuspatial/experimental/point_distance.cuh
```diff
@@ -0,0 +1,61 @@
...
+ * @tparam Cart2dItB iterator type for point array of the second point of each pair. Must meet
+ * the requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OutputIt iterator type for output array. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI], be mutable and be device-accessible.
+ *
+ * @param points1_first beginning of range of the point of the first point of each
+ * pair
+ * @param points1_last end of range of the point of the first point of each pair
+ * @param points2_first beginning of range of the point of the second point of each
+ * pair
```
harrism, line 40:
```suggestion
 * @param points1_first beginning of range of the first points of each pair
 * @param points1_last end of range of the first points of each pair
 * @param points2_first beginning of range of the second points of each pair
```

cpp/include/cuspatial/experimental/point_distance.cuh
```diff
@@ -0,0 +1,61 @@
...
+ * @param distances_first beginning iterator to output
+ * @param stream The CUDA stream to use for device memory operations and kernel launches
+ * @return Output iterator to one past the last element in the output range
+ *
+ * @pre all input iterators for coordinates must have `cuspatial::cartesian_2d` type.
```
harrism, line 45:
```suggestion
 * @pre all input iterators for coordinates must have a `value_type` of `cuspatial::cartesian_2d`.
```

cpp/src/spatial/point_distance.cu
```diff
@@ -0,0 +1,110 @@
...
+};
+
+std::unique_ptr<cudf::column> pairwise_point_distance(cudf::column_view const& points1_x,
+                                                      cudf::column_view const& points1_y,
+                                                      cudf::column_view const& points2_x,
+                                                      cudf::column_view const& points2_y,
+                                                      rmm::cuda_stream_view stream,
+                                                      rmm::mr::device_memory_resource* mr)
+{
+  CUSPATIAL_EXPECTS(points1_x.size() == points1_y.size() and
```
harrism, line 78: Also expect that there are no nulls?

cpp/tests/spatial/point_distance_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+  auto y2 = fixed_width_column_wrapper<T>{};
+
+  auto expect = fixed_width_column_wrapper<T>{};
+
+  auto got = pairwise_point_distance(x1, y1, x2, y2);
+
+  CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(expect, *got);
+}
+
+TYPED_TEST(PairwisePointDistanceTest, OnePair)
```
harrism, line 55: This tests nothing specific to the cuDF API, so could probably be eliminated.

harrism (changes requested):
Looks good. Mostly doc suggestions.

isVoid:
rerun tests

cpp/include/cuspatial/experimental/detail/point_distance.cuh
isVoid: Ah that's right, technically it's covered by the next assertion - since the `cartesian_2d` type is dependent on its template parameter, if the value type of the inputs mismatches, it would also fails the next assertion.

isVoid:
@thomcom that was my plan, given the constraints we have, I think it's better to reduce the size of PR and split into smaller pieces. What do you think?

thomcom:
That works, do you want to create an issue to track it? I don't think it should get overlooked.

thomcom (approved):

cpp/include/cuspatial/distances/point_distance.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * @brief Compute pairwise point to point cartesian distance
+ *
+ * @param points1_x Column of x coordinates to the first point in each pair
+ * @param points1_y Column of y coordinates to the first point in each pair
+ * @param points2_x Column of x coordinates to the second point in each pair
+ * @param points2_y Column of y coordinates to the second point in each pair
+ * @param stream The CUDA stream to use for device memory operations and kernel launches
+ * @return Column of distances between each pair of input points
+ */
+std::unique_ptr<cudf::column> pairwise_point_distance(
```
harrism, line 35: For discussion: thinking forward about naming: 

point-point distance: `pairwise_point_distance`
linestring-linestring distance; `pairwise_linestring_distance`
point-linestring distance: `pairwise_point_linestring_distance` (or is it `pairwise_linestring_point_distance`?)

Should we standardize on the number of geometries in the name of the function? E.g. should point-point distance be called `pairwise_point_point_distance`? Or is it obvious that `point` means `point-to-point`?  I kind of think it is, and it makes for shorter names.

Second, how do we order the geometries? Should the more complex geometry be first, or second?

What if there is a directed distance metric (e.g. Hausdorff is directed -- asymmetric). Do we need to support both point-linestring and linestring-point? I think we do need it one way or another...

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
```diff
@@ -241,15 +241,13 @@ void pairwise_linestring_distance(OffsetIterator linestring1_offsets_first,
...
-                                typename std::iterator_traits<Cart2dItB>::value_type::value_type,
-                                typename std::iterator_traits<OutputIt>::value_type>(),
+  static_assert(detail::is_same<T, typename std::iterator_traits<OutputIt>::value_type>(),
                 "Inputs and output must have the same value type.");
 
   static_assert(detail::is_same<cartesian_2d<T>,
                                 typename std::iterator_traits<Cart2dItA>::value_type,
                                 typename std::iterator_traits<Cart2dItB>::value_type>(),
-                "Inputs must be cuspatial::cartesian_2d");
+                "Input types mismatches or input types are not cuspatial::cartesian_2d");
```
harrism, line 250: I think the former error message was clearer and sufficient, since the mismatch of underlying float type is handled by the previous assertion.

```suggestion
                "Input types must be cuspatial::cartesian_2d");
```

cpp/include/cuspatial/experimental/detail/point_distance.cuh
```diff
@@ -0,0 +1,64 @@
...
+                              typename std::iterator_traits<OutputIt>::value_type>(),
+    "Inputs and output must be floating point types.");
+
+  static_assert(detail::is_same<T, typename std::iterator_traits<OutputIt>::value_type>(),
+                "Inputs and output must be the same types.");
+
+  static_assert(detail::is_same<cartesian_2d<T>,
+                                typename std::iterator_traits<Cart2dItA>::value_type,
+                                typename std::iterator_traits<Cart2dItB>::value_type>(),
+                "Input types mismatches or input types are not cuspatial::cartesian_2d");
```
harrism, line 51:
```suggestion
                "Input types must be cuspatial::cartesian_2d");
```

harrism (changes requested):
Just a couple more doc / error message suggestions.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
isVoid: I think the former assertion only checks whether the input types are of floating types, it's still only possible to mix double with floats? 
https://godbolt.org/z/T1Y5TfMa1

We can avoid this when `is_same_floating_point` is introduced.

harrism: But you have checks that they are all the same already, so this error message does not need to mention that. And there is a grammar error (types mismatches --> types mismatch)

cpp/include/cuspatial/distances/point_distance.hpp
isVoid: I looked at libgeos, geopandas and postgis and none give good hint about naming. So I say my feelings here. Personally I think `pairwise_point_distance` is sufficient to represent `point-to-point distance`. Because of the notion `pair`, it's unambiguous that two geometries are involved.

> Second, how do we order the geometries? Should the more complex geometry be first, or second?

This is what I felt too. But in more explicit way is to sort the geometries by dimensions.

> What if there is a directed distance metric 

One way is to use an extra enum argument `reflected`. The benefits are we don't have to break the ordering rule above while still providing the capability. The downside is that for same geometry (linestring-linestring), it's arguably redundant.

I also want to note the grouping of functions. The `pairwise_point_distance` should live under `distance/point_distance.hpp`. The name of the method "XXX_point_distance" should match the name of the file. In case in the future we need to implement a different parallel model e.g. `all_pairs_point_distance`, it should also live under this file. 

If you agree, I can make an issue about the names of distance method.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
isVoid: Grammar error aside... I think the changes to line 155 made it it's only checking if linestring 1 is of the same `value_type` to the output. It doesn't check if the linestring 1 and linestring 2 has the same `value_type`.  Consider if the user has the following input:

```
Cart2dItA: iterator to cartesian_2d<float> type
Cart2dItB: iterator to cartesian_2d<double> type
OutputIt: iterator to float type
```

It passes `is_floating_point` check, because all are floating point type. It passes first `is_same` check, because `Cart2dItA::value_type` and `OutputIt` are both `float`. It fails at the second `is_same` check, because `cartesian_2d<float> != cartesian_2d<double>`. But your suggestion would make this line to throw compile error
`Input types must be cuspatial::cartesian_2d`, which is confusing because the user does supply a `cartesian_2d` type.

The error would make sense for the following input:
```
struct MockType {
    using value_type = float;
};

Cart2dItA: iterator to MockType
Cart2dItB: iterator to MockType
OutputIt: iterator to float type
```

After `is_convertible_to` is introduced in #579 , we should use that here. Since `cartesian_2d<float>` shouldn't be convertible to `cartesian_2d<double>`, we should have the same type safety here.

cpp/include/cuspatial/distances/point_distance.hpp
```diff
@@ -0,0 +1,42 @@
+/*
```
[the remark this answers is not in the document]
isVoid: +1 for `distances` -> `distance`, I think currently haversine and hausdorff are already under the same category in the documentation, so +1 for moving them in the same directory too.

cpp/include/cuspatial/experimental/detail/linestring_distance.cuh
harrism: Good point. Still, I think the message can be simpler. 

```suggestion
                "Input types must be cuspatial::cartesian_2d with the same value type");
```

cpp/tests/CMakeLists.txt
```diff
@@ -105,10 +105,12 @@ ConfigureTest(SPATIAL_WINDOW_POINT_TEST
 ConfigureTest(HAVERSINE_TEST_EXP
     experimental/spatial/haversine_test.cu)
 
+ConfigureTest(POINT_DISTANCE_TEST_EXP
```
robertmaynard, line 108: This PR adds two test files 

1. `experimental/spatial/point_distance_test.cu`
2. `spatial/point_distance_test.cpp`

But we only have a test for `point_distance_test.cu`, making me think that `point_distance_test.cpp` needs to be deleted.

robertmaynard (changes requested):

harrism:
@isvoid will you be pushing this through in 22.08?

cpp/include/cuspatial/distances/point_distance.hpp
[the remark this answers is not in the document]
isVoid: Included these changes in the latest commit.

cpp/tests/CMakeLists.txt
isVoid: The `cpp` test was meant to test cudf column API and `cu` tests header only API. We need both so added both to CMakeLists.

cpp/include/cuspatial/distances/point_distance.hpp
isVoid: https://github.com/rapidsai/cuspatial/issues/607

robertmaynard (approved):
Approving CMake changes

cpp/benchmarks/hausdorff_benchmark.cpp
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#include <cuspatial/hausdorff.hpp>
+#include <cuspatial/distance/hausdorff.hpp>
```
harrism, line 17: So this is technically a breaking change. I will add the "breaking" tag.

cpp/include/cuspatial/distance/point_distance.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace cuspatial {
+
+/**
+ * @ingroup distance
+ * @brief Compute pairwise point-to-point Cartesian distance
+ *
+ * @param points1_x Column of x coordinates to the first point in each pair
+ * @param points1_y Column of y coordinates to the first point in each pair
+ * @param points2_x Column of x coordinates to the second point in each pair
+ * @param points2_y Column of y coordinates to the second point in each pair
```
harrism, line 30:
```suggestion
 * @param points1_x Column of x-coordinates of the first point in each pair
 * @param points1_y Column of y-coordinates of the first point in each pair
 * @param points2_x Column of x-coordinates of the second point in each pair
 * @param points2_y Column of y-coordinates of the second point in each pair
```

cpp/include/cuspatial/experimental/detail/point_distance.cuh
```diff
@@ -0,0 +1,64 @@
...
+  using T = typename std::iterator_traits<Cart2dItA>::value_type::value_type;
+
+  static_assert(
+    detail::is_floating_point<T,
+                              typename std::iterator_traits<Cart2dItB>::value_type::value_type,
+                              typename std::iterator_traits<OutputIt>::value_type>(),
+    "Inputs and output must be floating point types.");
+
+  static_assert(detail::is_same<T, typename std::iterator_traits<OutputIt>::value_type>(),
+                "Inputs and output must be the same types.");
```
harrism, line 46: If we merge #579 you can replace this with `is_same_floating_point`.

cpp/tests/experimental/spatial/point_distance_test.cu
```diff
@@ -0,0 +1,306 @@
...
+  using Cart2D = typename Cart2DVec::value_type;
+  using T      = typename Cart2D::value_type;
+  thrust::host_vector<Cart2D> h_point1(point1);
+  thrust::host_vector<Cart2D> h_point2(point2);
+  auto pair_iter =
+    thrust::make_zip_iterator(thrust::make_tuple(h_point1.begin(), h_point2.begin()));
+  auto result_iter = thrust::make_transform_iterator(pair_iter, [](auto p) {
+    auto p0 = thrust::get<0>(p);
+    auto p1 = thrust::get<1>(p);
+    return std::sqrt((p0.x - p1.x) * (p0.x - p1.x) + (p0.y - p1.y) * (p0.y - p1.y));
```
harrism, line 53: Can't you use the vec_2d operators here? 
```suggestion
    return std::sqrt(dot(p0-p1, p0-p1));
```

cpp/tests/spatial/point_distance_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+  auto x2 = fixed_width_column_wrapper<T>{};
+  auto y2 = fixed_width_column_wrapper<T>{};
+
+  auto expect = fixed_width_column_wrapper<T>{};
+
+  auto got = pairwise_point_distance(x1, y1, x2, y2);
+
+  CUDF_TEST_EXPECT_COLUMNS_EQUIVALENT(expect, *got);
+}
+
```
harrism, line 54: Test the runtime exceptions cases: size mismatch, type mismatch, nulls not supported.

harrism (changes requested):
Looks good. A couple of missing tests and other nits.

harrism:
Also need to update paths in Cython files to fix the build.

cpp/tests/spatial/point_distance_test.cpp
isVoid: Added.

isVoid:
Our cython file org fell out of sync with cpp file org. So in this PR I also moved them into `spatial` module.

harrism (approved):
Thanks @isVoid !

isVoid:
@gpucibot merge

PR #560, opened 2022-06-14
Relocate Utility Files

isVoid:
This PR moves several geometric utility functions to a separate header file so that they can be reused by other distance functions.

EDIT: per discussion below, this PR now also moves `vec_2d.cuh` to top level name space, `traits.hpp` and `geometry_utilities` are now under `cuspatial/detail/utility` and `geometry_utilities` are renamed to `linestring.cuh`.


cpp/include/cuspatial/utility/geometry_utility.cuh
```diff
@@ -0,0 +1,96 @@
+#pragma once
```
harrism, line 1: The path "utility/geometry_utility..." is a bit redundant. Also, I fear that "geometry utilities" could grow quite large. All of these are linestring / segment utilities, so perhaps a more specific filename is in order? 

Also these are all detail namespace so shouldn't `detail` be in the path? 

e.g. `include/cuspatial/detail/linestring_utilities.cuh` 

Looking at the utility folder, I think that `vec_2d.cuh` should be promoted to the top-level include directory (since it is not detail -- it's analogous to `constants.hpp`). Everything else in `utility` should move to `detail`. Thoughts?

isVoid: Great advice. In the other PR you mentioned we can move `device_atomics.cuh` to either within a `utility` folder or not. I think most methods here all fits under the utility category so should probably create a folder for that.

isVoid: Moved as noted except `device_atomics.cuh` will be handled in the other PR.

harrism (approved):

harrism:
pairwise linestring gtests now failing?

harrism:
In the interest of time, I think it's safe to merge this with one approval

harrism:
@gpucibot merge

PR #561, opened 2022-06-14
Combine `device_atomics`

### LINKED ISSUE #547: Combine device atomic functions
All four of these operators should probably be implemented as specializations of a single templated function. It'll need ~5 template parameters (the value type, the bit view type, the two conversion functions between value and view types, and the actual operation to perform e.g. std::max), which is kind of gross, but it'll significantly reduce duplication and the pain of adding new operators if you need other atomics in the future.

_Originally posted by @vyasr in https://github.com/rapidsai/cuspatial/pull/526#discussion_r884023580_

isVoid:
This PR makes device atomics inline and can be included in multiple header files. It also refactors the atomics to reduce duplication.

Closes #547


harrism (approved):
This is fantastic. I think it should move to `cuspatial/detail/device_atomics.cuh` or `cuspatial/detail/utility/device_atomics.cuh`.

isVoid:
rerun tests

isVoid:
> This is fantastic. I think it should move to `cuspatial/detail/device_atomics.cuh` or `cuspatial/detail/utility/device_atomics.cuh`.

Moved to `cuspatial/detail/utility/device_atomics.cuh`

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -14,6 +14,61 @@
...
  */
 
+#pragma once
+
+#include <algorithm>
+#include <type_traits>
+
+namespace {
+
+template <typename T>
```
vyasr, line 24: I would add comments that we need these functions defined here to resolve the otherwise ambiguous lookup of `std::[min|max]` in the template functions below.

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -14,6 +14,61 @@
...
+  return std::min(a, b);
+}
+
+template <typename T>
+const T& __device__ max(const T& a, const T& b)
+{
+  return std::max(a, b);
+}
+
+template <typename T,
```
vyasr, line 36: This needs docs, especially for all the template parameters.

isVoid: See updated docs.

cpp/include/cuspatial/detail/utility/device_atomics.cuh
isVoid: done

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -0,0 +1,191 @@
...
+#pragma once
+
+#include <algorithm>
+#include <type_traits>
+
+namespace {
+
+/**
+ * @internal
+ * This helper is required to avoid ambiguous lookup of std::min. There are
```
harrism, line 26: Should we really be using `std::min/std::max` at all? CUDA provides `min()` and `max()` overloads for float and double which map directly to hardware instructions.  Moreover, calling `std::min() / `std::max()` only  compiles successfully if you pass `--expt-relaxed-constexpr` to NVCC on the command line as this example shows: https://godbolt.org/z/8b85bx6e1

And this example (with that flag passed), shows that `std::min` results in  different code and more instructions (for the `float` case, which uses `FMNMX`).

https://godbolt.org/z/e6xcnTK99

In the `double` case, the code is different as well, but doesn't map directly to a `FMNMX` instruction: https://godbolt.org/z/afM7vP9od

I think we should stick to the NVIDIA math libraries in this case.  libcudac++ should probably add min and max that map to CUDA min and max in device code for consistency.

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * Reads the value from `addr`, performs `op(*addr, val)`, and stores the result
+ * to `addr` in one atomic transaction.
+ *
+ * @tparam T The type value to apply atomic operation to
+ * @tparam RepresentationType The unsigned integer type that has the same bit width as `T`
+ * @tparam OpType The type of the atomic operation
+ * @tparam ToRepFuncType The type of function to cast `T` to `RepresentationType`
+ * @tparam FromRepFuncType The type of function to cast `RepresentationType` to T
+ * @param addr The address where the atomic operation will perform on
```
harrism, line 58:
```suggestion
 * @param addr The address where the atomic operation will be performed
```

harrism (changes requested):
I think we should avoid `std::min/max` in device code.

cpp/include/cuspatial/detail/utility/device_atomics.cuh
vyasr: I agree, just calling `min`/`max` is probably what we should do here.

isVoid: Done.

harrism:
@vyasr if you are satisfied, we can merge this.

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -0,0 +1,186 @@
...
+ */
+template <typename T>
+const T __device__ min_(const T a, const T b)
+{
+  return min(a, b);
+}
+
+/**
+ * @internal
+ * A helper function to help lookup the correct overload of CUDA intrinsic
```
vyasr, line 37:
```suggestion
 * @brief A helper function to help lookup the correct overload of CUDA intrinsic
```

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -0,0 +1,186 @@
...
+ * @param val The right hand side value of the opeartion
+ * @param op The atomic operation to perform
+ * @param to_rep_func The function to cast `T` to `RepresentationType`, see notes below.
+ * @param from_rep_func The function to cast `RepresentationType` to `T`, see notes below.
+ * @return
+ *
+ * @note based on https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions
+ * @note `T`, `RepresentationType`, `to_rep_func` and `from_rep_func` are correlated. 32-bit floats
+ * corresponds to `unsigned int`, `__float_as_uint`, `__uint_as_float` respectively. 64-bit floats
+ * corresponds to `unsgined long long int`, `__double_as_longlong`, `__longlong_as_double`
```
vyasr, line 68:
```suggestion
 * corresponds to `unsigned long long int`, `__double_as_longlong`, `__longlong_as_double`
```

vyasr (approved):
Couple of minor suggestions for improvement, but assuming that they're implemented I approve!

cpp/include/cuspatial/detail/utility/device_atomics.cuh
```diff
@@ -0,0 +1,186 @@
...
+ * Atomically computes the min of the value stored in `addr` and `val` and stores it in `addr`,
+ * returning the previous value of `*addr`.
+ *
+ * @note based on https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions
+ *
+ * @param addr The address to atomically compare and update with the minimum.
+ * @param val The value to compare
+ * @return The old value stored in `addr`.
+ */
+__device__ inline double atomicMin(double* addr, double val)
```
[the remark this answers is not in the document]
isVoid: Done.

isVoid:
@gpucibot merge

PR #577, opened 2022-06-22
Add NVBench and linestring distance benchmark

isVoid:
This PR adds NVBench to cuspatial and introduces benchmarks for `pairwise_linestring_distance`.


isVoid:
Tagging @vyasr for cmake related changes, thanks!

cpp/benchmarks/CMakeLists.txt
```diff
@@ -52,6 +52,26 @@ function(ConfigureBench CMAKE_BENCH_NAME)
...
+  target_link_libraries(
+    ${CMAKE_BENCH_NAME} PRIVATE cuspatial_benchmark_common nvbench::main
+  )
+  install(
+    TARGETS ${CMAKE_BENCH_NAME}
+    COMPONENT testing
+    DESTINATION bin/benchmarks/libcuspatial
+    EXCLUDE_FROM_ALL
+  )
+endfunction(ConfigureNVBench)
```
vyasr, line 73: Modern CMake style is not to specify this since it's redundant and just another source of potential error. I would change `ConfigureBench` above too.
```suggestion
endfunction()
```

vyasr (approved):
One minor comment, otherwise the CMake LGTM.

cpp/benchmarks/pairwise_linestring_distance.cu
```diff
@@ -0,0 +1,170 @@
...
+
+#include <memory>
+
+namespace cuspatial {
+
+/**
+ * @brief Helper to generate linestrings used for benchmarks.
+ *
+ * The generator adopts a walking algorithm. The ith point is computed by
+ * walking (cos(i) * segment_length, sin(i) * segment_length) from the `i-1`
```
harrism, line 41: What is the goal for the shape of the linestrings? cos(i) and sin(i) are non-intuitive for me when `i` is an integer > 1.  Is this just a way to get a random value in [0,1]?

harrism (changes requested):
Suggest some simplification by dropping the cudf dependency.

cpp/benchmarks/pairwise_linestring_distance.cu
isVoid: Yup, it's more like a random walk algo here. The shape doesn't really matter but the locality of the linestring is more important, since the pair needs to separate far from each other to take the long code path to compute the distance.

cpp/benchmarks/pairwise_linestring_distance.cu
```diff
@@ -0,0 +1,170 @@
...
+{
+  cudf::size_type num_points = num_strings * (num_segments_per_string + 1);
+  rmm::device_uvector<cudf::size_type> offsets(num_points, stream);
+  thrust::transform(
+    rmm::exec_policy(stream),
+    thrust::make_counting_iterator(static_cast<cudf::size_type>(0)),
+    thrust::make_counting_iterator(static_cast<cudf::size_type>(num_points)),
+    offsets.begin(),
+    [num_segments_per_string] __device__(auto i) { return i * num_segments_per_string; });
+  auto points_x = cudf::make_fixed_width_column(
```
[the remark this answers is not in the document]
isVoid: Hmm, now that you mention, you are right about that. The current code is adapted from an earlier version that uses the cudf column API. With header only API in place, we can bypass using columns in all.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -52,6 +52,26 @@ function(ConfigureBench CMAKE_BENCH_NAME)
...
+  add_executable(${CMAKE_BENCH_NAME} ${ARGN})
+  set_target_properties(
+    ${CMAKE_BENCH_NAME}
+    PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/benchmarks>"
+               INSTALL_RPATH "\$ORIGIN/../../../lib"
+  )
+  target_link_libraries(
+    ${CMAKE_BENCH_NAME} PRIVATE cuspatial_benchmark_common nvbench::main
+  )
+  install(
```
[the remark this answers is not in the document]
vyasr: They both install them, the install command in `ConfigureBench` just doesn't show up in the Github diff unless you expand. At least, I assume that's why you didn't see it, because the same thing happened to me when I reviewed :slightly_smiling_face: 

As for whether or not we _should_ install them, I think this logic is copied from [the corresponding cudf file](https://github.com/rapidsai/cudf/blob/branch-22.08/cpp/benchmarks/CMakeLists.txt#L80). I agree that the installation probably isn't necessary, though.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ target_include_directories(cuspatial_benchmark_common
 function(ConfigureBench CMAKE_BENCH_NAME)
     add_executable(${CMAKE_BENCH_NAME} ${ARGN})
     set_target_properties(${CMAKE_BENCH_NAME}
-        PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/gbenchmarks>"
+        PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/benchmarks>"
```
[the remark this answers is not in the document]
isVoid: I don't think it's necessary to keep them in separate directories. Considering the naming is very similar (with a single character difference), it may be easily overlooked. Because cuspatial only has Hausdorff benchmark at the moment. I'm a bit wary that it might get neglected as we start populating the `benchmarks` folder with other NVBench. The concern also goes away once we rewrite Hausdorff benchmark with NVBench and removes gbenchmark dependency.

cpp/benchmarks/pairwise_linestring_distance.cu
harrism: Can you explain this in the comment?

thomcom (approved):
Great

cpp/benchmarks/pairwise_linestring_distance.cu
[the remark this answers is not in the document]
isVoid: Well actually, since there's an `exclusive_scan` call I'm not sure if this is possible. My knowledge of thrust iterators tells me we can't make an `exclusive_scan` iterator.

cpp/benchmarks/CMakeLists.txt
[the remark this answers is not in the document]
isVoid: Updated the component name to be `benchmark`.

cpp/benchmarks/pairwise_linestring_distance.cu
```diff
@@ -0,0 +1,170 @@
...
+                    thrust::counting_iterator(static_cast<cudf::size_type>(num_points)),
+                    points_y->mutable_view().begin<T>(),
+                    [] __device__(auto i) { return sin(i); });
+  thrust::exclusive_scan(
+    rmm::exec_policy(stream),
+    points_y->view().begin<T>(),
+    points_y->view().end<T>(),
+    points_y->mutable_view().begin<T>(),
+    init_xy,
+    [segment_length] __device__(T prev, T rad) { return prev + segment_length * rad; });
```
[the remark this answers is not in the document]
isVoid: No longer necessary because of the rewrite in response to https://github.com/rapidsai/cuspatial/pull/577#discussion_r904437609

cpp/benchmarks/pairwise_linestring_distance.cu
```diff
@@ -0,0 +1,145 @@
...
+#include <cuspatial/experimental/type_utils.hpp>
+
+#include <rmm/device_vector.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <thrust/iterator/counting_iterator.h>
+
+#include <memory>
+
+namespace cuspatial {
```
harrism, line 31: Not sure why these functions are being added to the cuSpatial namespace?

harrism (approved):
I approve. Just one question.

cpp/benchmarks/pairwise_linestring_distance.cu
isVoid: Good catch. The benchmarks shouldn't be counted in the library namespace because they are considered external code that treats the target API as a black box. I think it would be helpful to mention this in https://github.com/rapidsai/cudf/blob/branch-22.08/cpp/docs/BENCHMARKING.md

isVoid: Updated.

isVoid:
rerun tests

isVoid:
rerun tests

isVoid:
@gpucibot merge

cpp/benchmarks/pairwise_linestring_distance.cu
harrism: Did you open an issue for adding a namespace mention to BENCHMARKING.md?

PR #579, opened 2022-06-23
Header-only refactoring of `points_in_spatial_window`

### LINKED ISSUE #564: [FEA] Refactor `cuspatial::points_in_spatial_window()` to header-only API
Create header-only APIs for the following and refactor existing APIs / tests on top of the header-only API:

* cuspatial::points_in_spatial_window()

See https://github.com/rapidsai/cuspatial/blob/branch-22.08/cpp/include/cuspatial/spatial_window.hpp

harrism:
Fixes #564 

A couple of notes:
1. For clarity and generality, this name for the header-only version of the function is now `copy_points_in_range()` rather than `points_in_spatial_window()`. This is the reflect that it is a range filter, and we may want to enable constraining the dimensions of the range in the future.
2. I have renamed the parameters for the range delimiters and clarified their documentation. This API is not limited to Cartesian coordinates and the documents make that clear by specifying that the input points and the range limits are specified using points in the same coordinate system. It also uses "quadrilateral" rather than "rectangle", since in lon/lat (spherical) coordinates the range is not a rectangle, but it is a quadrilateral.
3. Note that the change to the iterator-based API means that the user is responsible for allocating the output (as they are with STL and Thrust). Therefore, they need a way to know how much to allocate. In addition to the header-only `copy_points_in_range()`, this PR also adds a new function, `cuspatial::count_points_in_range()`. 

Previously, the cuDF-based API used `cudf::copy_if`, which does the equivalent of both of the above functions and returns a properly sized pair of columns. The new column-based implementation calls `count_points_in_range`, then allocates the result columns, and then calls `cuspatial::copy_points_in_range()`.


cpp/include/cuspatial/experimental/detail/spatial_window.cuh
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
```
isVoid, line 2:
```suggestion
 * Copyright (c) 2022, NVIDIA CORPORATION.
```

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+  InputIt points_first,
+  InputIt points_last,
+  rmm::cuda_stream_view stream = rmm::cuda_stream_default);
+
+/**
+ * @brief Find all points (x,y) that fall within a rectangular query window.
+ *
+ * @ingroup spatial_relationship
+ *
+ * A point (x, y) is in the window if `x > window_min_x && x < window_min_y && y > window_min_y &&
```
isVoid, line 66:
```suggestion
 * A point (x, y) is in the window if `x > window_min_x && x < window_max_x && y > window_min_y &&
```

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+#include <thrust/iterator/iterator_traits.h>
+
+namespace cuspatial {
+
+/**
+ * @brief Count of points (x,y) that fall within a rectangular query window.
+ *
+ * @ingroup spatial_relationship
+ *
+ * A point (x, y) is in the window if `x > window_min_x && x < window_min_y && y > window_min_y &&
```
isVoid, line 32:
```suggestion
 * A point (x, y) is in the window if `x > window_min_x && x < window_max_x && y > window_min_y &&
```

cpp/include/cuspatial/experimental/detail/spatial_window.cuh
```diff
@@ -0,0 +1,103 @@
...
+template <class InputIt, class OutputIt, class T>
+OutputIt points_in_spatial_window(vec_2d<T> window_min,
+                                  vec_2d<T> window_max,
+                                  InputIt points_first,
+                                  InputIt points_last,
+                                  OutputIt output_points_first,
+                                  rmm::cuda_stream_view stream)
+{
+  using Point       = typename std::iterator_traits<InputIt>::value_type;
+  using OutputPoint = typename std::iterator_traits<InputIt>::value_type;
```
isVoid, line 86:
```suggestion
  using OutputPoint = typename std::iterator_traits<OutputIt>::value_type;
```

cpp/include/cuspatial/experimental/detail/spatial_window.cuh
```diff
@@ -0,0 +1,103 @@
...
+
+  static_assert(detail::is_convertible_to<cuspatial::vec_2d<T>, Point, OutputPoint>(),
+                "Input and Output points must be convertible to cuspatial::vec_2d");
+
+  static_assert(detail::is_same_floating_point<T,
+                                               typename Point::value_type,
+                                               typename OutputPoint::value_type>(),
+                "Inputs and output must have the same value type.");
```
isVoid, line 69: This function doesn't have an output iterator. We can drop these checks.

isVoid (commented):
Only several minor comments. At first I find it striking to see the refactor breaks the original function into two kernel calls. Later I realized `cudf::copy_if` does two kernel launches internally (one pass to count the output size and another pass to copy the result).

harrism:
> Only several minor comments. At first I find it striking to see the refactor breaks the original function into two kernel calls. Later I realized `cudf::copy_if` does two kernel launches internally (one pass to count the output size and another pass to copy the result).

Yes that's what I explained here: https://github.com/rapidsai/cuspatial/pull/579#issue-1281713165

cpp/include/cuspatial/experimental/detail/spatial_window.cuh
harrism: Fixed

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: Fixed

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: Fixed

cpp/include/cuspatial/experimental/detail/spatial_window.cuh
harrism: Fixed

cpp/include/cuspatial/experimental/detail/spatial_window.cuh
harrism: Thanks, fixed.

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * @ingroup spatial_relationship
+ *
+ * A point (x, y) is in the window if `x > window_min_x && x < window_max_x && y > window_min_y &&
+ * y < window_max_y`.
+ *
+ * Swaps `window_min.x` and `window_max.x` if `window_min.x > window_max.x`.
+ * Swaps `window_min.y` and `window_max.y` if `window_min.y > window_max.y`.
+ *
+ * @param[in] window_min lower-left (x, y) coordinate of the query window
+ * @param[in] window_max upper-right (x, y) coordinate of the query window
```
isVoid, line 39: There isn't a uniform coordinate system in cuspatial so calling it lower-left and upper-right could be confusing. I think it's best to refer to them as "one of the opposite corners of the query window". Analogous to opencv: https://docs.opencv.org/4.x/d6/d6e/group__imgproc__draw.html#ga07d2f74cadcf8e305e810ce8eed13bc9

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @pre All iterators must have the same `value_type`, with the same underlying floating-point
+ * coordinate type (e.g. `cuspatial::vec_2d<float>`).
+ *
+ * @returns The number of input points that fall within the specified window.
+ */
+template <class InputIt, class T>
+typename thrust::iterator_traits<InputIt>::difference_type count_points_in_spatial_window(
+  vec_2d<T> window_min,
+  vec_2d<T> window_max,
```
isVoid, line 56: If we are swapping the window corners, makes me wonder if we want to name them as `min`, `max` since it really doesn't matter?

opencv calls it `pt1` and `pt2` for example: https://docs.opencv.org/4.x/d6/d6e/group__imgproc__draw.html#ga07d2f74cadcf8e305e810ce8eed13bc9.

isVoid (approved):
Just some doc related issue.

cpp/include/cuspatial/experimental/spatial_window.cuh
thomcom: Should we specify that these are cartesian coordinates, in this context?

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * Swaps `window_min.y` and `window_max.y` if `window_min.y > window_max.y`.
+ *
+ * @param[in] window_min lower-left (x, y) coordinate of the query window
+ * @param[in] window_max upper-right (x, y) coordinate of the query window
+ * @param[in] points_first beginning of range of (x, y) coordinates of points to be queried
+ * @param[in] points_last end of range of (x, y) coordinates of points to be queried
+ * @param[in] stream: The CUDA stream on which to perform computations
+ *
+ * @tparam InputIt Iterator to input points. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
```
thomcom, line 45:
```suggestion
 * [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) and be device-accessible.
```

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+template <class InputIt, class T>
+typename thrust::iterator_traits<InputIt>::difference_type count_points_in_spatial_window(
+  vec_2d<T> window_min,
+  vec_2d<T> window_max,
+  InputIt points_first,
+  InputIt points_last,
+  rmm::cuda_stream_view stream = rmm::cuda_stream_default);
+
+/**
+ * @brief Find all points (x,y) that fall within a rectangular query window.
```
thomcom, line 62: Specify cartesian coordinates, I think.

thomcom: Down the road this, along with pip, would be really effective if they took a coordinate system pair as arguments.

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * @param[in] window_min lower-left (x, y) coordinate of the query window
+ * @param[in] window_max upper-right (x, y) coordinate of the query window
+ * @param[in] points_first beginning of range of (x, y) coordinates of points to be queried
+ * @param[in] points_last end of range of (x, y) coordinates of points to be queried
+ * @param[out] output_points_first beginning of output range of (x, y) coordinates within the
+ * query window
+ * @param[in]  stream: The CUDA stream on which to perform computations and allocate memory.
+ *
+ * @tparam InputIt Iterator to input points. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
```
thomcom, line 81:
```suggestion
 * [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) and be device-accessible.
```

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * @tparam OutputIt Output iterator. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible and mutable.
```
thomcom, line 83:
```suggestion
 * LegacyRandomAccessIterator] and be device-accessible and mutable.
```

Like a definition I think it only needs one link per document block.

cpp/include/cuspatial/experimental/spatial_window.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * @tparam T The underlying coordinate type. Must be a floating-point type.
+ *
+ * @pre The range `[points_first, points_last)` may equal the range `[output_points_first,
+ * output_points_first + std::distance(points_first, points_last)), but the ranges may not
+ * partially overlap.
```
thomcom, line 88: I did not see where the partial overlap prohibition is being enforced?

cpp/tests/experimental/spatial/spatial_window_test.cu
```diff
@@ -0,0 +1,121 @@
...
+    Vec{5.5, 5.5}, Vec{1.5, 1.5}, points.begin(), points.end());
+
+  EXPECT_EQ(result_size, expected_points.size());
+
+  auto result_points = rmm::device_vector<Vec>(result_size);
+  cuspatial::points_in_spatial_window(
+    Vec{5.5, 5.5}, Vec{1.5, 1.5}, points.begin(), points.end(), result_points.begin());
+
+  EXPECT_EQ(expected_points, result_points);
+}
```
thomcom, line 121: Suggest two more tests: one for "all points in window" which is trivial. The other is less trivial - do we need to test edge cases, like points along the edge? What level of precision will it work/fail? That's a great thing to cover in test cases.

thomcom (changes requested):

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: I like this suggestion. So the names `min` and `max` are not appropriate either...

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: it's an HTML link, so I think it needs to be repeated wherever you want the link.

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: It can't be enforced. It's just Undefined Behavior if you try to partially overlap the ranges (because you would have a race condition). This is a common precondition in Thrust, for example.

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: Missing link definition added

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: Missing link reference added.

cpp/include/cuspatial/experimental/spatial_window.cuh
thomcom: :+1:

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: Actually I don't think we have to limit this to cartesian coordinates. This is effectively a 2D range query. There's no reason you can't use lonlat points and a 2D lonlat range.

harrism: (In the future we may want to replace this API with a generic ND range query API...)

cpp/tests/experimental/spatial/spatial_window_test.cu
harrism: Good ideas. Added new tests and cleaned up the tests (and factored out repetitive code).

cpp/include/cuspatial/experimental/spatial_window.cuh
thomcom: I don't think that's strictly true - lon/lat is not cartesian. There exist points along a line of longitude that fall inside of the spherical "rectangle" defined by a lower left and upper right pair of coordinates that do not fall inside of the cartesian rectangle that is defined by those same coordinates.

cpp/tests/experimental/spatial/spatial_window_test.cu
thomcom: Did you create a new file for the new tests? I see one new test, `Empty`, and that you renamed the file from `.cpp` to `.cu`, but I don't see an all points or edge test.

thomcom (changes requested):

harrism:
> Did you create a new file for the new tests? I see one new test, Empty, and that you renamed the file from .cpp to .cu, but I don't see an all points or edge test.

@thomcom I think you went looking in an old commit for new changes. :) Click "Files Changed" at the top or see here: https://github.com/rapidsai/cuspatial/pull/579/commits/464cbcc5b8cf1c5b2835dfce00507524a30cede9

cpp/include/cuspatial/experimental/spatial_window.cuh
harrism: The API doesn't specify that the points are in any projection -- they can be Cartesian, Lon/Lat, or anything else, as long as the query window and the points are coordinates in the same dimensions.  This is a good reason to replace this with an API called "points_in_range" or something like that.

harrism: I've renamed the APIs and clarified the docs. See commit #[88e7ee1](https://github.com/rapidsai/cuspatial/pull/579/commits/88e7ee18c9db8d9662a237f07e3e0a55bc1d8198) and after.

cpp/tests/experimental/spatial/spatial_window_test.cu
thomcom: I see them, looking good!

thomcom (approved):

harrism:
@gpucibot merge

PR #581, opened 2022-06-29
Fix GeoArrow format for MultiLine and MultiPolygon root object offsets.

### LINKED ISSUE #575: [BUG] cuSpatial does not correctly implement GeoArrow format for MultiLineString and MultiPolygon
**Describe the bug**
My implementation of the named datatypes uses a poorly imagined "brackets" data structure to identify where the Multi features are located within the offsets arrays for LineString and Polygon coordinates. Instead of using brackets, to match the [GeoArrow](https://github.com/geopandas/geo-arrow-spec/issues/4) format,  the offsets for LineStrings and Polygons should simply be 1, and the offset for Multi's should be the number of that feature that the Multi is made from in sequence.

**Steps/Code to reproduce bug**

The issue is documented in precise detail here: https://notebooksharing.space/view/517f3172b12354804179f248247ab5ffd6573214e9f9810d13494533f1aefd8a#displayOptions=

**Plan**
As part of fixing the bug, I'll convert `cuspatial` to use `pyarrow` for all of the Shapely/GeoPandas serialization. Presently I am doing this serialization manually, which requires many lines of code and is hard for other developers to understand. Additionally, that means that all indexing, host/device copies, and deserialization are manual, too. Fixing this bug will have four steps:

- [x] Drop manual input serialization of GeoSeries objects and use `pyarrow` native types for all serialization. 
#582
#583
- [x] Replace all the manual indexing into GeoArrow objects with simple `pyarrow` based indexing.
- [x] Add proper `getitem` for `UnionArray` in pyarrow.
- [x] Replace any manual device/host data movement with `pyarrow` native code.
- [x] Create all Shapely objects and GeoSeries objects using `pyarrow` buffers.
- [x] Refactor parent class of GeoColumn to be ListColumn instead of NumericalColumn

As each step begins I'll create an issue and a PR for it and document them here.

thomcom:
This PR closes #575

### DIFF
--- a/python/cuspatial/cuspatial/geometry/geoarrowbuffers.py
+++ b/python/cuspatial/cuspatial/geometry/geoarrowbuffers.py
@@ -603,19 +603,4 @@ class PolygonArray(OffsetArray):
         return result
 
     def __len__(self):
-        if len(self._mpolys) > 0:
-            mlength = (
-                self._mpolys.values[
-                    np.arange(
-                        1, len(self._mpolys), 2, like=self._mpolys.values
-                    )
-                ]
-                - self._mpolys.values[
-                    np.arange(
-                        0, len(self._mpolys), 2, like=self._mpolys.values
-                    )
-                ]
-            ).sum() - (len(self._mpolys) // 2)
-        else:
-            mlength = 0
-        return (len(self.polys) - 1) - int(mlength)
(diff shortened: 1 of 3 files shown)


harrism:
@thomcom is this superceded by #585 ? Or in addition to it? If the latter, should it be retargeted to 22.10?

PR #583, opened 2022-06-30
Create `pygeoarrow` and use it for cuSpatial feature storage and i/o

### LINKED ISSUE #582: [FEA] Use `pyarrow` to read and serialize geoPandas geometry
In #575, I'm correcting an error with my GeoArrow implementation. The specification in `pyarrow` for polygons and linestrings is
```py
polygon_type = pa.list_(
    pa.field("polygons", pa.list_(
        pa.field("rings", pa.list_(
            pa.field("vertices", pa.list_(
                pa.field("xy", pa.float64(), nullable=False), 2),
            nullable=False)), nullable=False))))

linestring_type = pa.list_(
    pa.field("lines", pa.list_(
        pa.field("offsets", pa.list_(
            pa.field("xy", pa.float64(), nullable=False), 2),
        nullable=False)), nullable=False))
```

Because I implemented `arrow` from scratch in https://github.com/rapidsai/cuspatial/pull/300, and in a spasm of reasoning, I forgot to use the prefix-scan method for indexing the above types.

The first step in correcting it, given where I am with `pyarrow` now, is to change the input of our GeoArrow types to use arrow directly.

`cuspatial/io/geopandas_adapter.py` implements the arrow spec when reading from `GeoSeries` objects in a `GeoPandas` dataframe. 

`_load_geometry_offsets` computes the sizes and prefix sum/scan buffers for each of the four `arrow` lists and allocates memory for them.

`_read_geometries` iterates over the geometries again, copying individual coordinates and offsets into the correct position in each buffer.

Instead, specify the `GeoArrow` format types as per above in `geoarrow.py`, then use those formats and densely packed tuple buffers from `GeoSeries[i].__geo_interface__['coordinates']` to create the `geoarrowbuffers.py` objects.

This will break unit tests for LineStrings and MultiLineStrings, which have been implemented incorrectly, leading to the original Issue #575 which will have another PR to fix them.

thomcom:
This closes #582.
This PR removes the input based on an iterative Shapely reader and `cudf` buffers. Now the input is stored directly in a pyarrow `UnionArray`. The next PR will remove most of the interior functionality that is not necessary.

### DIFF
--- a/python/cuspatial/cuspatial/geometry/geoarrowbuffers.py
+++ b/python/cuspatial/cuspatial/geometry/geoarrowbuffers.py
@@ -1,9 +1,10 @@
-# Copyright (c) 2021 NVIDIA CORPORATION
+# Copyright (c) 2021-2022 NVIDIA CORPORATION
 
 from typing import TypeVar, Union
 
 import numpy as np
 import pandas as pd
+import pyarrow as pa
 
 import cudf
 
@@ -83,6 +84,7 @@ class GeoArrowBuffers:
         self._multipoints = None
         self._lines = None
         self._polygons = None
+
         if isinstance(data, dict):
             if data.get("points_xy") is not None:
                 self._points = CoordinateArray(
@@ -120,6 +122,7 @@ class GeoArrowBuffers:
                     data.get("mpolygons"),
                     data_locale=data_locale,
                 )
+
         elif isinstance(data, GeoArrowBuffers):
             if data.points is not None:
                 self._points = CoordinateArray(
@@ -149,10 +152,34 @@ class GeoArrowBuffers:
                     data.polygons.z,
                     data_locale=data_locale,
                 )
-        else:
-            raise TypeError(
-                f"Invalid type passed to GeoArrowBuffers ctor {type(data)}"
+
+        elif isinstance(data, pa.lib.UnionArray):
+            self._points = CoordinateArray(
+                data.field(0).values, [], data_locale=data_locale
             )
+            self._multipoints = MultiPointArray(
+                data.field(1).values.values,
+                data.field(1).offsets,
+                [],
+                data_locale=data_locale,
+            )
+            self._lines = LineArray(
+                data.field(2).values.values.values,
+                data.field(2).values.offsets,
+                data.field(2).offsets,
+                [],
+                data_locale=data_locale,
+            )
+            self._polygons = PolygonArray(
+                data.field(3).values.values.values.values,
+                data.field(3).values.values.offsets,
+                data.field(3).values.offsets,
+                data.field(3).offsets,
+                [],
+                data_locale=data_locale,
+            )
+        else:
+            raise TypeError(f"Invalid type passed to GeoArrowBuffers ctor {type(data)}")
 
     @property
     def points(self):
@@ -205,12 +232,8 @@ class GeoArrowBuffers:
         multipoints_length = (
             len(self._multipoints) if self.multipoints is not None else 0
         )
-        polygons_length = (
-            len(self._polygons) if self.polygons is not None else 0
-        )
-        return (
-            points_length + lines_length + multipoints_length + polygons_length
-        )
+        polygons_length = len(self._polygons) if self.polygons is not None else 0
+        return points_length + lines_length + multipoints_length + polygons_length
 
     def copy(self, deep=True):
         """
@@ -257,6 +280,10 @@ class GeoArrowBuffers:
 
 
 class CoordinateArray:
+    _data_location = None
+    _xy = None
+    _z = None
+
     def __init__(self, xy, z=None, data_locale=cudf):
         """
         A GeoArrow column of points. The CoordinateArray stores all of the
@@ -279,17 +306,18 @@ class CoordinateArray:
 
     @data_location.setter
     def data_location(self, data_location):
-        if data_location not in (cudf, pd):
+        if data_location not in (cudf, pd, pa):
             raise NotImplementedError(
-                "only cudf and pandas CoordinateArrays "
-                "are supported at this time"
+                "only cudf, pandas, and pa CoordinateArrays " "are supported at this time"
             )
         else:
             self._data_location = data_location
 
     def _serialize(self, data):
         try:
-            if self._data_location == pd:
+            if self._data_location == pa:
+                return data
+            elif self._data_location == pd:
                 if isinstance(data, cudf.Series):
                     return data.to_pandas()
                 else:
@@ -316,7 +344,9 @@ class CoordinateArray:
     def xy(self, xy):
         if (len(xy) % 2) != 0:
             raise ValueError("xy must have even length")
-        self._xy = self._serialize(xy)
+        temp = self._serialize(xy)
+        print(temp)
+        self._xy = temp
 
     @property
     def z(self):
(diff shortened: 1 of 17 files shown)


thomcom:
The next PR will have tons more - than + I think.

thomcom:
This PR is messy because the old wrapper and `pyarrow` are co-existing.

python/cuspatial/cuspatial/geometry/geocolumn.py
```diff
@@ -37,54 +37,51 @@ def __init__(self, meta: Union[GeoArrowBuffers, dict]):
...
-            else:
-                self.input_types.extend(repeat("poly", len(buffers.polygons)))
-                self.input_lengths.extend(repeat(1, len(buffers.polygons)))
+                if poly_len > 1:
+                    self.input_types.extend([5])
+                    self.input_lengths.extend([poly_len])
+                else:
+                    self.input_types.extend([4])
+                    self.input_lengths.extend([1])
+        self.input_types = pa.array(self.input_types).cast(pa.int8())
```
isVoid, line 69:
```suggestion
        self.input_types = pa.array(self.input_types, type=pa.int8())
```

python/cuspatial/cuspatial/geometry/geocolumn.py
```diff
@@ -37,54 +37,51 @@ def __init__(self, meta: Union[GeoArrowBuffers, dict]):
...
+        self.input_lengths = pa.array(self.input_lengths).cast(pa.int32())
 
     def copy(self):
         return type(self)(
```
isVoid, line 73:
```suggestion
        return self.__class__(
```

python/cuspatial/cuspatial/tests/test_from_geopandas.py
```diff
@@ -36,38 +36,42 @@ def test_dataframe_column_access(gs):
...
+    assert cugs.polygons.rings.sum().as_py() == 327
 
 
 def test_from_geopandas_point():
     gs = gpd.GeoSeries(Point(1.0, 2.0))
     cugs = cuspatial.from_geopandas(gs)
-    cudf.testing.assert_series_equal(cugs.points.xy, cudf.Series([1.0, 2.0]))
+    cudf.testing._utils.assert_eq(
+        cugs.points.xy, cudf.Series([1.0, 2.0], dtype="float64").to_arrow()
+    )
```
isVoid, line 52: Since methods defined in `_utils` is internal, using them is discouraged. Can we make it work by
```
cudf.testing.assert_series_equal(cugs.points.xy.to_pandas(), cudf.Series([1.0, 2.0], dtype="float64").to_pandas())
```

Or simply resort to `pd.testing.assert_series_equal`

isVoid (commented):
Since this is my first review of geoarrow PRs, so bear with me if I have missed some background knowledge. I have a high level question for `GeoColumn`. Why is `GeoColumn` a `NumericalColumn`? It seems to me no single existing cudf column type is sufficient to represent what `GeoColumn` expresses. It seems a bit awkward to piggy back on existing methods defined for `NumericalColumn`.

However, each "pure geometry" arrays can be represented as a `ListColumn` in cuDF. My intuition is that we can use `ListColumn` for each of the pure geometry arrays. The indexing logic introduced here for `GeoColumn` can continue serve as the top level indexing.

python/cuspatial/cuspatial/geometry/pygeoarrow.py
```diff
@@ -0,0 +1,223 @@
...
+                                pa.field("xy", pa.float64(), nullable=False), 2
+                            ),
+                            nullable=False,
+                        )
+                    ),
+                    nullable=False,
+                )
+            ),
+        )
+    )
```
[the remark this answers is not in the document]
thomcom: Sure, that's cleaner and `black` sure formats it more nicely. :D

python/cuspatial/cuspatial/geometry/geocolumn.py
```diff
@@ -229,14 +233,16 @@ def __getitem__(self, index):
...
-            "l": LineStringShapelySerializer,
-            "ml": MultiLineStringShapelySerializer,
-            "poly": PolygonShapelySerializer,
-            "mpoly": MultiPolygonShapelySerializer,
+            0: PointShapelySerializer,
+            1: MultiPointShapelySerializer,
+            2: LineStringShapelySerializer,
+            3: MultiLineStringShapelySerializer,
+            4: PolygonShapelySerializer,
+            5: MultiPolygonShapelySerializer,
```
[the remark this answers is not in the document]
thomcom: Thanks, I was planning on doing this but by python enum was feeling rusty. Are you ok if I do this in the next PR and we merge this one presently?

python/cuspatial/cuspatial/geometry/geoarrowbuffers.py
```diff
@@ -257,6 +292,10 @@ def __repr__(self):
 
 
 class CoordinateArray:
+    _data_location = None
+    _xy = None
+    _z = None
+
```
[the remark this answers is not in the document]
thomcom: It appears they are not. I was having a problem with a `self.` accessor complaining about `_xy` not being initialized before assignment, but it went away.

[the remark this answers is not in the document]
thomcom: Deleted.

python/cuspatial/cuspatial/geometry/geoarrowbuffers.py
```diff
@@ -385,6 +427,8 @@ def __init__(self, xy, z=None, data_locale=cudf):
 
 
 class OffsetArray(CoordinateArray):
+    _offsets = None
+
```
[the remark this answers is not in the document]
thomcom: Done.

python/cuspatial/cuspatial/tests/test_from_geopandas.py
thomcom: Done. I made the pyarrow array a Series.

isVoid (approved):
Awesome, as requested I only went pretty broad-strokes this time and I see no red flags. I see there are internal calls to `assert_eq`s that can hopefully be replaced in future PRs.

And just to be clear, `groupby` is only promised at a `indexed_frame` level in cudf and is very much a separate implementation unrelated to column. see:
https://github.com/rapidsai/cudf/blob/branch-22.08/python/cudf/cudf/core/groupby/groupby.py

It totally makes sense to implement geoseries and geodataframe inheriting from series and dataframe to make use of the compute APIs for non-geometric columns. But `geocolumn` is too eccentric that I think it's best to implement an underlying `UnionColumn` type in cuspatial and `geocolumn` can inherit from it.

thomcom:
@gpucibot merge

python/cuspatial/cuspatial/geometry/pygeoarrow.py
```diff
@@ -0,0 +1,189 @@
...
+                raise TypeError(type(geom))
+            type_buffer.append(
+                {
+                    Point: 0,
+                    MultiPoint: 1,
+                    LineString: 2,
+                    MultiLineString: 3,
+                    Polygon: 4,
+                    MultiPolygon: 5,
+                }[type(geom)]
```
trxcllnt, line 85: @thomcom Sorry I'm late to the review game here. Isn't it slow to create a new dict and do this lookup for every value in the `GeoSeries`? Why not append the typecode in the conditionals above (like we do with the offsets), or at least define this dict at the top once and read from it?

thomcom: Thanks for the comment, I'll clean that up on the next go around, which is coming up today.

thomcom: I just moved it into the append step like you suggested. 3 appends seems like a lot. :parrot:

PR #590, opened 2022-07-13
Add missing rmm includes

harrism:
A recent change to libcudf removed a forward declaration and/or RMM header includes, and cuSpatial was errantly relying on these to get the definitions of `device_memory_resource` and `get_current_device_resource()`. 

This PR fixes that by adding missing includes of `rmm/mr/device/per_device_resource.hpp` in several files. It also generally cleans up include ordering and spacing, and adds a few missing copyright headers.

### DIFF
--- a/cpp/include/cuspatial/error.hpp
+++ b/cpp/include/cuspatial/error.hpp
@@ -16,8 +16,9 @@
 
 #pragma once
 
-#include <cuda.h>
+//#include <cuda.h>
 #include <cuda_runtime_api.h>
+
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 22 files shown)


cpp/include/cuspatial/error.hpp
```diff
@@ -16,8 +16,9 @@
 
 #pragma once
 
-#include <cuda.h>
+//#include <cuda.h>
```
isVoid, line 19: We should remove this because it's redundant of `cuda_runtime_api.h`?

isVoid (approved):
One minor comments. Can merge once CI passes.

cpp/include/cuspatial/error.hpp
harrism: Oops, I was testing commenting it out and forgot to remove it.  Yes, will remove.

harrism:
@gpucibot merge

PR #595, opened 2022-07-19
Add benchmark for points_in_spatial_window

harrism:
This currently calls the cuDF-based C++ API for `points_in_spatial_window`, in order to capture "before refactoring" throughput of the API. I plan to also run the post-refactoring version, and then update the benchmark to run the refactored API in order to not depend on libcudf in this benchmark.

### DIFF
--- a/cpp/benchmarks/spatial_window.cu
+++ b/cpp/benchmarks/spatial_window.cu
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <benchmarks/fixture/rmm_pool_raii.hpp>
+#include <benchmarks/utility/random.cuh>
+
+#include <cuspatial/detail/iterator.hpp>
+#include <cuspatial/spatial_window.hpp>
+#include <cuspatial/vec_2d.hpp>
+
+#include <cudf_test/column_wrapper.hpp>
+
+#include <rmm/device_vector.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <nvbench/nvbench.cuh>
+
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/random/linear_congruential_engine.h>
+#include <thrust/random/normal_distribution.h>
+#include <thrust/random/uniform_int_distribution.h>
+
+#include <memory>
+
+using namespace cuspatial;
+
+/**
+ * @brief Helper to generate random points within a rectangular window
+ *
+ * @p begin and @p end must be iterators to device-accessible memory
+ *
+ * @tparam PointsIter The type of the iterator to the output points container
+ * @tparam T The floating point type for the coordinates
+ * @param begin The start of the range of points to generate
+ * @param end The end of the range of points to generate
+ *
+ * @param window_min the lower left window corner
+ * @param window_max the upper right window corner
+ *
+ */
+template <class PointsIter, typename T>
+void generate_points(PointsIter begin, PointsIter end, vec_2d<T> window_min, vec_2d<T> window_max)
+{
+  auto engine = deterministic_engine(std::distance(begin, end));
+  auto x_dist = make_uniform_dist(window_min.x, window_max.x);
+  auto y_dist = make_uniform_dist(window_min.y, window_max.y);
+  auto x_gen  = value_generator{window_min.x, window_max.x, engine, x_dist};
+  auto y_gen  = value_generator{window_min.y, window_max.y, engine, y_dist};
+  thrust::tabulate(rmm::exec_policy(), begin, end, [x_gen, y_gen] __device__(size_t n) mutable {
+    return vec_2d<T>{x_gen(n), y_gen(n)};
+  });
+}
+
+template <typename T>
+void points_in_spatial_window_benchmark(nvbench::state& state, nvbench::type_list<T>)
+{
+  // TODO: to be replaced by nvbench fixture once it's ready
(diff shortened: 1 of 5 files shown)


cpp/benchmarks/utility/random.cuh
```diff
@@ -0,0 +1,152 @@
...
+ * @brief Generates a normal distribution between zero and upper_bound.
+ */
+template <typename T>
+auto make_normal_dist(T lower_bound, T upper_bound)
+{
+  using realT    = integral_to_realType<T>;
+  T const mean   = lower_bound + (upper_bound - lower_bound) / 2;
+  T const stddev = (upper_bound - lower_bound) / 6;
+  return thrust::random::normal_distribution<realT>(mean, stddev);
+}
```
thomcom, line 69: This seems like overkill, to respecify a normal distribution in terms of upper and lower bounds instead of mean and stddev. Also, I think that with this stddev you're going to generate ~2% of your points outside of the bounds?

thomcom (approved):

cpp/benchmarks/utility/random.cuh
```diff
@@ -0,0 +1,152 @@
...
+ * @brief Real Type that has atleast number of bits of integral type in its mantissa.
+ *  number of bits of integrals < 23 bits of mantissa in float
+ * to allow full range of integer bits to be generated.
+ * @tparam T integral type
+ */
+template <typename T>
+using integral_to_realType =
+  std::conditional_t<std::is_floating_point_v<T>,
+                     T,
+                     std::conditional_t<sizeof(T) * 8 <= 23, float, double>>;
```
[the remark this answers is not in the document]
harrism: The whole file is copied from libcudf.  This chooses a floating point type that is large enough to represent the range of the specified integer type. So for  example an int32_t is 4 bytes so it would need a double.

cpp/benchmarks/utility/random.cuh
harrism: This is a convenience to allow your 6 sigma to fit within a specified range. It's not 2%, it's like 0.27% (99.73% fall within 6 standard deviations of the mean). The 0.27% that fall outside will be handled by the `clamp`s in the calling function.  So you will get 0.27% of the points clustered at the lower and upper bounds. 

(Again, I didn't write this code, it's taken from libcudf. It was much less work to copy it than to reinvent it.)

thomcom: :+1:

cpp/benchmarks/utility/random.cuh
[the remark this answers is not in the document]
thomcom: Ok, the floating point needs to have the same level of precision that the int does. This seems like it is also really just a LUT - int types below 32 bit can be represented with a 32 bit float, int32 can be represented with a double, and int64 can't be represented fully by a float.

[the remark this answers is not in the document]
harrism: The key is that this is done at compile time. There are other ways to write the conditional, but I can't think of a simpler way -- e.g. I think building a LUT using C++ template metaprogramming would be much more complex.

cpp/benchmarks/spatial_window.cu
```diff
@@ -0,0 +1,114 @@
...
+
+  auto x_dist = make_uniform_dist(window_min.x, window_max.x);
+  auto y_dist = make_uniform_dist(window_min.y, window_max.y);
+
+  auto x_gen = value_generator{window_min.x, window_max.x, engine_x, x_dist};
+  auto y_gen = value_generator{window_min.y, window_max.y, engine_y, y_dist};
+
+  thrust::tabulate(rmm::exec_policy(), begin, end, [x_gen, y_gen] __device__(size_t n) mutable {
+    return vec_2d<T>{x_gen(n), y_gen(n)};
+  });
```
isVoid, line 66: Trying to learn something here: What's the difference between a `thrust::tabulate` and a `thrust::copy` + `make_counting_transform_iterator`?

cpp/benchmarks/spatial_window.cu
```diff
@@ -0,0 +1,114 @@
...
+  auto window_max = vec_2d<T>{100, 100};
+
+  auto range_min = vec_2d<T>{-200, -200};
+  auto range_max = vec_2d<T>{200, 200};
+
+  auto d_points = rmm::device_uvector<vec_2d<T>>(num_points, rmm::cuda_stream_default);
+  generate_points(d_points.begin(), d_points.end(), range_min, range_max);
+
+  auto d_x = rmm::device_uvector<T>(num_points, rmm::cuda_stream_default);
+  auto d_y = rmm::device_uvector<T>(num_points, rmm::cuda_stream_default);
```
isVoid, line 87: I think you can use `make_zipped_vec_2d_output_iterator` and pass the zipped iterator to `generate_points`?

cpp/benchmarks/spatial_window.cu
```diff
@@ -0,0 +1,114 @@
...
+    rmm::exec_policy(), d_points.begin(), d_points.end(), d_y.begin(), [] __device__(auto point) {
+      return point.y;
+    });
+
+  auto xs = cudf::column(cudf::data_type{cudf::type_to_id<T>()}, num_points, d_x.release());
+  auto ys = cudf::column(cudf::data_type{cudf::type_to_id<T>()}, num_points, d_y.release());
+
+  state.add_element_count(num_points);
+
+  points_in_spatial_window(window_min.x, window_max.x, window_min.y, window_max.y, xs, ys);
```
isVoid, line 103: Is this just for preheating? I believe nvbench already does that.

isVoid: Since data is prepared on device, would it be a good practice to explicitly sync stream before we start the benchmark?

isVoid (commented):
I have a few questions. But nothing stops it from merging.

cpp/benchmarks/spatial_window.cu
harrism: Tabulate requires no input -- it generates a sequence of zero-based integer indices and passes each to the provided unary operator, writing the output to the output range. Copy requires an input and an output.

cpp/benchmarks/spatial_window.cu
harrism: Yes that can be removed. And a sync might be good.

harrism: Done.

cpp/benchmarks/spatial_window.cu
harrism: Good idea. Done.

isVoid (approved):

harrism:
@gpucibot merge

PR #597, opened 2022-07-20
Build benchmarks in CI

vyasr:
This PR adds the ability to build benchmarks to build.sh, and passes the necessary option through from the conda build.sh. This change ensures that benchmarks must be able to build against changes in new PRs.

### DIFF
--- a/conda/recipes/libcuspatial/build.sh
+++ b/conda/recipes/libcuspatial/build.sh
@@ -1,4 +1,4 @@
 # Copyright (c) 2018-2022, NVIDIA CORPORATION.
 
 # build cuspatial with verbose output
-./build.sh -v libcuspatial tests --allgpuarch -n
+./build.sh -v libcuspatial tests benchmarks --allgpuarch -n
--- a/build.sh
+++ b/build.sh
@@ -18,12 +18,13 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libcuspatial cuspatial tests -v -g -n -h --allgpuarch --show_depr_warn"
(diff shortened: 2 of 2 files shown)


conda/recipes/libcuspatial/build.sh
```diff
@@ -1,4 +1,4 @@
 # Copyright (c) 2018-2022, NVIDIA CORPORATION.
 
 # build cuspatial with verbose output
-./build.sh -v libcuspatial tests --allgpuarch -n
+./build.sh -v libcuspatial tests benchmarks --allgpuarch -n
```
bdice, line 4: Does this change alter the contents of the conda package, or are the benchmarks building as a “side effect”?

harrism: I don't think we need to add them as part of this PR. If there's a need for them in the packages in the future we can revisit the decision. The main goal was to make sure we are compiling all benchmarks in CI so that we don't end up with broken benchmark builds.

Ethyling (approved):

conda/recipes/libcuspatial/build.sh
ajschmidt8: @vyasr, the `libcudf-tests` package are used in the GPU testing stage of CI.

i.e. here - https://github.com/rapidsai/cudf/blob/edc5062bdcc3e12755603b0ad07a4d271fe95261/ci/gpu/build.sh#L168

then we run the GTests like this: https://github.com/rapidsai/cudf/blob/edc5062bdcc3e12755603b0ad07a4d271fe95261/ci/gpu/build.sh#L183-L187

vyasr: OK got it thanks. Since we don't run benchmarks in CI, I think it's fine to not include the benchmarks in there for now. I suspect that `libcudf` mostly just includes them incidentally because they're part of the same CMake component, not for any particular design reason.

vyasr:
@gpucibot merge

PR #604, opened 2022-07-27
Memory_usage method for GeoSeries/GeoDataFrame and `GeoDataFrame` refactor.

thomcom:
This PR adds the memory_usage for the large, complex `GeoSeries` object, and tests for `GeoSeries` and `GeoDataframe`

This also includes a `GeoDataFrame` refactor that makes it possible for `cudf` to call `memory_usage` through `GeoDataFrame.super()` and works toward better integration with `cudf`.

Closes https://github.com/rapidsai/cuspatial/issues/714


harrism:
Please target 22.10 since burn down for 22.08 starts tonight.

harrism:
Draft status contradicts "ready to review" tag. Which is it?

thomcom:
Ehh definitely not ready for review due to good-ole-squash/merge. I'll need a minute... :D

thomcom:
rerun tests

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -130,10 +131,11 @@ def __init__(
             self.polygons.name = "polygons"
         else:
             raise TypeError("All four Tuple arguments must be cudf.ListSeries")
-        base = cudf.core.column.column.arange(0, len(self), dtype="int32").data
-        super().__init__(base, size=len(self), dtype="int32")
+        super().__init__(None, size=len(self), dtype="geometry")
         if shuffle_order is not None:
-            self._data = shuffle_order
+            self._gather_map = shuffle_order
```
isVoid, line 136: Ok... I see where it's going. `GeoColumn` wants to maintain an intrinsic order that's different from the data's physical layout. I can think of both side of pros and cons. But perhaps it's the best to let you explain the reasons behind this design decision.

isVoid: And it's an important detail that we should mention in `libary_design`.

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -126,50 +126,87 @@ def _copy_type_metadata(
...
-            sliced_geo_columns, sliced_data_columns
+        # combine
+        result = self._recombine_columns(geo_data, cudf_sorted)
+        # gather GeoColumns
+        [geo_data[geo]._set_gather_map(gather_map) for geo in geo_data.keys()]
+        # return
+        return result
+
+    def _get_sorted_inds(self, by=None, ascending=True, na_position="last"):
+        geo_data, cudf_data = self._split_out_geometry_columns()
```
isVoid, line 204:
```suggestion
        _, cudf_data = self._split_out_geometry_columns()
```

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -126,50 +126,87 @@ def _copy_type_metadata(
...
-            [isinstance(self[col], GeoSeries) for col in self.columns],
-            dtype="bool",
+            [
+                isinstance(self._data.get(col), GeoColumn)
+                for col in self._data.names
+            ]
         )
-        geo_columns = self[columns_mask[geocolumn_mask]]
+        geo_data = {
+            col: self._data.get(col) for col in columns_mask[geocolumn_mask]
```
isVoid, line 159:
```suggestion
            col: self._data[col] for col in columns_mask[geocolumn_mask]
```

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -126,50 +126,87 @@ def _copy_type_metadata(
...
+        }
+
         # Send the rest of the columns to `cudf` to slice.
-        data_columns = cudf.DataFrame(
-            self[columns_mask[~geocolumn_mask].values]
-        )
-        return (geo_columns, data_columns)
+        cudf_data = {
+            col: self._data.get(col)
```
isVoid, line 164:
```suggestion
            col: self._data[col]
```

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -126,50 +126,87 @@ def _copy_type_metadata(
...
+            for col in columns_mask[~geocolumn_mask].values
```
isVoid, line 165:
```suggestion
            for col in columns_mask[~geocolumn_mask]
```

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -126,50 +126,87 @@ def _copy_type_metadata(
 
         return type_copied
 
+    def _slice(self: T, arg: slice) -> T:
```
isVoid, line 129: If possible, next time let's keep the unchanged function in it's original place and add new function at different lines. Otherwise the diff message will become hard to read.

isVoid (commented):
`memory_usage` doesn't seem to use `_slice` and `_gather`. Are they separate features of this PR?

python/cuspatial/cuspatial/core/geodataframe.py
thomcom: Good idea. I could move it now.

python/cuspatial/cuspatial/core/_column/geocolumn.py
thomcom: Let me see how much of this I can recall. I needed to be taking notes during this process.

At first I wrote `GeoSeries.memory_usage`, but noticed that I needed to support `GeoDataFrame.memory_usage`. It is inherited from `DataFrame`, but was failing because `DataFrame.memory_usage` calls the `Series._from_data` function, which was failing because, I think, the `base` data as `int32` wasn't compatible. I'm going to see if I can reproduce the failure case that lead to this rabbit holing. It's possible it should be separated into another PR, but I don't think so.

thomcom: Ohhhh haha. What you're looking at here is part of #709, which I incorrectly said was #585 in the description. How about you review #709 first, then after merging it this should be simple again.

thomcom: In #585, using `super()._init_from_dict` like causes a host of internal `cudf.DataFrame` methods to be called.

isVoid (approved):

thomcom:
@gpucibot merge

thomcom:
@gpucibot merge

PR #609, opened 2022-07-29
[FEA] Integrate `GeoSeries` with `read_polygon_shapefile`

### LINKED ISSUE #668: [FEA] Add offset column accessors
Our APIs mostly use offset columns passed in as arguments, though @isVoid is making progress to refactoring them to all use `GeoSeries` as input instead. To make development of `GeoSeries` integration easier and to make it much easier to continue to use our algorithms, while also using `GeoSeries` objects, I need to add a `offset_column` accessor to each of the four underlying geometry types stored in `GeoSeries`: `points`, `multipoints`, `lines`, and `polygons`.

I'm currently working on this as

`geoseries.polygons.offset_column(0...3)` in #609. #609 will resolve this.

thomcom:
This useful PR adds a `GeoSeries` constructor for the tuple returned by `read_polygon_shapefile`. Now users can say

```py
geoseries = cuspatial.GeoSeries(cuspatial.read_polygon_shapefile('the_shapefile'))
```

and load the data directly into GeoArrow.

This PR touches many files because I add a `reversed` argument to python and C++ so that the user can swap back and forth if they desire. This was required because the original implementation was reverse-ordering the polygons, which does not match GeoPandas results when reading a shapefile. Now it does, and is tested in both configurations.

Closes #668.

### DIFF
--- a/cpp/include/cuspatial/shapefile_reader.hpp
+++ b/cpp/include/cuspatial/shapefile_reader.hpp
@@ -23,6 +23,8 @@
 
 namespace cuspatial {
 
+typedef enum class winding_order : bool { CLOCKWISE, COUNTER_CLOCKWISE } winding_order;
+
 /**
  * @addtogroup io
  * @{
@@ -32,6 +34,8 @@ namespace cuspatial {
  * @brief read polygon data from an ESRI Shapefile.
  *
  * @param[in] filename: ESRI Shapefile file path (usually ends in .shp)
+ * @param[in] outer_ring_winding: the ordering of the outer ring of polygon vertices; clockwise or
+ * counter-clockwise
  * @param[in] mr:       Optional, The resource to use to allocate the returned data
  *
  * @return Vector of 4 columns representing one or more polygons:
@@ -45,7 +49,8 @@ namespace cuspatial {
  **/
 std::vector<std::unique_ptr<cudf::column>> read_polygon_shapefile(
   std::string const& filename,
-  rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
+  const winding_order outer_ring_winding = winding_order::COUNTER_CLOCKWISE,
+  rmm::mr::device_memory_resource* mr    = rmm::mr::get_current_device_resource());
 
 /**
  * @} // end of doxygen group
--- a/cpp/src/io/shp/polygon_shapefile_reader.cpp
+++ b/cpp/src/io/shp/polygon_shapefile_reader.cpp
@@ -14,11 +14,15 @@
  * limitations under the License.
  */
 
+#include <cuspatial/detail/iterator.hpp>
 #include <cuspatial/error.hpp>
+#include <cuspatial/shapefile_reader.hpp>
 
 #include <ogrsf_frmts.h>
 
 #include <cudf/types.hpp>
+
+#include <thrust/iterator/counting_iterator.h>
 
 #include <memory>
 #include <string>
@@ -26,16 +30,32 @@
 
 namespace {
 
+// TODO: a better approach is to precompute the total size needed by xs and ys
+// as a preprocess and preallocate xs, ys. `read_ring` should simply accept
+// an iterator range and return the pass-the-end iterator.
 cudf::size_type read_ring(OGRLinearRing const& ring,
                           std::vector<double>& xs,
-                          std::vector<double>& ys)
+                          std::vector<double>& ys,
+                          const cuspatial::winding_order ring_order)
 {
-  cudf::size_type num_vertices = ring.getNumPoints();
+  auto num_vertices      = ring.getNumPoints();
+  auto prev_num_vertices = xs.size();
+  xs.resize(xs.size() + num_vertices);
+  ys.resize(ys.size() + num_vertices);
 
-  // append points in reverse order
-  for (cudf::size_type i = num_vertices - 1; i >= 0; i--) {
-    xs.push_back(ring.getX(i));
-    ys.push_back(ring.getY(i));
+  auto output_it = thrust::make_zip_iterator(
+    thrust::make_tuple(xs.begin() + prev_num_vertices, ys.begin() + prev_num_vertices));
+  if (ring_order == cuspatial::winding_order::COUNTER_CLOCKWISE) {
+    auto ring_it = cuspatial::detail::make_counting_transform_iterator(
+      0, [&ring](auto i) { return thrust::make_tuple(ring.getX(i), ring.getY(i)); });
+
+    std::copy_n(ring_it, num_vertices, output_it);
+  } else {
+    auto ring_it =
+      cuspatial::detail::make_counting_transform_iterator(0, [&ring, &num_vertices](auto i) {
+        return thrust::make_tuple(ring.getX(num_vertices - i - 1), ring.getY(num_vertices - i - 1));
+      });
+    std::copy_n(ring_it, num_vertices, output_it);
   }
 
   return num_vertices;
@@ -44,15 +64,16 @@ cudf::size_type read_ring(OGRLinearRing const& ring,
 cudf::size_type read_polygon(OGRPolygon const& polygon,
                              std::vector<int>& ring_lengths,
                              std::vector<double>& xs,
-                             std::vector<double>& ys)
+                             std::vector<double>& ys,
+                             const cuspatial::winding_order ring_order)
 {
-  auto num_vertices = read_ring(*(polygon.getExteriorRing()), xs, ys);
+  auto num_vertices = read_ring(*(polygon.getExteriorRing()), xs, ys, ring_order);
   ring_lengths.push_back(num_vertices);
 
   cudf::size_type num_interior_rings = polygon.getNumInteriorRings();
 
   for (cudf::size_type i = 0; i < num_interior_rings; i++) {
-    auto num_vertices = read_ring(*(polygon.getInteriorRing(i)), xs, ys);
+    auto num_vertices = read_ring(*(polygon.getInteriorRing(i)), xs, ys, ring_order);
     ring_lengths.push_back(num_vertices);
   }
 
@@ -62,12 +83,15 @@ cudf::size_type read_polygon(OGRPolygon const& polygon,
 cudf::size_type read_geometry_feature(OGRGeometry const* geometry,
                                       std::vector<int>& ring_lengths,
                                       std::vector<double>& xs,
-                                      std::vector<double>& ys)
+                                      std::vector<double>& ys,
+                                      const cuspatial::winding_order ring_order)
 {
   OGRwkbGeometryType geometry_type = wkbFlatten(geometry->getGeometryType());
 
   if (geometry_type == wkbPolygon) {
-    return read_polygon(*((OGRPolygon*)geometry), ring_lengths, xs, ys);
+    auto polygon = dynamic_cast<const OGRPolygon*>(geometry);
+    if (polygon == nullptr) { CUSPATIAL_FAIL("Can't cast `wkbPolygon` to `OGRPolygon&`"); }
+    return read_polygon(*polygon, ring_lengths, xs, ys, ring_order);
   }
 
   if (geometry_type == wkbMultiPolygon || geometry_type == wkbGeometryCollection) {
@@ -76,8 +100,8 @@ cudf::size_type read_geometry_feature(OGRGeometry const* geometry,
     int num_rings = 0;
 
     for (int i = 0; i < geometry_collection->getNumGeometries(); i++) {
-      num_rings +=
-        read_geometry_feature(geometry_collection->getGeometryRef(i), ring_lengths, xs, ys);
+      num_rings += read_geometry_feature(
+        geometry_collection->getGeometryRef(i), ring_lengths, xs, ys, ring_order);
     }
 
     return num_rings;
@@ -90,7 +114,8 @@ cudf::size_type read_layer(const OGRLayerH layer,
                            std::vector<cudf::size_type>& feature_lengths,
                            std::vector<cudf::size_type>& ring_lengths,
                            std::vector<double>& xs,
-                           std::vector<double>& ys)
+                           std::vector<double>& ys,
+                           const cuspatial::winding_order ring_order)
 {
   cudf::size_type num_features = 0;
 
@@ -99,11 +124,11 @@ cudf::size_type read_layer(const OGRLayerH layer,
   OGRFeatureH feature;
 
   while ((feature = OGR_L_GetNextFeature(layer)) != nullptr) {
-    auto geometry = (OGRGeometry*)OGR_F_GetGeometryRef(feature);
+    auto geometry = static_cast<OGRGeometry*>(OGR_F_GetGeometryRef(feature));
 
     CUSPATIAL_EXPECTS(geometry != nullptr, "Invalid Shape");
 
-    auto num_rings = read_geometry_feature(geometry, ring_lengths, xs, ys);
+    auto num_rings = read_geometry_feature(geometry, ring_lengths, xs, ys, ring_order);
 
     feature_lengths.push_back(num_rings);
 
@@ -124,7 +149,7 @@ std::tuple<std::vector<cudf::size_type>,
            std::vector<cudf::size_type>,
            std::vector<double>,
            std::vector<double>>
-read_polygon_shapefile(std::string const& filename)
+read_polygon_shapefile(std::string const& filename, cuspatial::winding_order outer_ring_winding)
 {
   GDALAllRegister();
 
@@ -141,8 +166,7 @@ read_polygon_shapefile(std::string const& filename)
   std::vector<double> xs;
   std::vector<double> ys;
 
-  read_layer(dataset_layer, feature_lengths, ring_lengths, xs, ys);
-
+  read_layer(dataset_layer, feature_lengths, ring_lengths, xs, ys, outer_ring_winding);
   feature_lengths.shrink_to_fit();
   ring_lengths.shrink_to_fit();
   xs.shrink_to_fit();
--- a/cpp/tests/io/shp/polygon_shapefile_reader_test.cpp
+++ b/cpp/tests/io/shp/polygon_shapefile_reader_test.cpp
@@ -15,7 +15,10 @@
  */
 
 #include <cuspatial/error.hpp>
+#include <cuspatial/point_in_polygon.hpp>
 #include <cuspatial/shapefile_reader.hpp>
+
+#include <rmm/device_vector.hpp>
 
 #include <cudf_test/base_fixture.hpp>
 #include <cudf_test/column_utilities.hpp>
@@ -45,8 +48,30 @@ void test(std::string const& shapefile_name,
           std::vector<double> xs,
           std::vector<double> ys)
 {
-  auto shape_filename  = get_shapefile_path(shapefile_name);
-  auto polygon_columns = cuspatial::read_polygon_shapefile(shape_filename);
+  auto shape_filename = get_shapefile_path(shapefile_name);
+  auto polygon_columns =
+    cuspatial::read_polygon_shapefile(shape_filename, cuspatial::winding_order::COUNTER_CLOCKWISE);
+
+  auto expected_poly_offsets  = wrapper<cudf::size_type>(poly_offsets.begin(), poly_offsets.end());
+  auto expected_ring_offsets  = wrapper<cudf::size_type>(ring_offsets.begin(), ring_offsets.end());
+  auto expected_poly_point_xs = wrapper<double>(xs.begin(), xs.end());
+  auto expected_poly_point_ys = wrapper<double>(ys.begin(), ys.end());
+
+  expect_columns_equivalent(expected_poly_offsets, polygon_columns.at(0)->view(), verbosity);
+  expect_columns_equivalent(expected_ring_offsets, polygon_columns.at(1)->view(), verbosity);
+  expect_columns_equivalent(expected_poly_point_xs, polygon_columns.at(2)->view(), verbosity);
+  expect_columns_equivalent(expected_poly_point_ys, polygon_columns.at(3)->view(), verbosity);
+}
+
+void test_reverse(std::string const& shapefile_name,
+                  std::vector<cudf::size_type> poly_offsets,
+                  std::vector<cudf::size_type> ring_offsets,
+                  std::vector<double> xs,
+                  std::vector<double> ys)
+{
+  auto shape_filename = get_shapefile_path(shapefile_name);
+  auto polygon_columns =
+    cuspatial::read_polygon_shapefile(shape_filename, cuspatial::winding_order::CLOCKWISE);
 
   auto expected_poly_offsets  = wrapper<cudf::size_type>(poly_offsets.begin(), poly_offsets.end());
   auto expected_ring_offsets  = wrapper<cudf::size_type>(ring_offsets.begin(), ring_offsets.end());
@@ -70,9 +95,14 @@ TEST_F(PolygonShapefileReaderTest, NonExistentFile)
 
 TEST_F(PolygonShapefileReaderTest, ZeroPolygons) { test("empty_poly.shp", {}, {}, {}, {}); }
 
+TEST_F(PolygonShapefileReaderTest, OnePolygonReversed)
+{
+  test_reverse("one_poly.shp", {0}, {0}, {-10, 5, 5, -10, -10}, {-10, -10, 5, 5, -10});
+}
+
 TEST_F(PolygonShapefileReaderTest, OnePolygon)
 {
-  test("one_poly.shp", {0}, {0}, {-10, 5, 5, -10, -10}, {-10, -10, 5, 5, -10});
+  test("one_poly.shp", {0}, {0}, {-10, -10, 5, 5, -10}, {-10, 5, 5, -10, -10});
 }
 
 TEST_F(PolygonShapefileReaderTest, TwoPolygons)
@@ -80,6 +110,24 @@ TEST_F(PolygonShapefileReaderTest, TwoPolygons)
   test("two_polys.shp",
        {0, 1},
        {0, 5},
-       {-10, 5, 5, -10, -10, 0, 10, 10, 0, 0},
-       {-10, -10, 5, 5, -10, 0, 0, 10, 10, 0});
+       {-10, -10, 5, 5, -10, 0, 0, 10, 10, 0},
+       {-10, 5, 5, -10, -10, 0, 10, 10, 0, 0});
+}
+
+TEST_F(PolygonShapefileReaderTest, OnePointInPolygon)
+{
+  auto shape_filename  = get_shapefile_path("one_poly.shp");
+  auto polygon_columns = cuspatial::read_polygon_shapefile(shape_filename);
+
+  auto polygons = polygon_columns.at(0)->view();
+  auto rings    = polygon_columns.at(1)->view();
+  auto xs       = polygon_columns.at(2)->view();
+  auto ys       = polygon_columns.at(3)->view();
+  fixed_width_column_wrapper<double> test_xs({0.0});
+  fixed_width_column_wrapper<double> test_ys({0.0});
+  fixed_width_column_wrapper<int32_t> expected({true});
+
+  auto ret = cuspatial::point_in_polygon(test_xs, test_ys, polygons, rings, xs, ys);
+
+  expect_columns_equivalent(ret->view(), expected);
 }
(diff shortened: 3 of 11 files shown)


zhangjianting:
There are very informative discussions on the ordering of polygon vertices: https://gis.stackexchange.com/questions/119150/order-of-polygon-vertices-in-general-gis-clockwise-or-counterclockwise 
The OGC SFS and ESRI shapefile have different orderings. As long as the inner rings and the outer rings have different directions, PIP algorithms should still work.  Also see 
https://wrfranklin.org/Research/Short_Notes/pnpoly.html#Listing%20the%20Vertices

zhangjianting:
The Shapefile does have a problem in dealing with multi-polygons. The initial design was to break multi-polygons (as well as GeometryCollection) into multiple (single/regular) polygons (with or without holes). Yet the implementations was to consolidate the rings of multi-polygons, which essentially treats the first ring as the outer ring and the rest as the inner rings, i.e., a single/regular polygon with holes. Subsequently, it will break PIP algorithms. There are two solutions: 
1) Simply declaring not supporting multi-polygons or GeometryCollection.
2) Modify the Shapefile reader and extend the code of all relevant operations to truly support  multi-polygons.

zhangjianting:
One more comment: the current code does support GeometryCollection of polygons that have only one ring.

cpp/tests/io/shp/polygon_shapefile_reader_test.cpp
```diff
@@ -72,14 +75,32 @@ TEST_F(PolygonShapefileReaderTest, ZeroPolygons) { test("empty_poly.shp", {}, {}
 
 TEST_F(PolygonShapefileReaderTest, OnePolygon)
 {
-  test("one_poly.shp", {0}, {0}, {-10, 5, 5, -10, -10}, {-10, -10, 5, 5, -10});
+  test("one_poly.shp", {0}, {0}, {-10, -10, 5, 5, -10}, {-10, 5, 5, -10, -10});
```
isVoid, line 78: Instead of changing the test, we should add an extra test case to verify the `reverse=False` case.

python/cuspatial/cuspatial/_lib/cpp/shapefile_reader.pxd
```diff
@@ -9,4 +9,6 @@ from cudf._lib.cpp.column.column cimport column
 
 cdef extern from "cuspatial/shapefile_reader.hpp" namespace "cuspatial" nogil:
     cdef vector[unique_ptr[column]] \
-        read_polygon_shapefile(const string filename) except +
+        read_polygon_shapefile(
+            const string filename, const bint reversed
```
isVoid, line 13: You would need `from libcpp import bool`
```suggestion
            const string filename, const bool reversed
```

isVoid (commented):
For discussion: would you think it's better to support factory functions like `GeoColumn.from_shapefile` and `GeoSeries.from_shapefile`? We can still support constructing the objects in constructor, but under the hood punt to the factory functions.

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -26,16 +26,25 @@
 
 namespace {
 
+bool _reverse_winding = false;
```
harrism, line 29: Please, no global variables. Make this a parameter of the function.

```suggestion
```

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -26,16 +26,25 @@
...
+
 cudf::size_type read_ring(OGRLinearRing const& ring,
                           std::vector<double>& xs,
                           std::vector<double>& ys)
```
harrism, line 33:
```suggestion
                          std::vector<double>& ys,
                          const winding_order ring_order)
```

harrism: Let's also replace raw loops with algorithms, and rely on OGR's iterator support. (Sadly they don't provide `rbegin()` and `rend()`, but presumably we can adapt using `std::reverse_iterator`.

The whole function would look something like this.

```
int read_ring(std::vector<point> const& ring,
                          std::vector<double>& xs,
                          std::vector<double>& ys,
                          winding_order ring_order)
{
  xs.reserve(ring.size());
  ys.reserve(ring.size());

  auto vert_to_tuple = [](auto vertex) { return thrust::make_tuple(vertex.getX(), vertex.getY()); };

  if (ring_order == winding_order::CLOCKWISE) {
    std::transform(ring.begin(), ring.end(), thrust::make_zip_iterator(thrust::make_tuple(xs.begin(), ys.begin())),
      vert_to_tuple);
  } else {
    std::transform(std::reverse_iterator(ring.end()), std::reverse_iterator(ring.begin()),
      thrust::make_zip_iterator(thrust::make_tuple(xs.begin(), ys.begin())), vert_to_tuple);
  }

  return ring.size();
}

Comment (harrism) [reply cpp/src/io/shp/polygon_shapefile_reader.cpp:33]: Note that below, where you call this function, you will need to reverse the direction at each level of nested rings (assuming we support multiple levels of nesting).  For this, I would suggest defining a convenience function:

```
winding_order reverse_winding(winding_order order) {
  return static_cast<winding_order>(!static_cast<bool>(order));
}
```

File: cpp/src/io/shp/polygon_shapefile_reader.cpp [hunk 5]
```diff
@@ -124,7 +133,7 @@ std::tuple<std::vector<cudf::size_type>,
            std::vector<cudf::size_type>,
            std::vector<double>,
            std::vector<double>>
-read_polygon_shapefile(std::string const& filename)
+read_polygon_shapefile(std::string const& filename, const bool reversed = false)
```
harrism, line 136:
```suggestion
read_polygon_shapefile(std::string const& filename, const winding_order outer_ring_order = winding_order::CLOCKWISE)
```

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -141,6 +150,7 @@ read_polygon_shapefile(std::string const& filename)
   std::vector<double> xs;
   std::vector<double> ys;
 
+  _reverse_winding = reversed;
```
harrism, line 153: Remove global variable. Plumb the parameter down through the internal functions.

python/cuspatial/cuspatial/geometry/geocolumn.py
```diff
@@ -30,23 +31,89 @@ def __init__(
...
+            rings_col = data[1].astype("int32")
+            coordinates = (
+                data[2].stack().astype("float64").reset_index(drop=True)
+            )
+            coordinate_offsets = cudf.concat(
+                [
+                    cudf.Series(
+                        cp.arange(len(coordinates) // 2) * 2, dtype="int32"
+                    ),
+                    cudf.Series([len(coordinates) // 2 * 2], dtype="int32"),
```
harrism, line 47: What is this arithmetic doing? Perhaps document it. (Also  below for the point offsets).

harrism (changes requested):
I think handling winding order can be cleaner, made some suggestions.

cpp/src/io/shp/polygon_shapefile_reader.cpp
thomcom: But it is so many! Isn't there a neater way? :)

cpp/include/cuspatial/shapefile_reader.hpp
```diff
@@ -45,6 +46,7 @@ namespace cuspatial {
  **/
 std::vector<std::unique_ptr<cudf::column>> read_polygon_shapefile(
   std::string const& filename,
+  const bool reversed                 = false,
```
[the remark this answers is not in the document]
thomcom: I'm not sure if we can specify `CLOCKWISE` or `COUNTER_CLOCKWISE` here because we don't necessarily know the order of the polygons that are about to be read. That's why I just went with reversed.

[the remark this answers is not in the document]
harrism: It's a parameter. The caller tells you the order. Once on inner rings you are just alternating.

cpp/src/io/shp/polygon_shapefile_reader.cpp
harrism: read_polygon_shapefile, read_layer, read_ring. That's not many in my book.

cpp/include/cuspatial/shapefile_reader.hpp
```diff
@@ -32,7 +37,7 @@ namespace cuspatial {
  * @brief read polygon data from an ESRI Shapefile.
  *
  * @param[in] filename: ESRI Shapefile file path (usually ends in .shp)
- * @param[in] reversed: Change the ordering of the polygons, clockwise or counter-clockwise
+ * @param[in] outer_ring_winding: Change the ordering of the polygons, clockwise or counter-clockwise
```
harrism, line 40:
```suggestion
 * @param[in] outer_ring_winding: the ordering of the outer ring of polygon vertices; clockwise or counter-clockwise
```

cpp/include/cuspatial/shapefile_reader.hpp
```diff
@@ -23,6 +23,11 @@
 
 namespace cuspatial {
 
+typedef enum winding_order : bool {
```
harrism, line 26: For type safety.

```suggestion
typedef enum class winding_order : bool {
```

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -26,16 +27,15 @@
 
 namespace {
 
-bool _reverse_winding = false;
-
 cudf::size_type read_ring(OGRLinearRing const& ring,
                           std::vector<double>& xs,
-                          std::vector<double>& ys)
+                          std::vector<double>& ys,
+                          const cuspatial::winding_order ring_order)
```
harrism, line 33: I don't think const is necessary here.

```suggestion
                          cuspatial::winding_order ring_order)
```

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -53,15 +53,16 @@ cudf::size_type read_ring(OGRLinearRing const& ring,
...
+                             std::vector<double>& ys,
+                             const cuspatial::winding_order ring_order)
 {
-  auto num_vertices = read_ring(*(polygon.getExteriorRing()), xs, ys);
+  auto num_vertices = read_ring(*(polygon.getExteriorRing()), xs, ys, ring_order);
   ring_lengths.push_back(num_vertices);
 
   cudf::size_type num_interior_rings = polygon.getNumInteriorRings();
 
   for (cudf::size_type i = 0; i < num_interior_rings; i++) {
```
harrism, line 64: Where is the order flipping happening? At each nested level the order should be reversed.

cpp/include/cuspatial/shapefile_reader.hpp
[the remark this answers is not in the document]
thomcom: Er, right. :D Your suggested improvements make me feel a lot better about my contribution, rather than using a global.

cpp/src/io/shp/polygon_shapefile_reader.cpp
thomcom: :+1:

cpp/include/cuspatial/shapefile_reader.hpp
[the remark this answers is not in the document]
thomcom: I like the refactor according to your suggestion, the code feels cleaner. I still worry though because we don't explicitly know the winding of the rings as-input: if a shapefile happens to have CLOCKWISE wound rings and we specify CLOCKWISE as the winding, they will be the opposite. I think I will put in the docs that it is assumed that all input shapefiles will natively have COUNTER_CLOCKWISE ordering.

zhangjianting:
There is a simple way to determine the orientation of a ring/curve, e.g., signed area. 
https://en.wikipedia.org/wiki/Curve_orientation

thomcom:
Ah nice! Computing the handedness of the polygon is nothing more than the determinant of the matrix formed by the first three points. That handedness is guaranteed to continue through all of the rest of the points. It is tempting to implement this for completeness.

thomcom:
We should definitely have an area computation as well, and it looks quite simple. I'll file an issue for it.

python/cuspatial/cuspatial/geometry/geocolumn.py
thomcom: Done

cpp/include/cuspatial/shapefile_reader.hpp
thomcom: Can I get you to bend on this one? The enum is linked to via Cython and Cython doesn't like using the `class` identifier here. I don't know if there is a workaround and finding one might take hours.

cpp/src/io/shp/polygon_shapefile_reader.cpp
thomcom: Order flipping occurs in the interior rings as well in the next couple of [lines](https://github.com/rapidsai/cuspatial/pull/609/files/a971db87aa08b253764f795844c11a3d305e1f74..dae0fe361808676cb8456994c90121c51468a640#diff-ff99c11a9d678705f3f645969a9e6d8d684e5c8e0f3ec589967bcc29fca58a48R65)

harrism: Seems to be it's just passed down through the rings without alternating orders.

cpp/include/cuspatial/shapefile_reader.hpp
thomcom: Ok, thanks to @shwina this is resolved.

cpp/src/io/shp/polygon_shapefile_reader.cpp
thomcom: The interior rings are assumed to be opposite winding of the outer ring. The argument only specifies the winding for the outer ring, so if we wind the outer ring reverse, all the children rings get reversed as well. Otherwise neither are changed.

thomcom:
Ready for your re-review @harrism @isVoid

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -14,28 +14,48 @@
...
+  xs.resize(xs.size() + num_vertices);
+  ys.resize(ys.size() + num_vertices);
+
+  auto output_it = thrust::make_zip_iterator(
+    thrust::make_tuple(xs.begin() + prev_num_vertices, ys.begin() + prev_num_vertices));
+  if (ring_order == cuspatial::winding_order::COUNTER_CLOCKWISE) {
+    auto ring_it = cuspatial::detail::make_counting_transform_iterator(
+      0, [&ring](auto i) { return thrust::make_tuple(ring.getX(i), ring.getY(i)); });
+
+    thrust::copy_n(ring_it, num_vertices, output_it);
```
harrism, line 52: Both input and output data are on the host, right? If so, I would just use `std::copy_n` instead.

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -62,12 +83,13 @@ cudf::size_type read_polygon(OGRPolygon const& polygon,
...
+                                      const cuspatial::winding_order ring_order)
 {
   OGRwkbGeometryType geometry_type = wkbFlatten(geometry->getGeometryType());
 
   if (geometry_type == wkbPolygon) {
-    return read_polygon(*((OGRPolygon*)geometry), ring_lengths, xs, ys);
+    return read_polygon(*((OGRPolygon*)geometry), ring_lengths, xs, ys, ring_order);
   }
 
   if (geometry_type == wkbMultiPolygon || geometry_type == wkbGeometryCollection) {
```
harrism, line 95: I can't comment on line 96 directly, but it uses a C-style cast. That has a really bad code smell and should be eliminated.  For this code to be correct, you need a `dynamic_cast`.  That may cause this code to not work if compiled with RTTI disabled (we don't do that by default, so I think it's OK).  With `dynamic_cast` you need to check the value of the result for `nullptr` and throw an exception if it is.

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -103,7 +126,7 @@ cudf::size_type read_layer(const OGRLayerH layer,
```
harrism, line 126: Can't comment directly on line 125, but please eliminate the C-style cast there. If a cast is needed there, it should probably be a `static_cast`.

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,118 @@ def __init__(
...
-            or not isinstance(data[2], cudf.Series)
-            or not isinstance(data[3], cudf.Series)
+        if from_read_polygon_shapefile:
+            """
+            A cudf.ListSeries needs four levels of nesting to represent
+            a polygon shapefile. The rings_offsets and polygons_offsets
+            have already been computed in the `read_polygon_shapefile`
+            function. In order to convert it into an arrow list<...>
+            we need a set of offsets buffers for each point tuple, and
+            and also an offsets buffer for each individual point.
```
harrism, line 43:
```suggestion
            an offsets buffer for each individual point.
```

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,118 @@ def __init__(
...
+
+            Coordinates: List of size 2 offsets
```
harrism, line 45: What is a "size 2 offset"?

harrism (changes requested):
Nearly there. A couple of smelly C-style casts need to be fixed up.

cpp/include/cuspatial/experimental/points_in_range.cuh
```diff
@@ -20,6 +20,8 @@
 
 #include <rmm/cuda_stream_view.hpp>
 
+#include <thrust/copy.h>
+#include <thrust/count.h>
```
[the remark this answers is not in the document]
thomcom: I put them in back when the header updates for thrust 17 caused many issues. It looks like they are no longer required to build.

cpp/src/io/shp/polygon_shapefile_reader.cpp
thomcom: :+1:

python/cuspatial/cuspatial/core/_column/geocolumn.py
thomcom: I changed it to "length 2" to match the below line of length 1 offsets. I also added a small example.

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,118 @@ def __init__(
...
+            Up to the size of the original input.
+            """
+            coordinate_offsets = cudf.concat(
+                [
+                    cudf.Series(
+                        cp.arange(len(coordinates) // 2) * 2, dtype="int32"
+                    ),
+                    cudf.Series([len(coordinates) // 2 * 2], dtype="int32"),
+                ]
+            ).reset_index(drop=True)
```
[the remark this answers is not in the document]
thomcom: Thanks for this comment. Presently `GeoColumn` is a `ListColumn`, but it uses an extra column for the coordinate offset array since `ListColumn` does not support `FixedWidthList` from GeoArrow.

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,118 @@ def __init__(
...
+                children=(polygons_offsets._column, rings),
+            )
+            self.points = cudf.Series([])
+            self.points.name = "points"
+            self.mpoints = cudf.Series([])
+            self.mpoints.name = "mpoints"
+            self.lines = cudf.Series([])
+            self.lines.name = "lines"
+            self.polygons = cudf.Series(parent)
+            self.polygons.name = "polygons"
```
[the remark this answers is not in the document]
thomcom: A `GeoColumn` is actually a `Column` and three `ListColumns`, right? I appreciate your curiousity here and I'm not sure what the answer is. If our `read_shapefile` methods supported geometry types other than `polygon`, then `points`, `mpoints`, and `lines` would appear using `read_polygon_shapefile`.

I'm not sure what context you're using to suggest creating these attributes from a `cudf.Series`, however. So far I don't think that is a use case. How are you thinking of it?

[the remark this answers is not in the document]
isVoid: Ah, I might have confused myself here. It's a read polygon input and thus it's natural to have no data with points, mpoints and lines. Please ignore this comment so far.

python/cuspatial/cuspatial/core/_column/geocolumn.py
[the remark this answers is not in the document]
isVoid: We can model the geometry arrays as `ListColumn`. This makes the actual arrow format of the arrays to slightly deviate from the geo-arrow spec, for example, polygon:
`List<List<FixedSizeList<double>[2]>>` v.s. `List<List<List<double>>`. But if this makes implementation cleaner, that's a fair tradeoff to make.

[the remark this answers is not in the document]
thomcom: Yes, that is how it is currently implemented exactly.

isVoid, line 75: This essentially is just:
```suggestion
            coordinate_offsets = cudf.Series(cupy.arrange(0, len(coordinates) + 1, 2), dtype="int32")
```
right?

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,118 @@ def __init__(
...
+            """
+            A cudf.ListSeries needs four levels of nesting to represent
+            a polygon shapefile. The rings_offsets and polygons_offsets
+            have already been computed in the `read_polygon_shapefile`
+            function. In order to convert it into an arrow list<...>
+            we need a set of offsets buffers for each point tuple, and
+            an offsets buffer for each individual point.
+
+            Coordinates: List of length 2 offsets: [0, 2, 4, ... n/2]
+            Points: List of length 1 offsets: No multipoints
```
isVoid, line 46: I don't think this level of indirection is needed. Each element of a ring must be a single-point. With this indirection a polygon type becomes `List<List<List<List<double>>>>`, with the last two level grouping points into multipoints.
But geoarrow spec is `List<List<List<double>>>`.

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,118 @@ def __init__(
...
+            rings = cudf.core.column.ListColumn(
+                size=len(polygons_offsets) - 1,
+                dtype=cudf.ListDtype(points.dtype),
+                children=(polygons_offsets._column, points),
+            )
+            parent = cudf.core.column.ListColumn(
+                size=len(polygons_offsets) - 1,
+                dtype=cudf.ListDtype(rings.dtype),
+                children=(polygons_offsets._column, rings),
+            )
```
isVoid, line 118: I think `points` ListColumn is not needed. The number of rings should equal to the length of `ring_offsets - 1`, not `polygon_offsets-1`.

```suggestion
            coords = cudf.core.column.ListColumn(
                size=len(coordinate_offsets) - 1,
                dtype=cudf.ListDtype(coordinates.dtype),
                children=(coordinate_offsets._column, coordinates._column),
            )
            rings = cudf.core.column.ListColumn(
                size=len(rings_offsets) - 1,
                dtype=cudf.ListDtype(points.dtype),
                children=(rings_offsets._column, coords),
            )
            parent = cudf.core.column.ListColumn(
                size=len(polygons_offsets) - 1,
                dtype=cudf.ListDtype(rings.dtype),
                children=(polygons_offsets._column, rings),
            )
```

python/cuspatial/cuspatial/core/_column/geocolumn.py
thomcom: I agree. I tried to do `List<List<List<double>>>` but it did not match what was produced by `from_geopandas` on the same dataset, which lead to discovering that `ListColumn` appears to be producing the `coordinates` and `points` buffers,

python/cuspatial/cuspatial/core/_column/geocolumn.py
thomcom: I would prefer this, but it simply doesn't produce the same data structure as what is made in `from_geopandas`. Maybe there is a bug there...

python/cuspatial/cuspatial/core/_column/geocolumn.py
isVoid: Can you give a reproducer for this example? It sounds to me if `from_geopandas` is doing something funky to introduce an extra level of indirection.

python/cuspatial/cuspatial/core/_column/geocolumn.py
thomcom: The bug had to do with my naming scheme, which was confused. In the latest PR the structures are now named appropriately for what they represent. Thanks for noticing!

python/cuspatial/cuspatial/core/_column/geocolumn.py
thomcom: We figured out chatting that the extra buffer is for the `Multipolygons` offset array, which for a set of polygons with only exteriors, matches the `polygons` offset array exactly.

thomcom:
@harrism @isVoid this is cleaner and readier for another round of review.

python/cuspatial/cuspatial/core/_column/geocolumn.py
```diff
@@ -30,23 +31,102 @@ def __init__(
...
+            polygons = cudf.core.column.ListColumn(
+                size=len(polygons_offsets) - 1,
+                dtype=cudf.ListDtype(rings.dtype),
+                children=(polygons_offsets._column, rings),
+            )
+            mpolygons = cudf.core.column.ListColumn(
+                size=len(polygons_offsets) - 1,
+                dtype=cudf.ListDtype(polygons.dtype),
+                children=(polygons_offsets._column, polygons),
+            )
```
isVoid, line 102: Hmmm, I think the multi-polygon offsets should be a sequence? e.g. `[0, 1, 2, 3, 4...]` so that the each of the multi-polygon contains only a single polygon.

```suggestion
            mpolygons = cudf.core.column.ListColumn(
                size=len(polygons_offsets) - 1,
                dtype=cudf.ListDtype(polygons.dtype),
                children=(cupy.arange(len(polygons) + 1), polygons),
            )
```

thomcom: Oh good point, I was writing that code based on my test data. :eyes:

cpp/include/cuspatial/shapefile_reader.hpp
```diff
@@ -23,6 +23,11 @@
 
 namespace cuspatial {
 
+typedef enum class winding_order : bool {
+  CLOCKWISE,
+  COUNTER_CLOCKWISE  // or "anticlockwise" or "widdershins", if you prefer
```
isVoid, line 28: I'm not sure if this is a stale comment, or do you mean to say this is sometimes referred to as "anticlockwise"/"widdershins" in other literature? We should remove this in the former, or change the comment for the latter.
```suggestion
  COUNTER_CLOCKWISE
```

isVoid (approved):
Thanks for addressing all comments. I think you should update the PR description from mentioning `reversed` to `winding_order`.

cpp/include/cuspatial/shapefile_reader.hpp
harrism: I inserted that comment purely as a linguistic joke when I originally made the suggestion. Please delete.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -98,6 +122,25 @@ def __init__(self, list_series):
             self._series = list_series
             self._col = self._series._column
 
+        def offset_buffer(self, index):
```
harrism, line 125: What is this doing? It's not obvious from the function name, so I think some docs would help.

harrism (approved):

python/cuspatial/cuspatial/core/geoseries.py
thomcom: I'm going to revert this and use the accessors that @isVoid wrote in https://github.com/rapidsai/cuspatial/pull/660/files#diff-c522c9afb3364b1aed2b2589c0d0c260dbc634bc54844536b1d382cecb92bf29

cpp/include/cuspatial/shapefile_reader.hpp
thomcom: Aww. Done! :D

thomcom:
rerun tests

thomcom:
rerun tests

thomcom:
@gpucibot merge

Issue #599, opened 2022-07-20
[DOC] Improve Python developer documentation


harrism:
We would like to improve cuSpatial's developer documentation to make it easier to contribute to cuSpatial.

I think we should follow the lead of cuDF, which has a relatively comprehensive set of documentation for developers. [cuDF provides](https://github.com/rapidsai/cudf/tree/branch-22.08/docs/cudf/source/developer_guide) a `library_design.md` which provides an overview of the core classes and their relationships. It also provides `documentation.md` which documents how developers should write docstrings for cuSpatial Python APIs.

cuDF is in the process of further improving its Python developer guides, which will provide even better examples for cuSpatial. For example, there are PRs for [creating the main Python developer guide](https://github.com/rapidsai/cudf/pull/11235), [adding developer docs for writing tests,](https://github.com/rapidsai/cudf/pull/11199), and [adding developer docs for benchmarking](https://github.com/rapidsai/cudf/pull/11235). (All by @vyasr)

In addition to developer docs, cuDF also provides a comprehensive [user guide](https://github.com/rapidsai/cudf/tree/branch-22.08/docs/cudf/source/user_guide). This is helpful to developers as well as to users.  We should consider this for cuSpatial as well.

thomcom:
The python User Guide is ready for review in and should close or contribute greatly to this issue. https://github.com/rapidsai/cuspatial/pull/666

PR #610, opened 2022-07-29
Python User's Guide

thomcom:
In process.

Contributes to #599


isVoid:
Would it be better to place the dev docs in rapids documentation along side with API reference for better visibility?

thomcom:
What do you mean?

vyasr:
Michael is asking whether this documentation should be placed inside `docs/source` so that it can be compiled along with the Sphinx documentation and made visible online, as opposed to being put in a file like this one that will only be visible if someone navigates to it on their local filesystem or on GH.

In cudf the C++ documentation is all like this, whereas [the Python developer docs are published](https://docs.rapids.ai/api/cudf/nightly/developer_guide/index.html). FWIW I would eventually like to get all of them published somewhere, but I have not put much thought or effort into making that happen yet. If you all come up with a solution in for cuspatial I would love to hear it.

CC @harrism who I think wrote most of the libcudf docs originally in case he has thought about this before.

harrism:
This is Python docs, so should be in docs/source just like they are in cuDF. (That's what I suggested in #599).

I agree the C++ docs should also be rendered into HTML, but I have never implemented that myself.

thomcom:
Closing in favor of #666.

PR #627, opened 2022-08-03
Make `read_polygon_shapefile` place read-coordinates in the right order.

thomcom:
`read_polygon_shapefile` inserts the coordinates in the opposite order. This PR fixes that issue. I also wrote a pip test in `polygon_shapefile_reader_test.cpp`, thinking that would prove that the ordering mattered.

However, the pip test works in both cases - it does not test for pip based on the "left hand rule" of point-to-linesegment projection. I'm looking at writing a different test to demonstrate the winding.


cpp/benchmarks/pairwise_linestring_distance.cu
```diff
@@ -78,7 +78,7 @@ std::tuple<rmm::device_vector<cartesian_2d<T>>, rmm::device_vector<int32_t>> gen
   std::vector<cartesian_2d<T>> points(num_points);
 
   auto random_walk_func = [segment_length](auto const& prev, auto const& rad) {
-    return cartesian_2d<T>{prev.x + segment_length * rad.x, prev.y + segment_length * rad.y};
+    return cartesian_2d<T>({prev.x + segment_length * rad.x, prev.y + segment_length * rad.y});
```
harrism, line 81: Please don't include this change in this PR. I don't agree that this is the right way to fix this issue, and we haven't even root-caused the issue yet. I have been discussing it with @trxcllnt

cpp/tests/io/shp/polygon_shapefile_reader_test.cpp
```diff
@@ -72,14 +75,37 @@ TEST_F(PolygonShapefileReaderTest, ZeroPolygons) { test("empty_poly.shp", {}, {}
...
+
+  auto ret = cuspatial::point_in_polygon(test_xs,
+                                         test_ys,
+                                         polygons, 
+                                         rings,
+                                         xs,
+                                         ys);
+
+  expect_columns_equivalent(ret->view(), expected);
+}
```
harrism, line 111:
```suggestion
}

```

cpp/src/io/shp/polygon_shapefile_reader.cpp
```diff
@@ -32,8 +32,7 @@ cudf::size_type read_ring(OGRLinearRing const& ring,
 {
   cudf::size_type num_vertices = ring.getNumPoints();
 
-  // append points in reverse order
-  for (cudf::size_type i = num_vertices - 1; i >= 0; i--) {
+  for (cudf::size_type i = 0; i < num_vertices; ++i) {
```
harrism, line 35: I interpreted @zhangjianting 's comment here: https://github.com/rapidsai/cuspatial/pull/609#issuecomment-1204575080 to mean that clockwise vs. counterclockwise is a matter of choice since the OGC and ESRI shapefile formats are opposites. In which case why not either leave this as-is, or provide a parameter to specify which is expected?

harrism (changes requested):

cpp/benchmarks/pairwise_linestring_distance.cu
thomcom: Understood!

cpp/src/io/shp/polygon_shapefile_reader.cpp
thomcom: Providing a parameter might be the way to go here, just in case. In this context, GeoPandas is used to read a shapefile and `read_polygon_shapefile` is used to read a shapefile. While both are valid technically, the coordinates are wound in the opposite direction for the result that is produced by `read_polygon_shapefile` which is why I want to introduce this change.

thomcom:
Abandoned in favor of #609

Issue #563, opened 2022-06-15
[FEA] Refactor trajectory APIs to header-only API


harrism:
Create header-only APIs for the following and refactor existing APIs / tests on top of the header-only API:

- [x] `cuspatial::derive_trajectories` #628 
- [ ] `cuspatial::trajectory_distances_and_speeds`
- [x] `cuspatial::trajectory_bounding_boxes`

See https://github.com/rapidsai/cuspatial/blob/branch-22.08/cpp/include/cuspatial/trajectory.hpp

https://github.com/rapidsai/cuspatial/issues/563#tasklist-block-0ad734e5-e92a-454a-bf8e-72701ce77eb6

jarmak-nv:
@isVoid to help unblock

PR #628, opened 2022-08-04
Header-only refactoring of derive_trajectories

harrism:
Contributes to #563 .


cpp/include/cuspatial/experimental/detail/derive_trajectories.cuh
```diff
@@ -0,0 +1,149 @@
...
+}  // namespace detail
+
+template <typename IdInputIt,
+          typename PointInputIt,
+          typename TimestampInputIt,
+          typename IdOutputIt,
+          typename PointOutputIt,
+          typename TimestampOutputIt,
+          typename OffsetType>
+std::unique_ptr<rmm::device_vector<OffsetType>> derive_trajectories(
```
harrism, line 103: Let's standardize on returning `device_uvector` when we need to allocate output vectors.

cpp/include/cuspatial/experimental/detail/derive_trajectories.cuh
```diff
@@ -0,0 +1,149 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "cub/device/device_merge_sort.cuh"
+#include "cub/device/device_radix_sort.cuh"
+#include "cub/device/device_segmented_sort.cuh"
```
trxcllnt, line 21: Also move these to the bottom under Thrust:

```suggestion
#include <cub/device/device_merge_sort.cuh>
#include <cub/device/device_radix_sort.cuh>
#include <cub/device/device_segmented_sort.cuh>
```

cpp/include/cuspatial/experimental/detail/derive_trajectories.cuh
```diff
@@ -0,0 +1,149 @@
...
+{
+  using id_type        = typename std::iterator_traits<IdInputIt>::value_type;
+  using timestamp_type = typename std::iterator_traits<TimestampInputIt>::value_type;
+  using tuple_type     = thrust::tuple<id_type, timestamp_type>;
+
+  auto key_comparator = [] __device__(tuple_type const& lhs, tuple_type const& rhs) {
+    auto lhs_id = thrust::get<0>(lhs);
+    auto rhs_id = thrust::get<0>(rhs);
+    auto lhs_ts = thrust::get<1>(lhs);
+    auto rhs_ts = thrust::get<1>(rhs);
```
trxcllnt, line 64: I was curious the difference between this code as written, and the suggestion here, so I compared them in godbolt:

* current: https://godbolt.org/z/z3G3qrMa6
* with refs: https://godbolt.org/z/EWbjTYYc6

They're quite different! I'm not exactly sure what's going on, but it seems like the second one compiles down to a series of chained branches? Am I reading this right?

```suggestion
    auto& lhs_id = thrust::get<0>(lhs);
    auto& rhs_id = thrust::get<0>(rhs);
    auto& lhs_ts = thrust::get<1>(lhs);
    auto& rhs_ts = thrust::get<1>(rhs);
```

harrism: I don't see any reason to change this to references. In the case of `float` coordinates, a reference may actually be a bigger value (64-bit pointer vs 32-bit float)?  In the case of `double` coordinates, the reference is still no smaller.

trxcllnt: Yeah that's the sort of thing I was trying to check in godbolt. Didn't mean to suggest changing it, just using suggestion here to get a nice diff output. I've used refs to `thrust::get` vars in kernels before, but never been sure if it made any difference to performance.

harrism: I imagine it *could* help in some cases.

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -0,0 +1,190 @@
...
+
+using time_point = std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds>;
+
+/* Test data generation for trajectory APIs. Generates num_trajectories trajectories of random
+   size between 1 and max_trajectory_size points. Creates both a reference (sorted) set of
+   trajectory IDs, timestamps, and points, and a shuffled (unsorted) set of the same for test
+   input. The sorted data can be input directly into `trajectory_distance_and_speed`, while
+   the shuffled data can be input to `derive_trajectories`.
+
+   The times are not random, but do vary somewhat between trajectories and trajectory lenghts.
```
isVoid, line 47:
```suggestion
   The times are not random, but do vary somewhat between trajectories and trajectory lengths.
```

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -0,0 +1,190 @@
...
+                      time_step * std::chrono::milliseconds(100) +
+                      std::chrono::milliseconds(int(10 * cos(time_step)));
+      return time_point{duration};
+    }
+  };
+
+  struct point_functor {
+    __device__ cuspatial::vec_2d<T> operator()(time_point const& time, std::int32_t id)
+    {
+      // X is time in seconds, Y is cosine(time), offset by ID
```
isVoid, line 111: I think `count` only counts the ticks of the specified time unit: https://en.cppreference.com/w/cpp/chrono/duration/count
```suggestion
      // X is time in milliseconds, Y is cosine(time), offset by ID
```

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
isVoid: Perhaps provide a high level insight to how the time array is generated. I gave a shot myself:
```
Each trajectory has a distinct starting time point and increases monotonically for each samples. The interval between samples varies within a small range.
```

isVoid (approved):
To me, I don't see a fundamental difference between `trajectories_utilities.cuh` and `trajectory_test_utils.cuh`. After we do more refactoring we might be able to replace `trajectories_utilities.cuh` completely.

harrism:
> To me, I don't see a fundamental difference between `trajectories_utilities.cuh` and `trajectory_test_utils.cuh`. After we do more refactoring we might be able to replace `trajectories_utilities.cuh` completely.

Yes, my idea is to eliminate it. It is still used by `trajectory_distance_and_speed` and `trajectory_bounding_boxes` which will be refactored in separate PRs.

cpp/include/cuspatial/experimental/detail/derive_trajectories.cuh
harrism: Done.

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -0,0 +1,192 @@
...
+  rmm::device_vector<cuspatial::vec_2d<T>> points;
+
+  rmm::device_vector<std::int32_t> ids_sorted;
+  rmm::device_vector<time_point> times_sorted;
+  rmm::device_vector<cuspatial::vec_2d<T>> points_sorted;
+
+  trajectory_test_data(std::size_t num_trajectories, std::size_t max_trajectory_size)
+  {
+    make_data(num_trajectories, max_trajectory_size);
+  }
```
thomcom, line 68: This function doesn't seem to contribute anything except to add one more level of indirection?

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -0,0 +1,192 @@
...
+    {
+    }
+    __device__ std::int32_t operator()(int i)
+    {
+      gen.discard(i);
+      return size_rand(gen);
+    }
+  };
+
+  void make_data(std::size_t num_trajectories, std::size_t max_trajectory_size)
```
thomcom, line 135:
```suggestion
  void trajectory_test_data(std::size_t num_trajectories, std::size_t max_trajectory_size)
```

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
thomcom, line 68:
```suggestion
```

cpp/tests/trajectory/test_derive_trajectories.cpp
```diff
@@ -0,0 +1,108 @@
...
+  }
+  {
+    auto id = cudf::column(rmm::device_uvector<int>(size, rmm::cuda_stream_default));
+    auto xs = cudf::column(rmm::device_uvector<float>(size, rmm::cuda_stream_default));
+    auto ys = cudf::column(rmm::device_uvector<float>(size, rmm::cuda_stream_default));
+    auto ts =
+      cudf::column(rmm::device_uvector<cudf::timestamp_ms>(size / 2, rmm::cuda_stream_default));
+    EXPECT_THROW(cuspatial::derive_trajectories(id, xs, ys, ts, this->mr()),
+                 cuspatial::logic_error);
+  }
```
thomcom, line 63: Judging from these tests, I missed the spot above where you are in fact checking that all lengths are the same?

thomcom (approved):

cpp/include/cuspatial/experimental/detail/derive_trajectories.cuh
```diff
@@ -0,0 +1,156 @@
...
+                             ids_last,
+                             points_first,
+                             timestamps_first,
+                             ids_out_first,
+                             points_out_first,
+                             timestamps_out_first,
+                             stream,
+                             mr);
+
+  auto const num_points = std::distance(ids_first, ids_last);
```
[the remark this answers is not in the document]
harrism: That's not possible -- we only get a `last` iterator for ids.  Assuming (not asserting) the iterator ranges are equal is standard practice in iterator-based APIs like this (and Thrust, and STL).

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
harrism: This is a constructor for the class.  I generally don't like putting a ton of code inside a constructor, and `make_data` is a long function.

cpp/tests/trajectory/test_derive_trajectories.cpp
harrism: We check in the column-based API, but we can't check in the header-only API.

harrism: Columns have a defined size. Iterators don't.

cpp/include/cuspatial/experimental/detail/derive_trajectories.cuh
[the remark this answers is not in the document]
harrism: Also, the point of an iterator-based API is flexibility. The caller is free to pass anything that looks like an iterator -- it might be, for example, a counting iterator, or some other dynamic iterator type that generates an open-ended sequence. There is no end. :)

harrism:
rerun tests

harrism:
@gpucibot merge

Issue #566, opened 2022-06-15
[FEA] refactor point quadtree API to header-only


harrism:
Create header-only APIs for the following and refactor existing APIs / tests on top of the header-only API:

* `cuspatial::quadtree_on_points()`

See https://github.com/rapidsai/cuspatial/blob/branch-22.08/cpp/include/cuspatial/point_quadtree.hpp

Issue #397, opened 2021-05-06
[DOC/QST] Scale parameter in quadtree methods


buckeye17:
The API documentation describes the `scale` parameter as, "Scale to apply to each point’s distance from (x_min, y_min)."  While this is comprehensible, I'm struggling to understand why someone would want to change it.  Is it so that the location data can be kept in coordinate reference frame, while outputs for trajectories, etc. can be given in another?  I've looked over the demos and examples and can't find any explanation for why this feature is useful, and I'm not sure how it affects the quadtree results, if at all.  Thanks for any clarification you can provide!

zhangjianting:
quadtrees work on a discrete space. scale, together with x_min and y_min, defines the configuration of the grid that a quadtree works on. The smaller the scale, the more rows/columns of the grid and typically the larger and the deeper of the resulting quadtree. That being said, the tradeoff between filtering and refinement (for spatial join and other query types) when utilizing quadtree indexing does not have a definite relationship with scale (or grid resolution). 
Ideally, a "smarter" indexing technique can determine scale automatically but it is still pretty much an open research question. As such, cuspatial allows users to specify scale for better control. It is understandable that some background expertise is needed. Hopefully the slides at following link can provide some additional info https://github.com/zhangjianting/cuspatial/blob/fea-spatial-join/docs/quadtree_indexing_spatial_join.pdf

harrism:
We should add more of an explanation as provided by @zhangjianting to the documentation for this function.

harrism:
@trxcllnt can you add the explanation as part of your refactoring?

harrism:
We need to update the documentation here: https://github.com/rapidsai/cuspatial/blob/930a2f33953ce1abe8f0680ebbf418675db4b362/python/cuspatial/cuspatial/core/spatial/indexing.py#L35

We should also provide default values for scale, max_depth.

Also the Python API for min_size needs to be updated to match max_size in the C++ API.

Discussion:
* Scale is not a great name, since we divide by it. Almost a normalization factor?

harrism:
@trxcllnt can we also consider this issue fixed by #797 ?

PR #639, opened 2022-08-11
Add header only `cuspatial::quadtree_on_points`

trxcllnt:
Closes https://github.com/rapidsai/cuspatial/issues/566.


cpp/tests/experimental/indexing/point_quadtree_test.cu
```diff
@@ -0,0 +1,253 @@
...
+  thrust::device_vector<T> dvec(uvec.size());
+  thrust::fill(dvec.begin(), dvec.end(), T{0});
+  thrust::copy(rmm::exec_policy(uvec.stream()), uvec.begin(), uvec.end(), dvec.begin());
+
+  std::vector<T> hvec(dvec.size());
+  std::fill(hvec.begin(), hvec.end(), T{0});
+  thrust::copy(dvec.begin(), dvec.end(), hvec.begin());
+
+  return hvec;
+}
```
trxcllnt, line 41: For some reason all the other ways I tried copying a `device_uvector` to host kept segfault'ing. I'd love to know if there's a better way to do this.

harrism: CudaMemcpyAsync, unfortunately.

harrism: Those fill calls are redundant.

trxcllnt: Updated

harrism (changes requested):
Good work so far, this is a big task. I didn't do a complete review. I will re-review after you go over these comments.

cpp/include/cuspatial/experimental/point_quadtree.cuh
```diff
@@ -0,0 +1,123 @@
...
+   */
+  rmm::device_uvector<uint32_t> key;
+  /**
+   * @brief UINT8 column of quadtree levels
+   */
+  rmm::device_uvector<uint8_t> level;
+  /**
+   * @brief BOOL8 column indicating whether the node is a quad (true) or leaf (false)
+   */
+  rmm::device_uvector<uint8_t> is_quad;
```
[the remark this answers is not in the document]
trxcllnt: I did at first, but that makes the DtoH copy to a `std::vector<bool>` hard (in the tests) because of how vector packs bools into bits.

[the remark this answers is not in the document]
harrism: Yeah, in libcudf we avoid this by always using `thrust::host_vector` when we have vectors-of-bool in our tests.  But if you use the enum class you have `vector<node_type>` in your tests which won't be a `vector<bool>`.

cpp/include/cuspatial/experimental/point_quadtree.cuh
```diff
@@ -0,0 +1,123 @@
...
+ * @param mr The optional resource to use for output device memory allocations.
+ *
+ * @throw cuspatial::logic_error If the x and y column sizes are different
+ * @throw cuspatial::logic_error If scale is less than or equal to 0
+ * @throw cuspatial::logic_error If min_size is less than or equal to 0
+ * @throw cuspatial::logic_error If x_min is greater than x_max
+ * @throw cuspatial::logic_error If y_min is greater than y_max
+ * @throw cuspatial::logic_error If max_depth is less than 0 or greater than 15
+ *
+ * @return Pair of UINT32 column of sorted keys to point indices and a complete point_quadtree
```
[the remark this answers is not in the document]
trxcllnt: I think we're using `uint32_t` because the z-order curve does [this](https://github.com/trxcllnt/cuspatial/blob/dda0b7e7102d04999eb9b5a03d97cb08b5d8638f/cpp/include/cuspatial/detail/utility/z_order.cuh#L74-L77) fancy bit-shifted-lookup-table trick to make the [sorting](https://github.com/trxcllnt/cuspatial/blob/dda0b7e7102d04999eb9b5a03d97cb08b5d8638f/cpp/include/cuspatial/experimental/detail/indexing/construction/phase_1.cuh#L83-L84) fast. If we use a signed type, the overflow could cause two distinct points to have the same key and produce incorrect results.

cpp/include/cuspatial/experimental/point_quadtree.cuh
```diff
@@ -0,0 +1,123 @@
...
+ * @brief Construct a quadtree structure from points.
+ *
+ * @see http://www.adms-conf.org/2019-camera-ready/zhang_adms19.pdf for details.
+ *
+ * @note `scale` is applied to (x - x_min) and (y - y_min) to convert coordinates into a Morton code
+ * in 2D space.
+ * @note `max_depth` should be less than 16, since Morton codes are represented as `uint32_t`. The
+ * eventual number of levels may be less than `max_depth` if the number of points is small or
+ * `min_size` is large.
+ * @note All quadtree nodes should have fewer than `min_size` number of points except leaf
```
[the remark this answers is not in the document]
trxcllnt: You're right -- reading through the code, I think this is the max number of points allowed in a node before the node is subdivided into 4 new leaves.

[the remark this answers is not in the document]
zhangjianting: @harrism @trxcllnt It is the "min_size" to qualify as a non-leaf node, which should be the same as the "max_size" as a leaf-node. Just to want to add that the numbers of nodes for the last level nodes (which need to be leaf nodes) is not limited by the threshold. When min_size/and or max_depth is small, many (could be most) last-level nodes will have large numbers of points under them.

cpp/include/cuspatial/experimental/detail/indexing/construction/phase_1.cuh
```diff
@@ -58,43 +55,41 @@ namespace detail {
...
+  rmm::device_uvector<uint32_t> keys(num_points, stream);
+  thrust::transform(
+    rmm::exec_policy(stream),
+    points_first,
+    points_last,
+    keys.begin(),
+    [=] __device__(vec_2d<T> const& point) {
+      // auto const& x = point.x;
+      // auto const& y = point.y;
+      if (point.y < min.x || point.y > max.x || point.y < min.y || point.y > max.y) {
```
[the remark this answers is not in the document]
harrism: Since this is likely to be a common operation (already exists here: https://github.com/rapidsai/cuspatial/blob/cd0f2b3ef9272839efcdbef489fbd93bd445d402/cpp/include/cuspatial/experimental/detail/points_in_range.cuh#L46), perhaps it should be factored out to minimize mistakes like this.

cpp/include/cuspatial/experimental/detail/indexing/construction/phase_1.cuh
```diff
@@ -58,43 +55,41 @@ namespace detail {
...
+        // If the point is outside the bbox, return a max_level key
+        return static_cast<uint32_t>((1 << (2 * max_depth)) - 1);
+      }
+      return cuspatial::detail::utility::z_order((point.x - min.x) / scale,
```
harrism, line 83: Hmmm, is this an implicit cast from float/double to int16_t? I never noticed this before. Is this safe?

cpp/include/cuspatial/experimental/point_quadtree.cuh
```diff
@@ -0,0 +1,100 @@
...
+  // uint32_t vector for the first child position (if is_parent_node), or first point position
+  rmm::device_uvector<uint32_t> offset;
+};
+
+/**
+ * @brief Construct a quadtree structure from points.
+ *
+ * @see http://www.adms-conf.org/2019-camera-ready/zhang_adms19.pdf for details.
+ *
+ * @note `scale` is applied to (x - min.x) and (y - min.y) to convert coordinates into a Morton code
```
harrism, line 56: "applied" doesn't reflect that those differences are divided by scale...  If there is an intuitive explanation, let's add that. And if not, then put the exact math in here. See https://github.com/rapidsai/cuspatial/issues/397

harrism (changes requested):
Just a couple more things.

cpp/include/cuspatial/experimental/detail/indexing/construction/phase_1.cuh
trxcllnt: It is an implicit cast, but I'll update it to be explicit.

cpp/include/cuspatial/experimental/point_quadtree.cuh
harrism: As discussed, should probably use a reciprocal of the current scale to make the scale factor actually a directly proportional scale factor (which would be more intuitive).

harrism: Further discussion: let's save any semantic / breaking change for a later PR if necessary.

harrism (approved):

cpp/tests/experimental/indexing/point_quadtree_test.cu
```diff
@@ -0,0 +1,221 @@
...
+    auto level_h          = device_uvector_to_host<uint8_t>(level_d);
+    auto is_parent_node_h = device_uvector_to_host<bool>(is_parent_node_d);
+    auto length_h         = device_uvector_to_host<uint32_t>(length_d);
+    auto offset_h         = device_uvector_to_host<uint32_t>(offset_d);
+
+    EXPECT_THAT(key_h, ::testing::Pointwise(::testing::Eq(), expected_key));
+    EXPECT_THAT(level_h, ::testing::Pointwise(::testing::Eq(), expected_level));
+    EXPECT_THAT(is_parent_node_h, ::testing::Pointwise(::testing::Eq(), expected_is_parent_node));
+    EXPECT_THAT(length_h, ::testing::Pointwise(::testing::Eq(), expected_length));
+    EXPECT_THAT(offset_h, ::testing::Pointwise(::testing::Eq(), expected_offset));
```
isVoid, line 89: Using the new `expect_vector_equivalent` should simplify this. Although we might need to extend it to support integers.

isVoid (commented):
Sorry about the late review. I delved into the reference in depth but still got lost in the `remove_unqualified_quadtree`. According to this comment: https://github.com/rapidsai/cuspatial/pull/639#discussion_r958540554. It feels like `max_size` still should be called `min_size`, otherwise it makes `remove_unqualified_quads` strange to read. We should give it a more verbose name e.g. `min_non_leaf_node_size`. 

@zhangjianting Is it possible to have a leaf node that's not on the last-level? aka is quadtree full?

Not a lot of code smell in terms of the header only refactoring except for a few small issue (IIFE and structure binding, not all are labeled).

isVoid (approved):

harrism:
@gpuccibot merge

harrism:
@gpucibot merge

Issue #644, opened 2022-08-15
Proposed New Layout of the Python Library


isVoid:
## Motivation
Currently cuspatial's file structure is mostly flat, with two modules whose name have overlapping meanings: `core` and `geometry`. We should refactor the structure of the library so that the physical layout of the folder structure corresponds to the extent of the library, as described in the leading section of [docs](https://github.com/rapidsai/cuspatial/blob/branch-22.10/docs/source/index.md). This gives developers a clearer idea what's within the problem boundary of cuspatial, and where to dedicate their contributions to.

## Library Design
In general, cuspatial separates compute API and geodataframe APIs. The compute APIs are broadly organized by the data type it operates on (Spatial or Trajectory) and the computation type (generic, indexing, join, filter). We should avoid using general names for the module like `generic` etc. Developers should strive to find the corresponding subcategory of the compute API, or create a new ones should it does not exist.

### Functional v.s. OOP design
cuspatial endorses both functional and OOP design of APIs. `GeoSeries` and `GeoDataframe` is the core object the user will interact their geometric/geodesic data with. These APIs will strive to achieve function parity with `geopandas`. Beyond that, functional APIs supports a superset of functions that cuspatial supports, which may require to interact with custom objects within its own problem domain, such as `CubicSpline` and quad tree indexing table.

### Internal v.s. external
Column level implementation and geoarrow specifications of `geoseries` is internal API.

## Breaking v.s. Non-breaking
Most refactor that goes on here should be `non-breaking` change as this only affects internal library design. If desired, we may enforce the same user view of cuspatial APIs as the library design in the future.

The below are the proposed new layout:

- cuspatial
  - _lib/ (for cython files, untouched in this refactor)
  - core/ (for all python APIs)
    - spatial/
      - distance
      - projection
      - bounding
      - indexing
      - join
      - filtering
    - _column/
      - geocolumn.py
      - pygeoarrow.py
      - geometa.py
      - geoutil.py
    - geoseries.py
    - geodataframe.py
    - trajectory.py
    - interpolate.py
  - io/
  - tests/
  - utils/

harrism:
I don't like "generic". What are the functions that fall within generic? Perhaps we can come up with some categories. 

Why doesn't each function get it's own .py file like we do in C++? If that's just "Pythonic", let me know.

isVoid:
[...]

PR #645, opened 2022-08-15
Decouple `interpolate` functions from trajectory

isVoid:
## Description
closes https://github.com/rapidsai/cuspatial/issues/644

This PR renames `t` argument of `CubicSpline` to indicate that it can be used as general CubicSpline interpolation function.
Other minor documentation refresh are also included.

## Checklist


python/cuspatial/cuspatial/core/trajectory/generic.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019, NVIDIA CORPORATION.
+# Copyright (c) 2019-2022, NVIDIA CORPORATION.
```
harrism, line 1: I don't like naming this "generic.py". Why can't you leave it as `trajectory.py`?

harrism (changes requested):

isVoid:
As discussed in weekly, we maintain `trajectory` module and `interpolate` module. The `interpolate` module is refactored to be generic interpolation functions. Specifically, argument `t` is renamed as `x` for general independent variable.

isVoid:
During rewriting the docs, I was trying to run micro benchmarks to find out about the performance cut-off point and raise warnings for users about a "too-small input". But the result is a bit confusing as the cut off point seems to shift by the ratio of the number of curves and the complexity of the curve. See benchmark: https://gist.github.com/isVoid/108d388173e7199758e31c1ce7f6842a

Perhaps @thomcom can help explain?

thomcom:
Hey @isVoid I'm not sure what you mean by the cut-off point?

isVoid:
@thomcom currently, in `CubicSpline` code we intentionally to avoid API parity with scipy because we want to avoid user calling the API when the input data size is small: https://github.com/rapidsai/cuspatial/blob/19a8d3fada453f411d1b29a6d9449bc1c0e0c2fe/python/cuspatial/cuspatial/core/interpolate.py#L45-L47

This isn't helpful when user still calls the API with a small data size without aware of the performance implication. In this refactor I attempt to argue that the proper way to design the API is to use the same argument variable as scipy and raise a `UserWarning` when input size is small. This is similar to how cudf `groupby.apply` handles many group cases. The key question here is: how to determine when the input size is too small? There must exist some data size when scipy is more performant than cuspatial, and that data size is the "cut-off" point I mentioned above.

thomcom:
Ah, I see. I just ran a few quick benchmarks and I see that scipy is 15x faster than cuspatial on interpolations on my HP Z8 workstation. cuspatial only parallelizes each individual spline fit, so I estimate that if you've got 15 trajectories, cuspatial will be faster at interpolation. I was able to verify this with a set of 15 trajectories of length 1m. Creating the curve object still takes 15x as long (which is basically just creating 15 interpolations in parallel, each taking 15x as long as the hcurve) but interpolating all 15m samples takes 8ms, interpolating 15m samples in scikit learn takes 230ms.

isVoid:
> a set of 15 trajectories of length 1m

Do you mean 15 curves, each curve has 1,000,000 vertices?

thomcom:
> > a set of 15 trajectories of length 1m
> 
> Do you mean 15 curves, each curve has 1,000,000 vertices?

Yes

> Do you mean when cuspatial interpolates 15 curves in parallel, the total time of the kernel is 15x as long as scipy? What's the order of time cost here, `s`? `ms`?

I think that when cuspatial interpolates 15 curves in parallel the cost is similar to using scipy to compute them in serial.

> Sounds like cuspatial has advantage when sampling a complex curve (many-vertices curve) and does not have advantage computing the curve parameters when the total number of curve is small? I think we can raise a performance warning when we see small number of curves (<100) is constructed, do you agree?

I was thinking that < 15 curves was appropriate to compute the warning. I'll write a notebook to concrete the benchmark.

> Out of scope, but can you fit the curve with scipy and pass the parameter to device and sample with device? If you only have 15 curves, you only have 15x4 parameters.

This would be theoretically possible but our curve fitting equations are not identical to scipys, iirc.

isVoid:
rerun tests

python/cuspatial/cuspatial/core/interpolate.py
```diff
@@ -36,87 +36,82 @@ class CubicSpline:
...
+    -------
+    CubicSpline : callable `o`
+        ``o.c`` contains the coefficients that can be used to compute new
+        points along the spline fitting the original ``t`` data. ``o(n)``
+        interpolates the spline coordinates along new input values ``n``.
+
+    Note
+    ----
+    cuSpatial will outperform scipy when many splines are
     fit simultaneously. Data must be arranged in a SoA format, and the
```
harrism, line 70: SoA may be unclear to some readers...

harrism (approved):

python/cuspatial/cuspatial/core/interpolate.py
isVoid: Added full name for readability.

python/cuspatial/cuspatial/core/interpolate.py
```diff
@@ -67,9 +67,9 @@ class CubicSpline:
     Note
     ----
     cuSpatial will outperform scipy when many splines are
-    fit simultaneously. Data must be arranged in a SoA format, and the
-    exclusive `prefix_sum` of the separate curves must also be passed to the
-    function. See example for detail.
+    fit simultaneously. Data must be arranged in a structure of arrays (SoA)
+    format, and the exclusive `prefix_sum` of the separate curves must also be
```
harrism, line 72: These are the offsets to the . Don't call it a prefix sum, call it offsets. Prefix sum would be one possible way of computing these offsets, and is an implementation detail. The example below does not compute a prefix sum, it just names a variable `prefix_sum`. The example uses fixed-length splines so the offsets are all just multiples of that length.

Apologies, I know this example has been around for a while, but I think it is not very self-explanatory. It is made up of a bunch of fairly opaque numpy calls to generate data. If instead the example used trivial length splines (e.g. 4 or 5 points), then the data could be explicitly written out and the example would be easier to follow.

harrism (changes requested):

python/cuspatial/cuspatial/core/interpolate.py
isVoid: Fair enough, I updated the example to make it clearer.

python/cuspatial/cuspatial/core/interpolate.py
```diff
@@ -69,33 +69,35 @@ class CubicSpline:
...
-                0, (SPLINE_LENGTH - 1), (SPLINE_LENGTH - 1) * 2 + 1
-            ),) * NUM_SPLINES)
-        ).astype('float32')
-    >>> curve_ids = cudf.Series(np.repeat(
-            np.arange(0, NUM_SPLINES), SPLINE_LENGTH * 2 - 1
-        ), dtype="int32")
-    >>> new_points = curve(new_samples, curve_ids)
+    # The following example only serves as demonstration purposes of
+    # cuspatial.CubicSpline API. cuSpatial does not out perform scipy when
+    # fitting a small number of curves shown in the example below.
```
harrism, line 79:
```suggestion
    # The following example only serves to demonstrate the
    # cuspatial.CubicSpline API. cuSpatial does not out perform scipy when
    # fitting a small number of curves as shown.
```

harrism (changes requested):

thomcom (approved):
lgtm

isVoid:
close #525

harrism (approved):

thomcom:
@gpucibot merge

PR #657, opened 2022-08-19
Move `GeoSeries` `GeoDataframe` into `core` package and hide column implementation in internal `_column` package

isVoid:
This PR moves geoseries and geodataframe into `core` package and moves `geocolumn`, `geometa` into internal `_column` package as they are considered non-public interface. `pygeoarrow` is moved to `io` package as it's obviously about interoperability.

Minor changes includes removing `geoutils` module and moves the helper function into geodataframe since it's only used in one module.

Contributes to #644 #599 #525


isVoid:
rerun tests

.pre-commit-config.yaml
```diff
@@ -13,6 +13,7 @@ repos:
         hooks:
               - id: black
                 files: python/cuspatial/.*
+                args: ["--config", "pyproject.toml"]
```
thomcom, line 16: I see you're using black as a pre-commit hook. How do you enable this? Is it on by default and I haven't noticed?

thomcom (approved):
Great work improving this factorization. In particular, I have always called I/O In/Out and InterOperability is new to me. I like it!

.pre-commit-config.yaml
isVoid: You can either run `pre-commit` to manually run it prior committing or let git run it automatically!

https://pre-commit.com/#3-install-the-git-hook-scripts

isVoid: The change here is to make black's config file available to all subpackages in the repo. Unlike cudf (before https://github.com/rapidsai/cudf/pull/11258), `cuspatial-python/pyproject.toml` doesn't define black configs, so when you try to run `python -m black` in the subpackage, it may disregard the config file at root level. This is a minor bug fix unrelated to the refactor.

thomcom:
@gpucibot merge

Issue #646, opened 2022-08-15
[FEA] Nearest point and segment index of linestring to a point (pairwise)


harrism:
**Is your feature request related to a problem? Please describe.**

For linestring/point pairs, I would like to calculate not only the distance between points and linestrings, but also the nearest point on the linestring and possibly the index of the segment within the linestring that contains that nearest point.

**Describe the solution you'd like**

Ingestivate leveraging the functionality in #573 to provide the nearest point and segment index.

isVoid:
To clarify, if a point is the same as a vertex, what's the result segment idx?

isVoid:
Another clarification: does this feature include multi-linestring?

harrism:
For the current request I haven't heard that this is needed. However for completeness I guess we should. The question is what do you return if so?

isVoid:
My prototype returns a third column that provides the "part index" of the linestring in a multi-linestring.

isVoid:
A third clarification: in case there are multiple segments having equal distance to the point, how do you break the tie?

harrism:
If we are only returning one, I would suggest to not guarantee any ordering. This gives the most flexibility in a parallel implementation.

Issue #649, opened 2022-08-16
[FEA] Assume GeoArrow Format in C++


isVoid:
Currently, `offset` arguments for most c++ APIs are in ESRI shapefile polygon format (see [shapefile whitepaper](https://www.esri.com/content/dam/esrisites/sitecore-archive/Files/Pdfs/library/whitepapers/pdfs/shapefile.pdf), page 10, table 7). The offsets array length equals the number of underlying components. Geoarrow contains one more element in the offset array, which is the length of the underlying component. Adopting geoarrow format in c++ will enable https://github.com/rapidsai/cuspatial/issues/637, and is cleaner when iterating over pairs of components in kernels.

harrism:
We should probably add a checklists of all functions that need to be updated for this...

isVoid:
Checklist or separate issues?

harrism:
Seems like that would be a LOT of issues...

harrism:
This should be a milestone.

PR #658, opened 2022-08-20
Add C++ API for `point_linestring_nearest_points`

isVoid:
This PR adds C++ API of `point_linestring_nearest_points`. The host counterpart of this function is `shapely.ops.nearest_points`. Multiple fields are written when called on this API, including the geometry and part id indicating the geometry that the nearest point is on. In the header only API, the output iterator accepts a zipped iterator made of 4 output iterators. In the column API, a tuple is returned.

The kernel is parallelized per geometry. Each thread computes the nearest point between a pair of multipoint and multilinestring.

Contribute to #646


cpp/src/spatial/point_linestring_nearest_point.cu
```diff
@@ -0,0 +1,290 @@
...
+#include <utility>
+
+namespace cuspatial {
+
+namespace detail {
+
+namespace {
+
+template <bool is_multi_point, bool is_multi_linestring>
+struct launch_functor {
```
isVoid, line 46: Rename this functor

cpp/include/cuspatial/point_linestring_nearest_point.hpp
```diff
@@ -0,0 +1,148 @@
...
+ * @param linestring_part_offsets Beginning and ending indices for each linestring
+ * @param linestring_points_xy Interleaved x, y-coordinates of the linestring points
+ * @param mr Device memory resource used to allocate the returned column.
+ * @return A tuple containing the following:
+ * 1. The point id of the nearest point in multipoint (std::nullopt if input is not multipoint)
+ * 2. The linestring id where the nearest point in multilinestring locates (std::nullopt if input is
+ * not multlinestring)
+ * 3. The segment id where the nearest point in the (multi)linestring locates
+ * 4. The interleaved x, y-coordinate of the nearest point on the (multi)linestring
+ *
```
isVoid, line 134: Write throws

cpp/include/cuspatial/experimental/point_linestring_nearest_point.cuh
```diff
@@ -0,0 +1,69 @@
...
+ * @param stream
+ * @return
+ */
+template <class Vec2dItA,
+          class Vec2dItB,
+          class OffsetIteratorA,
+          class OffsetIteratorB,
+          class OffsetIteratorC,
+          class OutputIt>
+OutputIt pairwise_point_linestring_nearest_point(
```
isVoid, line 55: doc

cpp/include/cuspatial/experimental/detail/point_linestring_nearest_point.cuh
```diff
@@ -0,0 +1,166 @@
...
+#include <iterator>
+#include <limits>
+#include <memory>
+#include <type_traits>
+
+namespace cuspatial {
+namespace detail {
+
+/**
+ * @internal
```
isVoid, line 39: doc

cpp/tests/utility/vector_equality.hpp
```diff
@@ -16,26 +16,62 @@
...
+  if constexpr (std::is_same_v<T, float>) {
+    return AllOf(Field(&Vec2d::x, FloatEq(v.x)), Field(&Vec2d::y, FloatEq(v.y)));
   } else {
-    EXPECT_THAT(lhs, ::testing::Pointwise(::testing::DoubleEq(), rhs));
+    return AllOf(Field(&Vec2d::x, DoubleEq(v.x)), Field(&Vec2d::y, DoubleEq(v.y)));
   }
 }
 
+template <typename Vector1, typename Vector2>
+inline void expect_vec2d_vector_equivalent(Vector1 lhs, Vector2 rhs)
```
isVoid, line 57: Will be removed after https://github.com/rapidsai/cuspatial/pull/669

isVoid:
> Hi, I noticed that cython does not yet support `std::tuple` due to lack of variadic template support. Therefore `pairwise_point_linestring_nearest_points` cannot return a tuple of the results. We'll need to construct a custom struct for the return type. Working on a change.

done.

isVoid:
rerun tests

cpp/include/cuspatial/experimental/detail/point_linestring_nearest_points.cuh
```diff
@@ -0,0 +1,172 @@
...
+                                                OutputIt output_first)
+{
+  using T        = iterator_vec_base_type<Vec2dItA>;
+  using SizeType = iterator_value_type<OffsetIteratorA>;
+
+  auto num_pairs =
+    thrust::distance(points_geometry_offsets_first, points_geometry_offsets_last) - 1;
+  auto num_linestring_points = thrust::distance(linestring_points_first, linestring_points_last);
+
+  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < num_pairs;
```
harrism, line 75: Seeing this quadruple nested loop, I see now how important it will be to add automatic geometry iterators that know how to do the nested indexing. You want to be able to just iterate over the pairs, and for each pair, just have a for_each over the points and a nested for_each over the points of the multilinestring.

cpp/include/cuspatial/experimental/point_linestring_nearest_points.cuh
```diff
@@ -0,0 +1,103 @@
...
+ * stores the geometry id of the nearest linestring. The third element should be compatible with
+ * iterator_value_type<OffsetIteratorC>, stores the part id to the nearest segment. The forth
+ * element should be compatible with vec_2d, stores the coordinate of the nearest point on the
+ * (multi)linestring.
+ * @param stream The CUDA stream to use for device memory operations and kernel launches.
+ * @return Output iterator to the element past the last tuple computed.
+ *
+ * @pre all input iterators for coordinates must have `cuspatial::vec_2d` type.
+ * @pre all scalar types must be floating point types, and must be the same type for all input
+ * iterators and output iterators.
```
harrism, line 78: What floating point scalar types are there? Is this precondition necessary?

cpp/include/cuspatial/experimental/point_linestring_nearest_points.cuh
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * The nearest point from a test multipoint to a multilinestring is the nearest point that
+ * has the shortest distance in all pairs of points and linestrings.
+ *
+ * In addition, this API writes these geometry and part ID where the nearest point locates to output
+ * iterators:
+ * - The point id indicates which point in the multipoint is the nearest point.
+ * - The linestring id is the intra-offset to the linestring that nearest point locates
+ * - The segment id is the intra-offset to the segment that nearest point locates. It is
+ *   the same as the id to the starting point of the segment.
```
harrism, line 38:
```suggestion
 * - The linestring ID is the offset within the multilinestring of the linestring that contains the nearest point.
 * - The segment ID is the offset within the linestring of the segment that contains the nearest point. It is
 *   the same as the ID of the starting point of the segment.
```

cpp/include/cuspatial/point_linestring_nearest_points.hpp
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @var point_linestring_nearest_points_result::nearest_point_on_linestring_xy
+ * The interleaved x, y-coordinate of the nearest point on the (multi)linestring
+ */
+struct point_linestring_nearest_points_result {
+  std::optional<std::unique_ptr<cudf::column>> nearest_point_geometry_id;
+  std::optional<std::unique_ptr<cudf::column>> nearest_linestring_geometry_id;
+  std::unique_ptr<cudf::column> nearest_segment_id;
+  std::unique_ptr<cudf::column> nearest_point_on_linestring_xy;
+};
```
harrism, line 52: Why not document the members of the struct inline and eliminate the @struct, @var, etc?

```suggestion
struct point_linestring_nearest_points_result {
  /// The point id indicating which point in the multipoint is the nearest point
  /// (`std::nullopt` if input is not a multipoint array)
  std::optional<std::unique_ptr<cudf::column>> nearest_point_geometry_id;
  
  /// etc.
  std::optional<std::unique_ptr<cudf::column>> nearest_linestring_geometry_id;
  
  /// etc.
  std::unique_ptr<cudf::column> nearest_segment_id;
  std::unique_ptr<cudf::column> nearest_point_on_linestring_xy;
};
```

cpp/include/cuspatial/point_linestring_nearest_points.hpp
```diff
@@ -0,0 +1,174 @@
...
+ * @brief Compute the nearest points and geometry id between pairs of (multi)point and
+ * (multi)linestring
+ *
+ * The nearest point from a test point to a linestring is a point on the linestring that has
+ * the shortest distance to the test point compared to any other points on the linestring.
+ *
+ * The nearest point from a test multipoint to a multilinestring is the nearest point in
+ * the multilinestring that has the shortest distance between all pairs of points and linestrings.
+ *
+ * Returns a structure containing the id to the nearest point in (multi)point, the id to the
```
harrism, line 65:
```suggestion
 * Returns a structure of columns containing (for each pair) the id to the nearest point in (multi)point, the id to the
```

cpp/include/cuspatial/point_linestring_nearest_points.hpp
```diff
@@ -0,0 +1,174 @@
...
+ * point_linestring_nearest_points_result{
+ *  {1, 0, 0},
+ *  {0, 1, 0},
+ *  {0, 0, 2},
+ *  {3.545131432802666, 2.30503517215846, 9.9, 9.4, 0.0, -8.7}
+ * }
+ * ```
+ *
+ * @param multipoint_geometry_offsets Beginning and ending indices for each multipoint
+ * @param points_xy Interleaved x, y-coordinate of points
```
harrism, line 150: interleaved x and y doesn't match other cuspatial column-based APIs. We have always had the X/Y and Lon/Lat positions in separate columns, right?

cpp/tests/utility/vector_equality.hpp
```diff
@@ -28,7 +28,6 @@
 #include <gtest/gtest.h>
 
 #include <type_traits>
-
```
harrism, line 31: I think this blank line should be left in...

harrism (changes requested):
Thanks @isVoid !

cpp/include/cuspatial/point_linestring_nearest_points.hpp
isVoid: We started making this adaptation for https://github.com/rapidsai/cuspatial/issues/649. We could keep them separate but on the python side this would always require 2 scatter operations before we can call c++/

cpp/include/cuspatial/point_linestring_nearest_points.hpp
isVoid: This is better. Thanks.

cpp/include/cuspatial/point_linestring_nearest_points.hpp
harrism: Oh. I was under the impression that if you are using the column-based API you are using dataframes, which tend to have a separate column for each field. What does GeoPandas do here?

cpp/include/cuspatial/experimental/point_linestring_nearest_points.cuh
isVoid: We don't need a line for scalar types. Instead, I think we need a precondition for the base type of the input iterators.

isVoid: >  I think we need a precondition for the base type of the input iterators.

I merged this in the previous precondition.

cpp/include/cuspatial/point_linestring_nearest_points.hpp
isVoid: A GeoSeries of a geopandas is an array of structures. There's current a discussion over the format of xy-coordinates going on at geoarrow: 
https://github.com/geopandas/geo-arrow-spec/pull/26

cpp/include/cuspatial/experimental/detail/point_linestring_nearest_points.cuh
isVoid: Yes, abstracting different level of iteration is direly needed. Perhaps we can continue the discussion in #677 ?

cpp/include/cuspatial/experimental/point_linestring_nearest_points.cuh
```diff
@@ -0,0 +1,102 @@
...
+ * @tparam Cart2dItB iterator type for point array of the linestring element of each pair. Must meet
+ * the requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIteratorA iterator type for `point_geometry_offset` array. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIteratorB iterator type for `linestring_geometry_offset` array. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIteratorC iterator type for `linestring_part_offset` array. Must meet the
+ * requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OutputIt iterator type for output array. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
```
thomcom, line 51:
```suggestion
 * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
 
```

cpp/include/cuspatial/point_linestring_nearest_points.hpp
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * The second pair:
+ * Point: (1.0, 2.0)
+ * Linestring: (0.0, 0.0) -> (3.0, 1.0) -> (3.9, 4) -> (5.5, 1.2)
+ * Nearest segment ID in linestring: 0
+ * Nearest Point on LineString Coordinate: (1.5, 0.5)
+ *
+ * Input:
+ * multipoint_parts_offsets: std::nullopt
+ * points_xy: {0.0, 0.0}
```
thomcom, line 80:
```suggestion
 * points_xy: {0.0, 0.0, 1.0, 2.0}
```

cpp/include/cuspatial/point_linestring_nearest_points.hpp
thomcom: We argued successfully to use xy interleaved coordinates for points to improve cache coherency and word size (coalesced memory access) availability for kernels when `GeoArrow` was designed. @isVoid has been in the process of converting our existing APIs to support `GeoArrow` more directly. I'll be following up in the above conversation today.

thomcom (approved):

harrism (approved):

harrism:
rerun tests

isVoid:
Upon addressing this comment: https://github.com/rapidsai/cuspatial/pull/658#discussion_r974815787, I introduced a bug overwrites `nearest_points` under every iteration. When I attempt to fix this bug, I realized that `thrust::tuple` still doesn't work well with structure binding. (The binding works, but the bounded variable is a `thrust::detail::cons`) that requires more unwrapping. I simply reverted this change.

for thrust::tuple min repro: https://godbolt.org/z/drTr4dx8z

thomcom:
I think this is ready to merge @isVoid? Congrats! :)

isVoid:
@gpucibot merge

PR #660, opened 2022-08-23
Add Multi-Geometry support to `point_linestring_distance` and build python bindings

isVoid:
## Description

This PR adds multi-point and multi-linestring support to `point_linestring_distance`. The c++ API **always** supports multi-variant of the geometry type, since any single-geometry type can be wrapped with a `counting_iteartor` for its underlying geometry offset.

The column API accepts a `std::optional` for the geometry offset inputs. If not provided, they are considered single geometry by default. A `multigeom_dispatcher` is used to generalize the runtime to compile optional information.

This PR also builds python bindings for `point_linestring_distance`. First, the cython bindings is created with a backported `optional` module from cython 3.0 to support the above `std::optional` API. The cython APIs also made use of the dynamic typing property of python to allow geometry offsets to be optional arguments. The python API is the first examplar of a computing API accepting GeoSeries as input. For simplicity, we assume `GeoSeries` only contains single type geometry and there is no mixing of `points` and `multipoints`.

Contributes to #231 
Follow up to #573 

## Checklist


cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
```diff
@@ -83,34 +105,64 @@ void __global__ pairwise_point_linestring_distance(Cart2dItA points_first,
   for (auto idx = threadIdx.x + blockIdx.x * blockDim.x;
        idx < std::distance(linestring_points_first, thrust::prev(linestring_points_last));
        idx += gridDim.x * blockDim.x) {
-    auto offsets_iter =
-      thrust::upper_bound(thrust::seq, linestring_offsets_first, linestring_offsets_last, idx);
+    auto linestring_offsets_iter = thrust::upper_bound(
+      thrust::seq, linestring_part_offsets_first, linestring_part_offsets_last, idx);
```
harrism, line 109: If you pass a counting_iterator for `linestring_part_offsets_first` and `last`, does this get compiled away?  This would be a good thing to verify...  If the counting iterator starts at zero, it should go away completely. If not, it should compile down to a single add, right?

isVoid: I don't think it's getting optmized away: https://godbolt.org/z/YfKanr9eP, and also, the iteartor that may use counting iterator is `linestring_geometry_offsets_first`.

cpp/src/spatial/point_linestring_distance.cu
```diff
@@ -30,37 +30,74 @@
...
 #include <type_traits>
 
+#include "../utility/multigeom_dispatch.hpp"
+
 namespace cuspatial {
 
 namespace {
 
+template <bool has_value>
+struct get_iterator_functor;
```
isVoid, line 45: Move to a utility header file.

cpp/src/spatial/point_linestring_distance.cu
```diff
@@ -80,47 +117,69 @@ struct pairwise_point_linestring_distance_functor {
...
-  return cudf::type_dispatcher(points_xy.type(),
-                               pairwise_point_linestring_distance_functor{},
-                               points_xy,
-                               linestring_offsets,
-                               linestring_points_xy,
-                               stream,
-                               mr);
-}
+template <bool is_multi_point, bool is_multi_linestring>
+struct pairwise_point_linestring_distance_f {
```
isVoid, line 121: Find a better name.

cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
```diff
@@ -83,34 +105,64 @@ void __global__ pairwise_point_linestring_distance(Cart2dItA points_first,
...
-    auto point_idx = thrust::distance(linestring_offsets_first, thrust::prev(offsets_iter));
+    if (linestring_offsets_iter != linestring_part_offsets_last &&
+        *linestring_offsets_iter - 1 == idx) {
+      continue;
+    }
+
+    auto linestring_offsets_idx =
+      thrust::distance(linestring_part_offsets_first, thrust::prev(linestring_offsets_iter));
+
+    auto part_offset_iter = thrust::upper_bound(thrust::seq,
```
isVoid, line 121: Wrong name of iterator. Add some comments to explain. Explore factoring outa as utility functions for traversal.

cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
```diff
@@ -83,34 +105,64 @@ void __global__ pairwise_point_linestring_distance(Cart2dItA points_first,
...
-    cartesian_2d<T> const c = points_first[point_idx];
+    T min_distance_squared  = std::numeric_limits<T>::max();
+
+    for (auto point_idx = point_geometry_offset_first[part_idx];
+         point_idx < point_geometry_offset_first[part_idx + 1];
+         point_idx++) {
+      cartesian_2d<T> const c = points_first[point_idx];
 
-    auto const distance_squared = point_to_segment_distance_squared(c, a, b);
+      auto const distance_squared = point_to_segment_distance_squared(c, a, b);
```
isVoid, line 138: Add TODO: optimize redundant computation in this helper.

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -213,3 +221,41 @@ def pairwise_linestring_distance(offsets1, xs1, ys1, offsets2, xs2, ys2):
             )
         }
     )
+
+
+def pairwise_point_linestring_distance(
+    points: GeoSeries, linestrings: GeoSeries
+):
```
isVoid, line 228: Add docstring.

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -213,3 +221,128 @@ def pairwise_linestring_distance(offsets1, xs1, ys1, offsets2, xs2, ys2):
...
+    """Compute shortest distance between pairs of points and linestrings
+
+    The shortest distance between a pair of (multi)point and (multi)linestring
+    is defined as the shortest distance between every geometry in the
+    multipoint and every geometry in the multilinestring.
```
harrism, line 233: I think we should be explicit about what the geometric test is here.

```suggestion
    is defined as the shortest distance between every point in the
    multipoint and every line segment in the multilinestring.
```

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -213,3 +221,128 @@ def pairwise_linestring_distance(offsets1, xs1, ys1, offsets2, xs2, ys2):
...
+
+    Parameters
+    ----------
+    points : GeoSeries
+        The points to compute the distance from.
+    linestrings : GeoSeries
+        The linestrings to compute the distance to.
```
harrism, line 240: Be careful, this is not a directed distance. Also, "multi".

```suggestion
    points : GeoSeries
        The (multi)points to compute the distance from.
    linestrings : GeoSeries
        The (multi)linestrings to compute the distance from.
```

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -213,3 +221,128 @@ def pairwise_linestring_distance(offsets1, xs1, ys1, offsets2, xs2, ys2):
             )
         }
     )
+
+
+def pairwise_point_linestring_distance(
+    points: GeoSeries, linestrings: GeoSeries
+):
+    """Compute shortest distance between pairs of points and linestrings
```
harrism, line 229: I think the body of this docstring needs something explaining that this is pairwise, not all-pairs -- it's not clear from the input parameters how they are paired up.

harrism (changes requested):
Nice work!

cpp/include/cuspatial/detail/iterator.hpp
```diff
@@ -29,5 +31,25 @@ inline auto make_counting_transform_iterator(cudf::size_type start, UnaryFunctio
...
+  }
+};
+
+template <>
+struct get_iterator_functor<false> {
+  auto operator()(std::optional<cudf::device_span<cudf::size_type const>>)
+  {
+    return thrust::make_counting_iterator(0);
+  }
+};
```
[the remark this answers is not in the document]
isVoid: Perfect, I meant to call them this but forgot to rename after moving.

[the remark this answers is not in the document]
isVoid: Oh, this isn't meant to be included in here. I expect these iterators should only be used in `.cu` files so I included them in `src/utility/iteartor.hpp`. These should be removed.

cpp/src/utility/iterator.hpp
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include <cudf/types.hpp>
+#include <cudf/utilities/span.hpp>
+
+#include <optional>
+namespace cuspatial {
+namespace detail {
+
+template <bool has_value>
```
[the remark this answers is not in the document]
isVoid: I actually meant to keep this file and remove the other one in `include`. I think this might only needs to be used in `.cu` files.

cpp/src/utility/multigeom_dispatch.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#pragma once
+
+#include <utility>
+
+namespace cuspatial {
+
+template <template <bool is_multi_1, bool is_multi_2> class Functor, typename... Args>
+auto multigeom_dispatch(bool opt1, bool opt2, Args&&... args)
+{
```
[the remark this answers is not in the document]
isVoid: Done.

python/cuspatial/cuspatial/core/spatial/distance.py
isVoid: Done

cpp/src/utility/multigeom_dispatch.hpp
```diff
@@ -0,0 +1,53 @@
+
```
harrism, line 1: Should the file be renamed too?

harrism (approved):

cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
```diff
@@ -120,11 +177,12 @@ void pairwise_point_linestring_distance(Cart2dItA points_first,
...
                                 detail::iterator_value_type<Cart2dItB>>(),
-                "Inputs must be cuspatial::vec_2d<T>");
+                "Inputs must be cuspatial::vec_2d");
 
-  auto const num_pairs = thrust::distance(points_first, points_last);
+  auto const num_pairs =
+    thrust::distance(point_geometry_offset_first, point_geometry_offset_last) - 1;
 
-  if (num_pairs == 0) { return; }
+  if (num_pairs == 0) { return distances_first; }
```
thomcom, line 185: Need to check that the `num_pairs` is also equal to the number of linestrings in `linestring_geometry_offset_first`, or the number of `linestring_part_offsets` if `linestring_geometry_offset_first` is `nullptr`?

I'm not sure from having read yet but I think I saw that either the algorithm treats the multilinestring and linestring situations as different cases? What about always passing in the multilinestring offsets buffer, creating it if one isn't included by the caller, and making it as simple as `[0, 1]` if only a single point/linestring is passed in?

thomcom (changes requested):

cpp/include/cuspatial/distance/point_linestring_distance.hpp
```diff
@@ -35,30 +35,65 @@ namespace cuspatial {
...
+ * ```
+ *
+ * @param multipoint_geometry_offsets Beginning indices to each geometry in the multi-point
  * @param points_xy Interleaved x, y-coordinates of points
- * @param linestring_offsets Beginning and ending indices for each linestring in the `linestring_x`
+ * @param multilinestring_geometry_offsets Beginning indices to each geometry in the
+ multi-linestring
+ * @param linestring_part_offsets Beginning and ending indices for each linestring in the
+ `linestring_x`
  * and `linestring_y` arrays. Must satisfy `linestring_offsets.size() + 1 == points_xy.size()`.
```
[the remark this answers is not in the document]
isVoid: Also I noticed that the docs here doesn't mention the interleaved point coordinates.

cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
```diff
@@ -42,38 +42,60 @@ namespace detail {
...
+          class OffsetIteratorC,
+          class OutputIterator>
+void __global__ pairwise_point_linestring_distance(OffsetIteratorA point_geometry_offset_first,
+                                                   OffsetIteratorA point_geometry_offset_last,
+                                                   Cart2dItA points_first,
+                                                   Cart2dItA points_last,
+                                                   OffsetIteratorB linestring_geometry_offset_first,
+                                                   OffsetIteratorB linestring_geometry_offset_last,
+                                                   OffsetIteratorC linestring_part_offsets_first,
+                                                   OffsetIteratorC linestring_part_offsets_last,
```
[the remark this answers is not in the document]
isVoid: I believe you are just looking at a recursive templated iterator collection, there's some maximum level that c++ compiler allows the templates to be recursively nested within itself, but multilinestring should be well within that limit. 

Does this make sense to you? https://godbolt.org/z/v95jn5G3e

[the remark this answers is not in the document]
thomcom: Yes, that's more or less exactly what I was thinking of. Do you think it works in the context of your PR?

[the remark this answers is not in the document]
isVoid: This really is just my first attempt to take a stab at the idea. I'd like to have some more feedbacks before we decide on the iterator collection interface. So no, not planning to put this in this PR.

cc @harrism

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -213,3 +221,132 @@ def pairwise_linestring_distance(offsets1, xs1, ys1, offsets2, xs2, ys2):
...
+    linestrings : GeoSeries
+        The (multi)linestrings to compute the distance from.
+
+    Returns
+    -------
+    distance : cudf.Series
+
+    Notes
+    -----
+    The input `GeoSeries` must contain a single type geometry.
```
thomcom, line 252:
```suggestion
    The input `GeoSeries` should contain a single type geometry.
```

I don't think there's a problem passing in a GeoSeries that has other feature types in it, they'll just be ignored. Theoretically if a GeoSeries contained an equal number of points and linestrings it could be the first and second argument to this function? I think?

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -213,3 +221,132 @@ def pairwise_linestring_distance(offsets1, xs1, ys1, offsets2, xs2, ys2):
...
+    1    33.241540
+    2     0.680451
+    dtype: float64
+    """
+    if not contains_only_points(points):
+        raise ValueError("`points` array must contain only points")
+
+    if not contains_only_linestrings(linestrings):
+        raise ValueError("`linestrings` array must contain only linestrings")
+
```
thomcom, line 325: These don't seem strictly necessary as above, can't we use any `GeoSeries` as long as `len(points) == len(linestrings)`?

cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
isVoid: The current algorithm *is* basically what you mentions above :) Except users can use a counting iterator to avoid materializing the sequence. `linestring_geometry_offset_first` is assumed to never be null.

python/cuspatial/cuspatial/utils/column_utils.py
```diff
@@ -42,3 +44,46 @@ def normalize_timestamp_column(ts, fallback_dtype="datetime64[ms]"):
...
+    """
+
+    return contain_single_type_geometry(gs) and (
+        len(gs.points.xy) > 0 or len(gs.multipoints.xy) > 0
+    )
+
+
+def contains_only_linestrings(gs: GeoSeries):
+    """
+    Returns true if `gs` contains only linestring
```
thomcom, line 78:
```suggestion
    Returns true if `gs` contains only linestrings
```

python/cuspatial/cuspatial/utils/column_utils.py
```diff
@@ -42,3 +44,46 @@ def normalize_timestamp_column(ts, fallback_dtype="datetime64[ms]"):
...
+    """
+
+    return contain_single_type_geometry(gs) and len(gs.lines.xy) > 0
+
+
+def contains_only_polygons(gs: GeoSeries):
+    """
+    Returns true if `gs` contains only polygon
```
thomcom, line 86:
```suggestion
    Returns true if `gs` contains only polygons
```

thomcom (commented):
Overall looks really great. Just a few comments, questions, and textual updates.

cpp/src/spatial/point_linestring_distance.cu
```diff
@@ -30,37 +30,59 @@
...
-    pairwise_point_linestring_distance(points_it,
-                                       points_it + num_pairs,
-                                       linestring_offsets.begin(),
+    pairwise_point_linestring_distance(point_parts_it_first,
+                                       point_parts_it_first + num_pairs + 1,
+                                       points_it,
+                                       points_it + num_points,
+                                       linestring_parts_it_first,
+                                       linestring_part_offsets.begin(),
+                                       linestring_part_offsets.end(),
```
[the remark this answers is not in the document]
isVoid: Yes, the motivation is to avoid writing 4 functions. I'm leveraging c++'s template mechanism to help me do that :)

cpp/src/spatial/point_linestring_distance.cu
```diff
@@ -78,49 +100,69 @@ struct pairwise_point_linestring_distance_functor {
...
+
+    CUSPATIAL_EXPECTS(num_lhs == num_rhs, "Mismatch number of points and linestrings.");
+
+    CUSPATIAL_EXPECTS(points_xy.type() == linestring_points_xy.type(),
+                      "Points and linestring coordinates must have the same type.");
+
+    CUSPATIAL_EXPECTS(!(points_xy.has_nulls() || linestring_points_xy.has_nulls()),
+                      "All inputs must not have nulls.");
+
+    if (num_rhs - 1 == 0) return cudf::make_empty_column(points_xy.type());
```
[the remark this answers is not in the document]
isVoid: Any offset can be length `[1, inf]`. In either multilinestring or linestring's case, if there is 0 element, the length of the offset is 1 and contains a single element `0`. Therefore, if the `num_rhs == 1`, the input contains 0 pair of geometries.

cpp/tests/experimental/spatial/point_linestring_distance_test.cu
```diff
@@ -46,89 +46,151 @@ TYPED_TEST(PairwisePointLinestringDistanceTest, Empty)
...
+                                                d_linestring_points.end(),
+                                                got.begin());
 
   thrust::host_vector<T> hgot(got);
   expect_vector_equivalent(expect, hgot);
+  EXPECT_EQ(ret, got.end());
 }
 
-TYPED_TEST(PairwisePointLinestringDistanceTest, TwoPairFromVector)
+TYPED_TEST(PairwisePointLinestringDistanceTest, TwoPairFromVectorSingleComponent)
```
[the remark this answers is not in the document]
isVoid: This is no multi right? Both the geometry offsets (noticed the names require change) are `[0, 1, 2]`, effectively this is only a single point/linestring wrapped in multipoint and multilinestring.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -110,6 +110,33 @@ def y(self):
...
+            return cudf.Series(self._col.offsets.values)
+
+        @property
+        def part_offset(self):
+            return cudf.Series(self._col.elements.offsets.values)
+
+        @property
+        def ring_offset(self):
+            return cudf.Series(self._col.elements.elements.offsets.values)
+
```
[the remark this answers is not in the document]
isVoid: I would defend my implementation by that this provides a uniform accessor for both offset array and coordinate arrays and users don't have to resort to different APIs if they need to access both. And this provides some name consistencies (geometry -> part -> ring). We can also abstract this as a inheritance model (ColumnAccessor <- GeometryAccessor <- PartAccessor <- RingAccessor) and instantiate each type as needed.

python/cuspatial/cuspatial/core/spatial/distance.py
isVoid: The result of a pairwise distance function should be of the same length as its input. I would rather the user explicitly extract the points and linestring from their geoseries to form a new one before passing in this function. Do we have an easy way for user to do that?

python/cuspatial/cuspatial/core/spatial/distance.py
isVoid: See answer in https://github.com/rapidsai/cuspatial/pull/660/files/6c57187226f47e0c646f0c659e4e6a27f076af2f#r966410266

cpp/include/cuspatial/experimental/detail/point_linestring_distance.cuh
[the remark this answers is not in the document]
isVoid: There could be other simpler solutions, like just a `MultiLinestringIteratorCollection` class for example:
```c++
template<typename OffsetItA, typename OffsetItB, typename Vec2dIt>
struct MultiLinestringIteratorCollection
{
  OffsetItA geometry_offset_begin;
  OffsetItB part_offset_begin;
  Vec2dIt coordinate_begin;
};
```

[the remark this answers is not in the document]
isVoid: This deviates a little bit from you original suggestion of having a custom iterator to represent a nested of iterator. I believe you meant to create a custom iterator that iterates over the multilinestring array, when you dereference this iterator, it returns a multilinestring - which in turn has its own iterator, and allows you to iterate into each linestring, segment and points etc. Besides, we should provide different level of iterations so that developer can pick their favorite parallel model as needed.

I don't have a working example, but something like:
```c++
class linestring {
    class iterator{
        std::pair<vec2d, vec2d> operator* (); // returns two end points of the line segment
        iterator operator+(int i);
        iterator operator-(int i);
        // other requirements of of LRAIs
    };
    iterator begin();
    iterator end();
};

class multi_linestring
{
    class linestring_iterator
    {
        linestring operator*(); // this returns a linestring
        // LRAI requirements
    };
    class segment_iterator
    {
        std::pair<vec2d, vec2d> operator* (); // returns two end points of the line segment
        // LRAI requirments
    }

    linestring_iterator linestring_begin();
    linestring_iterator linestring_end();

    segment_iterator segment_begin();
    segment_iterator segment_end();

    // The default should still be immediate next level of indirection
    auto begin() { return linestring_begin(); }
    auto end() { return linestring_end(); }
};

class multi_linestring_array
{
    class multi_linestring_iterator
    {
        multi_linestring operator*();
        // other LRAI requirements
    }
    class linestring_iterator { /* same as above */ }
    class segment_iterator { /* same as above */ }

    multilinestring_iterator multilinestring_begin();
    multilinestring_iterator multilinestring_end();

    linestring_iterator linestring_begin();
    linestring_iterator linestring_end();

    segment_iterator segment_begin();
    segment_iterator segment_end();

    // The default should still be immediate next level of indirection
    auto begin() { return multilinestring_begin(); }
    auto end() { return multilinestring_begin(); }
};
```

cpp/include/cuspatial/distance/point_linestring_distance.hpp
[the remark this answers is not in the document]
isVoid: I updated this section a bit to address for the confusion. This line is removed since I expect it to be a common sense for the users.

cpp/include/cuspatial/experimental/point_linestring_distance.cuh
```diff
@@ -30,20 +30,30 @@ namespace cuspatial {
  * the requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
  * @tparam Cart2dItB iterator type for point array of the linestring element of each pair. Must meet
  * the requirements of [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
- * @tparam OffsetIterator iterator type for offset array. Must meet the requirements of
+ * @tparam OffsetIteratorA iterator type for offset array. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIteratorB iterator type for offset array. Must meet the requirements of
+ * [LegacyRandomAccessIterator][LinkLRAI] and be device-accessible.
+ * @tparam OffsetIteratorC iterator type for offset array. Must meet the requirements of
```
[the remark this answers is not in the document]
isVoid: Updated with their respective formal argument names.

python/cuspatial/cuspatial/core/spatial/distance.py
thomcom: No. In `geopandas`, the syntax commonly used is `points = geoseries[geoseries.type == "Point"]` and we need to implement some kind of the `.type` function in order to duplicate it. It can be hacked by using the `GeoColumn._meta` but we should just write the `.type` accessor. :)

thomcom (approved):

python/cuspatial/cuspatial/core/geoseries.py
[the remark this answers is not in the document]
thomcom: I got confused writing docs as to whether I should use `offset_buffer(0)` or `offset_buffer(1)` so let's go with yours. I'll add it to #609 when this is merged.

cpp/src/spatial/point_linestring_distance.cu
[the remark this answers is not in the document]
isVoid: (Will leave this discussion open, should not block this PR to merge)

I might have missed something you suggested here. Are you suggesting that these APIs should take a `LIST` type columns?

isVoid:
@gpucibot merge

PR #669, opened 2022-09-13
Expand expect_vector_equivalent to handle std::vector of vec_2d<T> and move traits out of detail

harrism:
## Description
* Expands the test utility `expect_vector_equivalent` to support comparing vectors of `vec_2d<T>` with approximate comparison of floating point equality (using GTEST functionality). This function now also explicitly copies all vectors to the host for comparison.
* Updates various tests to use the utility instead of directly calling `EXPECT_EQ`.
* Moves traits.hpp to top-level include directory and moves all traits helpers out of `detail` namespace so they are more easily used in tests and user code.

TODO: 

 ~- [ ] `coordinate_transform_test` fails with this change, so I have not committed it. Something is different with `EXPECT_EQ` and this function for that test. Investigating.~  This will be fixed in a followup PR -- we discovered the test is flawed.

## Checklist

### DIFF
--- a/cpp/tests/utility/vector_equality.hpp
+++ b/cpp/tests/utility/vector_equality.hpp
@@ -16,24 +16,77 @@
 
 #pragma once
 
+#include <cuspatial/vec_2d.hpp>
+
+#include <rmm/device_uvector.hpp>
+#include <rmm/device_vector.hpp>
+
+#include <thrust/host_vector.h>
+
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 #include <type_traits>
 
+template <typename T>
+std::ostream& operator<<(std::ostream& os, cuspatial::vec_2d<T> const& vec)
+{
+  return os << "(" << vec.x << "," << vec.y << ")";
+}
+
 namespace cuspatial {
 namespace test {
+
+template <typename T>
+auto float_matcher(T val)
+{
+  if constexpr (std::is_same_v<T, float>) {
+    return ::testing::FloatEq(val);
+  } else {
+    return ::testing::DoubleEq(val);
+  }
+}
+
+MATCHER(vec_2d_matcher,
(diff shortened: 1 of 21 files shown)


cpp/tests/utility/vector_equality.hpp
```diff
@@ -16,24 +16,77 @@
...
+      ::testing::Matches(float_matcher(rhs.y))(lhs.y))
+    return true;
+
+  //*result_listener << lhs << " != " << rhs;
+
+  return false;
+}
+
+template <typename T>
+constexpr bool is_vec_2d()
```
isVoid, line 66: and this in `traits.hpp`?

isVoid (approved):

cpp/tests/utility/vector_equality.hpp
harrism: My problem was that traits.hpp stuff is all in the `detail` namespace, and I don't like reaching into `detail` from tests.

harrism: Moved traits out of detail and added this to traits.hpp.

cpp/tests/utility/vector_equality.hpp
```diff
@@ -16,24 +16,77 @@
...
+
+#include <thrust/host_vector.h>
+
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 #include <type_traits>
 
+template <typename T>
+std::ostream& operator<<(std::ostream& os, cuspatial::vec_2d<T> const& vec)
```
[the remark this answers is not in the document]
harrism: Done.

isVoid (approved):

isVoid:
Lots of nice refactoring of old test code too... Nice improvement.

harrism:
@gpucibot merge

PR #670, opened 2022-09-14
`linestring_distance` Header Only API Returns Past-the-End iterator

### LINKED ISSUE #576: [BUG] Pairwise linestring distance should return one past the last written output iterator
According to [development doc](https://github.com/rapidsai/cuspatial/blob/branch-22.08/cpp/doc/libcuspatial_refactoring_guide.md#example-api), to be consistent with `std::transform` API, the method should return the output iterator to one past the last element being written to by the algorithm. Currently, pairwise linestring distance returns `void`:
https://github.com/rapidsai/cuspatial/blob/dbebf67d422104dc47e8ed2ac5a6d2ec75e2e8d0/cpp/include/cuspatial/experimental/linestring_distance.cuh#L61

isVoid:
## Description

This PR updates `linestring_distance` to return past-the-end iterator to conform to other interfaces. Besides, this PR updates the assumption of `offset` arrays to be arrow compliant format (`N+1` instead of `N`).

closes #576 

## Checklist


cpp/include/cuspatial/experimental/linestring_distance.cuh
```diff
@@ -59,15 +59,15 @@ namespace cuspatial {
  * "LegacyRandomAccessIterator"
```
harrism, line 59: Add @return doc

harrism (approved):
One missing @return doc.

thomcom (approved):

isVoid:
Thanks for the review. I updated the python tests to address the changes to offset format in this PR.

cpp/include/cuspatial/experimental/linestring_distance.cuh
isVoid: don

isVoid:
@gpucibot merge

PR #671, opened 2022-09-14
Fix and tidy coordinate_transform_test

harrism:
## Description
Converting from `EXPECT_EQ` to `expect_vector_equivalent` (#669) caused `coordinate_transform_test` to fail due to floating point differences. It's strange that it passed with exact equality testing (perhaps a problem in Thrust?). In any case, this fixes the test data generation and DRYs up the tests a bit.

## Checklist

### DIFF
--- a/cpp/include/cuspatial/constants.hpp
+++ b/cpp/include/cuspatial/constants.hpp
@@ -27,9 +27,12 @@ namespace cuspatial {
 
 constexpr double DEGREE_TO_RADIAN = M_PI / 180.0;
 constexpr double RADIAN_TO_DEGREE = 180.0 / M_PI;
+constexpr double CIRCLE_DEGREES   = 360.0;
 
 constexpr double EARTH_RADIUS_KM                = 6371.0;
 constexpr double EARTH_CIRCUMFERENCE_EQUATOR_KM = 40000.0;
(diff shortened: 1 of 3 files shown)


thomcom (approved):

isVoid (approved):
On a second thought:

Philosophically this alters the tests from testing against a golden answer to comparing the result between device compute and host compute. This is acceptable as source of truth only when we are certain that the host computation is correct (which is easy to verify since the kernel of `coordinate_transform` is very simple.

I think it would be nice in https://github.com/rapidsai/cuspatial/pull/672 we can discuss a bit about "sources of truth". Pen and paper always works and I think developers should exhaust corner cases with simple inputs first. For larger inputs I would also resort to random generated inputs/existing datasets and computes the result with a mature host counterpart, such as `shapely`, `geopandas`, `arcgis`, `postgregis`, `libgeos` etc.

It's also ok to include these in python testing docs too, if you think they fit more over there.

---

It also comes to me what happens if you compare the hardcoded value with the host computed result?

harrism:
> It also comes to me what happens if you compare the hardcoded value with the host computed result?

Indeed, I tested and the host-computed FP32 result does not match the hard-coded FP64 result. So if we want to hard-code expected values we need one per type.

harrism:
@gpucibot merge

Issue #676, opened 2022-09-14
[BUG] GeoDataframe has broken slicing.


thomcom:
**Describe the bug**
If I try to use `gpu_dataframe.head()` or `.iloc` a strange error that doesn't touch cuspatial is thrown:

```
In [25]: gpu_dataframe.head()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [25], line 1
----> 1 gpu_dataframe.head()

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/contextlib.py:75, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     72 @wraps(func)
     73 def inner(*args, **kwds):
     74     with self._recreate_cm():
---> 75         return func(*args, **kwds)

File ~/cudf/python/cudf/cudf/core/frame.py:2648, in Frame.head(self, n)
   2567 @_cudf_nvtx_annotate
   2568 def head(self, n=5):
   2569     """
   2570     Return the first `n` rows.
   2571     This function returns the first `n` rows for the object based
   (...)
   2646     1    1  11.0
   2647     """
-> 2648     return self.iloc[:n]

File ~/cudf/python/cudf/cudf/core/dataframe.py:144, in _DataFrameIndexer.__getitem__(self, arg)
    142 if not isinstance(arg, tuple):
    143     arg = (arg, slice(None))
--> 144 return self._getitem_tuple_arg(arg)

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/contextlib.py:75, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     72 @wraps(func)
     73 def inner(*args, **kwds):
     74     with self._recreate_cm():
---> 75         return func(*args, **kwds)

File ~/cudf/python/cudf/cudf/core/dataframe.py:443, in _DataFrameIlocIndexer._getitem_tuple_arg(self, arg)
    441 else:
    442     if isinstance(arg[0], slice):
--> 443         df = columns_df._slice(arg[0])
    444     elif is_scalar(arg[0]):
    445         index = arg[0]

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/contextlib.py:75, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     72 @wraps(func)
     73 def inner(*args, **kwds):
     74     with self._recreate_cm():
[...]

PR #680, opened 2022-09-19
Fix `GeoDataframe` slicing issue by adding `_slice_` method.

thomcom:
## Description

This PR will adds a `_slice` method which is called when a `cudf.DataFrame` is accessed in a variety of ways. It also adds the `name` member to `GeoSeries` that are pulled from the `GeoDataFrame`.

Fixes https://github.com/rapidsai/cuspatial/issues/676

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/geodataframe.py
+++ b/python/cuspatial/cuspatial/core/geodataframe.py
@@ -149,9 +149,7 @@ class GeoDataFrame(cudf.DataFrame):
             name: (
                 sliced_geo_columns[name] if mask else sliced_data_columns[name]
             )
-            for name, mask in zip(
-                columns_mask.values_host, geocolumn_mask.values
-            )
+            for name, mask in zip(columns_mask.values, geocolumn_mask.values)
         }
         return self.__class__(output)
 
--- a/python/cuspatial/cuspatial/core/geodataframe.py
+++ b/python/cuspatial/cuspatial/core/geodataframe.py
@@ -151,20 +151,10 @@ class GeoDataFrame(cudf.DataFrame):
         # Send the rest of the columns to `cudf` to slice.
         data_columns = cudf.DataFrame(
             self[
-                columns_mask[
-                    False == geocolumn_mask
-                ].values_host  # noqa: E712 E501
+                columns_mask[False == geocolumn_mask].values_host  # noqa: E712
             ]
         )
         sliced_data_columns = data_columns._slice(arg)
-        """
-        output = {}
-        for key, value in zip(columns_mask.values_host, geocolumn_mask.values):
-            if value:
-                output[key] = sliced_geo_columns[key]
-            else:
-                output[key] = sliced_data_columns[key]
-        """
         output = {
             key: (
                 sliced_geo_columns[key] if value else sliced_data_columns[key]
--- a/python/cuspatial/cuspatial/core/geodataframe.py
+++ b/python/cuspatial/cuspatial/core/geodataframe.py
@@ -128,12 +128,6 @@ class GeoDataFrame(cudf.DataFrame):
 
         return self
 
-    def __getitem__(self, arg):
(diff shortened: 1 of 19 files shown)


python/cuspatial/cuspatial/tests/test_geodataframe.py
```diff
@@ -241,6 +241,20 @@ def test_post_slice(gpdf, post_slice):
     assert_eq_geo_df(gi[post_slice], cugpdf_back[post_slice])
 
 
+@pytest.mark.parametrize(
+    "inline_slice",
+    [slice(0, 12)]
+    + [slice(0, 10, 1)]
+    + [slice(0, 3, 1)]
+    + [slice(3, 6, 1)]
+    + [slice(6, 9, 1)],
```
isVoid, line 250: This seems like a rather strange way to construct a list? Did black change it? Why not just
```suggestion
    [slice(0, 12),
    slice(0, 10, 1),
    slice(0, 3, 1),
    slice(3, 6, 1),
    slice(6, 9, 1)],
```

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -120,6 +126,35 @@ def _copy_type_metadata(
 
         return type_copied
 
+    def _slice(self: T, arg: slice) -> T:
+        """
+        Overload the _slice functionality from cudf's frame members.
+        """
+        # Collect the Geometry columns and slice them separately
+        columns_mask = cudf.Series(self.columns)
```
isVoid, line 134: cudf does not guarantee performance with wide table. Neither should geodataframe. There's no need to use device to handle the column name list.

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -120,6 +126,35 @@ def _copy_type_metadata(
...
+            self[columns_mask[~geocolumn_mask].values_host]
+        )
+        sliced_data_columns = data_columns._slice(arg)
+        output = {
+            key: (
+                sliced_geo_columns[key] if value else sliced_data_columns[key]
+            )
+            for key, value in zip(
+                columns_mask.values_host, geocolumn_mask.values
+            )
```
isVoid, line 154: I'm a bit not sure if it's name should be called "key" and "value".
```suggestion
            name: (
                sliced_geo_columns[name] if mask else sliced_data_columns[name]
            )
            for name, mask in zip(
                columns_mask.values_host, geocolumn_mask.values
            )
```

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -120,6 +126,35 @@ def _copy_type_metadata(
...
+        }
+        return __class__(output)
```
isVoid, line 156: Interestingly, there's a difference between `__class__` and `self.__class__`. The previous always binds to the current object while the latter gets dynamically dispatched to anything that `self` gets bound to.
```python
In [6]: class A:
   ...:     def __init__(self):
   ...:         print("A ctor")
   ...:     def f(self):
   ...:         print(__class__)
   ...:         print(self.__class__)
   ...: 

In [7]: class B(A):
   ...:     def __init__(self):
   ...:         print("B ctor")
   ...: 

In [8]: b = B()
B ctor

In [9]: b.f()
<class '__main__.A'>
<class '__main__.B'>
```

isVoid:
```suggestion
        return self.__class__(output)
```

isVoid (commented):
I assume this pattern will come up very often: filter out geocolumn, handle geocolumn, handle data column, combine geocolumn and data column. Can we generalize this into some helper functions?

thomcom:
Sure! I have done so in the latest commit. If you approve I'll merge soon.

isVoid (approved):

thomcom:
@gpucibot merge

PR #692, opened 2022-09-24
Vectorized Load, refactors `type_utils.hpp` into `iterator_factory.cuh`

isVoid:
## Description

This PR refactors `type_utils.hpp` into `iterator_factory.cuh` and introduce utilize sfinae to adapt vectorized load from interleaved coordinate inputs. It also adds documentation to previously missing APIs.

## Checklist


cpp/include/cuspatial/experimental/iterator_factory.cuh
```diff
@@ -14,22 +14,43 @@
...
 
 #include <type_traits>
 
 namespace cuspatial {
-
 namespace detail {
 
+/**
+ * @brief Get corresponding 2-element vector type from `Element` type.
+ * Only for floating point types.
```
harrism, line 38:
```suggestion
 * @brief Get corresponding 2-element vector type from `Element` type.
 *
 * Only for floating point types.
```

cpp/include/cuspatial/experimental/iterator_factory.cuh
```diff
@@ -54,6 +75,74 @@ struct vec_2d_to_tuple {
...
+template <typename Iter, typename Enable = void>
+struct interleaved_to_vec_2d {
+  using T = typename std::iterator_traits<Iter>::value_type;
+  Iter it;
+  __device__ vec_2d<T> operator()(std::size_t i) { return vec_2d<T>{it[2 * i], it[2 * i + 1]}; }
+};
+
+/**
+ * @brief Specialization for raw pointers to interleaved xy range.
+ *
```
harrism, line 97:
```suggestion
 *
 * This generic version makes use of vectorized load.
 *
```

cpp/include/cuspatial/experimental/iterator_factory.cuh
```diff
@@ -54,6 +75,74 @@ struct vec_2d_to_tuple {
...
+template <typename Iter>
+struct interleaved_to_vec_2d<Iter, typename std::enable_if_t<std::is_pointer_v<Iter>>> {
+  using pointer_t = typename std::pointer_traits<Iter>::pointer;
+  using T         = typename std::remove_const_t<typename std::pointer_traits<Iter>::element_type>;
+  using T2        = typename element_to_element2<T>::type;
+  pointer_t ptr;
+
+  __device__ vec_2d<T> operator()(std::size_t i)
+  {
+    T2 const* f2it = reinterpret_cast<T2 const*>(ptr);
```
harrism, line 109: Hmmm, I think this is technically UB.  I think it requires `std::memcpy` (or `std::bitcast` after C++20).  Does it work if you use `std::memcpy`?  If so, assuming the alignment of `vec_2d<T>` is the same as `T2` (float2, double2), then why not copy directly to a vec_2d rather than a T2?

```suggestion
    vec_2d<T> v2;
    std::memcpy(&v2, ptr, sizeof(v2));
    return v2;
```

cpp/include/cuspatial/experimental/iterator_factory.cuh
```diff
@@ -54,6 +75,74 @@ struct vec_2d_to_tuple {
...
+
+/**
+ * @internal
+ * @brief Functor to transform an index to strided index.
+ */
+struct strided_functor {
+  std::size_t _stride;
+  strided_functor(std::size_t stride) : _stride(stride) {}
+  auto __device__ operator()(std::size_t i) { return i * _stride; }
+};
```
harrism, line 144: Perhaps make stride a template parameter?  Seems it's only used with a compile-time value. I don't see a need for a 64-bit stride here, either?

```suggestion
template <int Stride>
struct strided_functor {
  auto __device__ operator()(std::size_t i) { return i * Stride; }
};
```

cpp/include/cuspatial/experimental/iterator_factory.cuh
```diff
@@ -151,7 +236,7 @@ template <typename Iter>
 auto vec_2d_iterator_to_output_interleaved_iterator(Iter d_points_begin)
```
harrism, line 236: Can this be an overload of `make_vec_2d_iterator`? For consistency, I think iterator factories should start with `make` and end in `iterator`.

harrism (changes requested):
Concerned about strict aliasing violation.

cpp/include/cuspatial/experimental/iterator_factory.cuh
isVoid: Inspired by @jrhemstad i think we can use a std::integer_constant here.

cpp/include/cuspatial/experimental/iterator_factory.cuh
isVoid: This has the same number of input argument to the existing make_vec2d_iterator though. How will the compiler know which overload the find?

harrism: The existing `make_vec2d_iterator` takes two iterator arguments (x and y). This one takes a single iterator argument.

isVoid: Part of this PR is refactoring `interleaved_to_vec_2d_iterator` into `make_vec2d_iterator` (single argument). For this I think it's best to be called `make_output_vec_2d_iterator`.

harrism: Yes, sorry for the confusion, I meant output_iterator.

harrism (approved):

thomcom (approved):

isVoid:
rerun tests

thomcom:
@gpucibot merge

PR #719, opened 2022-10-03
`unwrap_pyoptcol` is missing `except` keyword that causes exceptions ignored, fixes function bug

isVoid:
## Description

Fixes a bug in the cdef function when `pyoptcol` is `None` and gets mishandled (raised as error). This error is captured in one of the pytests for `point_linestring_distance`. However, due to the cdef function is missing the `except` keyword, the exception gets ignored by the python interpreter and not reraised.

## Checklist


thomcom (approved):

isVoid:
@gpucibot merge

PR #734, opened 2022-10-07
Add python bindings for (multi)point-(multi)point distance

isVoid:
## Description

This PR adds python bindings to Point-point distance. Depend on #731 

## Checklist


harrism (approved):
Pretty straightforward wrapper. Looks good.

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -124,6 +126,87 @@ def haversine_distance(p1_lon, p1_lat, p2_lon, p2_lat):
...
+    2     79.854261
+    3     56.844978
+    4    131.292340
+    5    139.818452
+    6    173.507174
+    7    124.346079
+    8     47.829125
+    9    159.156985
+    dtype: float64
+    """
```
thomcom, line 177: This is not a good example particularly because I don't think that cartesian distance in the haversine coordinate space is especially meaningful. You could use `lonlat_to_cartesian` first in order to give the points meter values, as one suggestion.

cpp/tests/spatial/point_distance_test.cpp
```diff
@@ -137,6 +137,6 @@ TEST_F(PairwisePointDistanceTestThrow, TypeMismatch)
   auto xy1     = fixed_width_column_wrapper<float>{1, 1, 2, 2, 3, 3};
   auto xy2     = fixed_width_column_wrapper<double>{1, 1, 2, 2, 3, 3};
 
-  EXPECT_THROW(pairwise_point_distance(xy1, offset1, xy2, offset2), cuspatial::logic_error);
+  EXPECT_THROW(pairwise_point_distance(offset1, xy1, offset2, xy2), cuspatial::logic_error);
```
thomcom, line 140: Did you intend to include all these argument reorderings in this PR?

thomcom (changes requested):
Python looks really solid with easy to read and comprehensive tests. Just a question and a concern about the example that you give in the python docs.

cpp/tests/spatial/point_distance_test.cpp
isVoid: No, this change exists to revert changes that I made in this comment: https://github.com/rapidsai/cuspatial/pull/731#discussion_r991794161

python/cuspatial/cuspatial/core/spatial/distance.py
isVoid: Yeah but that also doesn't make sense right? `lonlat_to_cartesian` (aka sinusoidal projection) assumes all points are the in proximity of the intersection of equator and prime meridian. The points we get from the cities are all over places. Maybe I should instead use a handcrafted example here?

thomcom: That's true, lonlat_to_cartesian across the entire globe would produce nonsense. Maybe just a set of points on a line pairwised against a set of points on a similar line. Something easy would be like [0, 0], [1, 1], [2, 2] ... in one direction and [0, 0], [-1, 1], [-2, 2], ... which would be easy to understand and visualize.

thomcom (approved):

isVoid:
@gpucibot merge

Issue #738, opened 2022-10-17
[FEA] Determine if point-in-polygon is exterior inclusive or not.


thomcom:
**Is your feature request related to a problem? Please describe.**
https://docs.geopandas.org/en/latest/docs/reference/api/geopandas.GeoSeries.contains.html#geopandas.GeoSeries.contains specifies that `.contains` is exterior-exclusive. That is, the boundary of a Polygon is defined to not be contained within the polygon. 

**Describe the solution you'd like**
In order to implement `.contains`, I need to verify if our implementations of point-in-polygon follow this same rule. I'll write a test that validates if a point that matches the point of a Polygon is in-polygon. I'll write another test to see if a point that is collinear to two points in the Polygon is in-polygon. If not, great!

thomcom:
The result of the test is that the basic pip algorithm is border-exclusive for points in the border, but it is border-inclusive for points that are collinear with points in the border. This contrasts GeoPandas which is border-exclusive for both cases.

harrism:
If it was entirely border-inclusive (including vertices) I would wonder if this is intentional. But the fact that border points are exclusive and border edges are inclusive seems inconsistent, and therefore I lean toward this being a bug.

PR #739, opened 2022-10-17
Test boundary exclusion in python and add boundary exclusion to original point_in_polygon

thomcom:
We don't properly boundary exclude edges during point-in-polygon tests. Previously, we excluded rejected points in `point_in_polygon` that matched points that composed the comparison polygon, but if a point was colinear with a pair of points on the edge of the polygon, that point would be included. This PR adds those tests, and also a fix for `point_in_polygon.cuh` that adds boundary exclusion for colinear points.

### DIFF
--- a/cpp/src/utility/point_in_polygon.cuh
+++ b/cpp/src/utility/point_in_polygon.cuh
@@ -56,7 +56,15 @@ inline __device__ bool is_point_in_polygon(T const x,
       T rise               = y1 - y0;
       T rise_to_point      = y - y0;
 
-      if (y_in_bounds && x < (run / rise) * rise_to_point + x0) { in_polygon = not in_polygon; }
+      // colinearity test
+      T d           = (x1 - x0) * (x1 - x0) + (y1 - y0) * (y1 - y0);
+      T d1          = (x - x0) * (x - x1) + (y - y0) * (y - y1);
+      T d2          = (x1 - x) * (x1 - x) + (y1 - y) * (y1 - y);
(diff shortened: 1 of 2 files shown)


thomcom:
Closes #738

thomcom:
Before this is merged I'd like to run python benchmarks against the change.

cpp/src/utility/point_in_polygon.cuh
```diff
@@ -56,7 +56,15 @@ inline __device__ bool is_point_in_polygon(T const x,
...
-      if (y_in_bounds && x < (run / rise) * rise_to_point + x0) { in_polygon = not in_polygon; }
+      // colinearity test
+      T d           = (x1 - x0) * (x1 - x0) + (y1 - y0) * (y1 - y0);
+      T d1          = (x - x0) * (x - x1) + (y - y0) * (y - y1);
+      T d2          = (x1 - x) * (x1 - x) + (y1 - y) * (y1 - y);
+      bool colinear = d1 + d2 == d;
+
+      if (!colinear && y_in_bounds && x < (run / rise) * rise_to_point + x0) {
+        in_polygon = not in_polygon;
+      }
```
harrism, line 67: This is a lot of additional computation. @isVoid since you recently updated the algorithm used here, can you check if there's anything about excluding boundaries in the reference paper from Eric Haines?

thomcom:
I posted the results of the benchmarking in #742. I definitely feel some discussion and possibly further verification is necessary before we merge this.

thomcom:
Wrong file.

PR #741, opened 2022-10-18
Header-only trajectory_bounding_boxes

harrism:
Contributes to #563 

Header-only refactoring of `cuspatial::trajectory_bounding_boxes`

Note the commit list is messy because it was originally based off of #628 which was merged into 22.10.

## Checklist


cpp/src/trajectory/trajectory_bounding_boxes.cu
```diff
@@ -66,40 +64,19 @@ struct dispatch_element {
...
-      });
+    auto points_begin = cuspatial::make_vec_2d_iterator(x.begin<T>(), y.begin<T>());
+
+    auto bbox_mins  = cuspatial::make_vec_2d_output_iterator(cols.at(0)->mutable_view().begin<T>(),
+                                                            cols.at(1)->mutable_view().begin<T>());
+    auto bbox_maxes = cuspatial::make_vec_2d_output_iterator(cols.at(2)->mutable_view().begin<T>(),
+                                                             cols.at(3)->mutable_view().begin<T>());
 
+    trajectory_bounding_boxes(object_id.begin<std::int32_t>(),
+                              object_id.end<std::int32_t>(),
```
isVoid, line 75: Should we use `cudf::size_type` here?

cpp/include/cuspatial/experimental/trajectory_bounding_boxes.cuh
```diff
@@ -0,0 +1,61 @@
...
+ * @return A `std::pair` of iterators to the ends of the ranges of output minimum and maximum
+ *         bounding box coordinates.
+ */
+template <typename IdInputIt, typename PointInputIt, typename PointOutputIt>
+std::pair<PointOutputIt, PointOutputIt> trajectory_bounding_boxes(
+  IdInputIt ids_first,
+  IdInputIt ids_last,
+  PointInputIt points_first,
+  PointOutputIt bounding_box_minima_first,
+  PointOutputIt bounding_box_maxima_first,
```
isVoid, line 56: I wonder if we should let the algorithm only accept a single output iterator and leave user the burden to construct the zipped output iterator before they pass it in. I think it's more consistent with thrust's algorithm design.

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -132,59 +186,34 @@ struct trajectory_test_data {
...
-
-    using namespace std::chrono_literals;
-
-    thrust::tabulate(rmm::exec_policy(),
-                     ids_sorted.begin(),
-                     ids_sorted.end(),
-                     id_functor(offsets.data().get(), offsets.data().get() + offsets.size()));
+  struct box_minmax {
+    using point_tuple = thrust::tuple<cuspatial::vec_2d<T>, cuspatial::vec_2d<T>>;
+    __host__ __device__ point_tuple operator()(point_tuple const& a, point_tuple const& b)
```
isVoid, line 191: This should be host use only?

isVoid (commented):
Mostly just small questions.

cpp/src/trajectory/trajectory_bounding_boxes.cu
harrism: I guess so. But the runtime check on line 117 is `cudf::type_id::INT32`. How do we ensure that it is `cudf::size_type`? What if size_type changes?

cpp/include/cuspatial/experimental/trajectory_bounding_boxes.cuh
```diff
@@ -0,0 +1,61 @@
...
+ * @param points_first beginning of the range of input point (x,y) coordinates
+ * @param bounding_box_minima_first beginning of the range of output minimum bounding box
+ *                                  coordinates, one per trajectory
+ * @param bounding_box_maxima_first beginning of the range of output maximum bounding box
+ *                                  coordinates, one per trajectory
+ * @param stream the CUDA stream on which to perform computations.
+ *
+ * @return A `std::pair` of iterators to the ends of the ranges of output minimum and maximum
+ *         bounding box coordinates.
+ */
```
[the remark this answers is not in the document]
harrism: Done.

cpp/include/cuspatial/experimental/trajectory_bounding_boxes.cuh
harrism: I looked at this and it's a bit tricky. This is a case where there are two outputs: the minimum and the maximum vec_2d point of each box. I can't think of a single example where Thrust expects a single zip iterator for algorithms that naturally produce more than one output.  Also, using this from the column-based API already requires zipping the min_x and min_y and zipping the  max_x and max_y columns together to output the 2D extrema. Zipping the extrema together would require the column-based API to do a 2-level zip. I also have to specify what type the extrema box is -- is it a pair? a tuple? or a new box struct?

I see three options.

1. Leave it as I have it now.
2. Do as you suggest, resulting in a 2-level zip in the column-based API, and single-level zips in header-only tests.
3. Introduce a new box struct type that is used in the header-only API. This would change the column-based API to use a 4-element zip (but single level).

I lean toward option 1 or 3. I don't like option 2.

cpp/include/cuspatial/experimental/trajectory_bounding_boxes.cuh
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <rmm/cuda_stream_view.hpp>
+
+namespace cuspatial {
+
```
harrism, line 22: Need doxygen group here.

cpp/src/trajectory/trajectory_bounding_boxes.cu
isVoid: I believe the type check should be `... == type_to_id<cudf::size_type>()`.

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -132,59 +186,34 @@ struct trajectory_test_data {
...
-                      ids_sorted.begin(),
-                      points_sorted.begin(),
-                      point_functor());
+    auto point_tuples = thrust::make_zip_iterator(points_sorted.begin(), points_sorted.begin());
 
-    // shuffle input data to create randomized order
-    rmm::device_vector<std::int32_t> map(total_points);
-    thrust::sequence(rmm::exec_policy(), map.begin(), map.end(), 0);
-    thrust::shuffle(rmm::exec_policy(), map.begin(), map.end(), gen);
+    thrust::reduce_by_key(ids_sorted.begin(),
```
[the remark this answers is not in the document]
isVoid: Ok. Reading a few guidelines for unit test I don't really see any thing that says against using test code from the core logic. Seeing the original API is a black box is fine and doing it this way it always makes sure the black box will behave according to what's done in the test code. At worst this avoids future regression (caused by changes in cuspatial code domain).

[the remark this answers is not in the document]
harrism: I view it as coincidental that the test reference implementation is the same as the library implementation.

Ideally we should test against a 3rd-party reference but that's difficult in C++ tests. Perhaps we could do that in the Python tests.

cpp/include/cuspatial/experimental/trajectory_bounding_boxes.cuh
harrism: Michael convinced me to do #2...

isVoid (approved):

trxcllnt (approved):

PR #750, opened 2022-10-24
Create `pairwise_point_in_polygon` to be used by pairwise `GeoSeries`

thomcom:
## Description
This PR provides two primary outcomes: it adds a `pairwise_point_in_polygon` API in the header-only and cudf-backed C++ functionality, and it modifies the `is_point_in_polygon` kernel to perform boundary exclusion.

The `pairwise_point_in_polygon` API has been added because `GeoPandas.GeoSeries.contains` only supports a pairwise operation and a many-to-one operation, but does not implement the all-pairs approach to point-in-polygon we developed.

Boundary exclusion has been added for two reasons: The main reason is because `GeoPandas.GeoSeries.contains` applies boundary exclusion: points that are in the boundary of a polygon are not considered to be "in" the polygon. The other reason to add boundary exclusion is because our results were previously undefined. Some boundary points were excluded in point-in-polygon, and others were not. This PR resolves that inconsistency.

## Checklist


cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
```diff
@@ -0,0 +1,382 @@
...
+                                                    {0.5, 0.0},
+                                                    {0.0, -0.5},
+                                                    {0.0, 0.5}};
+  auto poly_offsets      = this->make_device_offsets({0});
+  auto poly_ring_offsets = this->make_device_offsets({0});
+  auto poly_point =
+    this->make_device_points({{-1.0, -1.0}, {1.0, -1.0}, {1.0, 1.0}, {-1.0, 1.0}, {-1.0, -1.0}});
+
+  auto got      = rmm::device_vector<int32_t>(1);
+  auto expected = std::vector<int32_t>{false, false, false, false, true, true, true, true};
```
isVoid, line 66: This should be a boolean array.

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
```diff
@@ -0,0 +1,382 @@
...
+                                         poly_offsets.begin(),
+                                         poly_offsets.end(),
+                                         poly_ring_offsets.begin(),
+                                         poly_ring_offsets.end(),
+                                         poly_point.begin(),
+                                         poly_point.end(),
+                                         got.begin());
+    EXPECT_EQ(got, std::vector<int32_t>({expected[i]}));
+    EXPECT_EQ(ret, got.end());
+  }
```
isVoid, line 81: Ah ok. I see how you are computing these results 1 by 1. I think you can wrap the iterators of input point arrays with [thrust::permulation_iterator](https://nvidia.github.io/thrust/api/classes/classthrust_1_1permutation__iterator.html#class-thrustpermutation_iterator) and set the offsets accordingly to make 8 polygons without duplicate code.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,137 @@
...
+                                                 OffsetItADiffType const num_polys,
+                                                 OffsetIteratorB ring_offsets_first,
+                                                 OffsetItBDiffType const num_rings,
+                                                 Cart2dItB poly_points_first,
+                                                 Cart2dItBDiffType const num_poly_points,
+                                                 OutputIt result)
+{
+  using Cart2d     = iterator_value_type<Cart2dItA>;
+  using OffsetType = iterator_value_type<OffsetIteratorA>;
+  auto idx         = blockIdx.x * blockDim.x + threadIdx.x;
```
isVoid, line 55: We should switch to using grid-stride loop here.

isVoid (changes requested):
I think we should add a long input test to have more point/polygon points than the size of a grid.

harrism (changes requested):
Need an epsilon when comparing to zero...

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * @param[in] test_points_y:     y-coordinates of points to test
+ * @param[in] poly_offsets:      beginning index of the first ring in each polygon
+ * @param[in] poly_ring_offsets: beginning index of the first point in each ring
+ * @param[in] poly_points_x:     x-coordinates of polygon points
+ * @param[in] poly_points_y:     y-coordinates of polygon points
+ *
+ * @returns A column of booleans for each point/polygon pair.
+ *
+ * @note Direction of rings does not matter.
+ * @note Supports open or closed polygon formats.
```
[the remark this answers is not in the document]
thomcom: Does support, we talked about this yesterday.

cpp/include/cuspatial/experimental/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,112 @@
...
+ * @ingroup spatial_relationship
+ *
+ * @brief Tests whether the specified points are inside their corresponding polygon.
+ *
+ * Tests whether each point is inside a corresponding polygon.  Polygons are a collection of one or
+ * more rings. Rings are a collection of three or more vertices.
+ *
+ * Each input point will map to one `int32_t` element in the output. Each bit (except the sign bit)
+ * represents a hit or miss for each of the input polygons in least-significant-bit order. i.e.
+ * `output[3] & 0b0010` indicates a hit or miss for the 3rd point against the 2nd polygon.
```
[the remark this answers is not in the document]
thomcom: Yes missed comment thanks.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
```diff
@@ -0,0 +1,102 @@
...
+
+      // Points on the line segment are the same, so intersection is impossible.
+      // This is possible because we allow closed or unclosed polygons.
+      if (run == 0.0 && rise == 0.0) continue;
+
+      T rise_to_point = test_point.y - a.y;
+
+      // colinearity test
+      T run_to_point = test_point.x - a.x;
+      is_colinear    = (run * rise_to_point - run_to_point * rise) == 0;
```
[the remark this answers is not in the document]
thomcom: Thanks for those corner cases. I'll be sure to check them against GeoPandas and add an epsilon if they handle those cases better than us.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
```diff
@@ -0,0 +1,102 @@
...
+      if (is_colinear) { break; }
+
+      y1_flag = a.y > test_point.y;
+      if (y1_flag != y0_flag) {
+        // Transform the following inequality to avoid division
+        //  test_point.x < (run / rise) * rise_to_point + a.x
+        auto lhs = (test_point.x - a.x) * rise;
+        auto rhs = run * rise_to_point;
+        if ((rise > 0 && lhs < rhs) || (rise < 0 && lhs > rhs))
```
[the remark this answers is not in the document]
thomcom: It works :)

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
thomcom: GeoPandas has inconsistent results along the boundary as well. I'm not sure how much time we should spend chasing floating point error, what do you recommend?

I wrote the following test for GeoPandas and cuSpatial:

```
@pytest.mark.parametrize(
    "point, polygon, expects",
    [
        # unique failure cases identified by @mharris
        [Point([0.6, 0.06]), Polygon([[0, 0], [10, 1], [1, 1], [0, 0]]), False],
        [Point([3.3, 1.1]), Polygon([[6, 2], [3, 1], [3, 4], [6, 2]]), True],
    ],
)
def test_float_precision_limits(point, polygon, expects):
    gpdpoint = gpd.GeoSeries(point)
    gpdpolygon = gpd.GeoSeries(polygon)
    point = cuspatial.from_geopandas(gpdpoint)
    polygon = cuspatial.from_geopandas(gpdpolygon)
    result = polygon.contains(point)
    assert gpdpolygon.contains(gpdpoint).values == result.values_host
    assert result.values_host[0] == expects
```

The third value in the `parametrize` statement is the value that is "expected", that is, the one that GeoPandas returns. The first thing to notice is that GeoPandas (GEOS) doesn't successfully identify these special cases both as being in the boundary, the first is a boundary case and the second is not.

I was able to identify EPSILON values to use for the `is_colinear` test and for the final intersection test (different epsilons are required because `is_colinear` needs to be squared to use as an inequality, whereas the intersection test is already an inequality so EPSILON can simply be added or subtracted.

After I identified working EPSILON values for these specific cases, I expanded them to higher precision examples:

```
        [Point([0.66, 0.006]), Polygon([[0, 0], [10, 1], [1, 1], [0, 0]]), False],
        [Point([0.666, 0.0006]), Polygon([[0, 0], [10, 1], [1, 1], [0, 0]]), False],
        [Point([3.33, 1.11]), Polygon([[6, 2], [3, 1], [3, 4], [6, 2]]), True],
        [Point([3.333, 1.111]), Polygon([[6, 2], [3, 1], [3, 4], [6, 2]]), True],
```

At this point GeoPandas's consistency breaks down. The first two work as expected. The third breaks my epsilon test, suggesting perhaps further EPSILON refinement. But the fourth example breaks GeoPandas's consistency: `[3.3, 1.1]` and `[3.33, 1.11` are "in" the polygon, but `[3.333, 1.111]` is not. It looks like GEOS uses a similar [RayCrossingCounter](https://github.com/libgeos/geos/blob/main/src/algorithm/RayCrossingCounter.cpp) with a complex object oriented design an no epsilons that I can see.

I'm inclined to drop epsilons and accept that boundary cases are fuzzy and prone to error. What do you think?

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * :██████:    :██████:   :██:  :████:  :████:  :██:
+ * +------;----+██████:   :██:  :----:  :----:  :██:
+ *        :███████████:   :██+------------------+██:
+ *        :███████████:   :████████████████████████:
+ *        +-----------+   +------------------------+
+ * ```
+ */
+std::unique_ptr<cudf::column> pairwise_point_in_polygon(
+  cudf::column_view const& test_points_x,
+  cudf::column_view const& test_points_y,
```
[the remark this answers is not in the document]
thomcom: That's a good question. GeoPandas and others are talking frequently about dropping the xy interleave format, or supporting both. I'm not sure how hard we should push to keep them interleaved. https://github.com/geoarrow/geoarrow/pull/26

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
```diff
@@ -0,0 +1,382 @@
...
+TYPED_TEST(PairwisePointInPolygonTest, OnePolygonOneRing)
+{
+  auto point_list        = std::vector<std::vector<float>>{{-2.0, 0.0},
+                                                    {2.0, 0.0},
+                                                    {0.0, -2.0},
+                                                    {0.0, 2.0},
+                                                    {-0.5, 0.0},
+                                                    {0.5, 0.0},
+                                                    {0.0, -0.5},
+                                                    {0.0, 0.5}};
```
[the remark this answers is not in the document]
thomcom: Code says I need `this->template make_device_points<vec_2d<TypeParam>>` which doesn't build. Would this be an iterable container? I just need to get the points iterable so I can test each pair of point/polygon one at a time.

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
```diff
@@ -0,0 +1,382 @@
...
+  auto poly_offsets      = this->make_device_offsets({0});
+  auto poly_ring_offsets = this->make_device_offsets({0});
```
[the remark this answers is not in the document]
thomcom: It looks like `point_in_polygon` hasn't been updated yet to `GeoArrow` format - the last index is removed at the python layer before using there, and if I include it here then it breaks tests. Can I postpone that refactor at this time?

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
harrism: @thomcom these are not corner cases. I'm saying that the approach is flawed. You can't compare directly to zero and expect consistent results. The examples I gave simply demonstrate that some decimal numbers are not representable in binary floating point. That was just an easy way to show the flaw in the `is_colinear` approach.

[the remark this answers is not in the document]
thomcom: Specifying Epsilon will probably just have different (fewer, probably) failure cases. The original algorithm also ignores epsilon as well for these kinds of cases and pip. I think I prefer leaving the epsilon out and just clarifying in the docs that floating point error may occur. I think it is implicit, just like with osGeo/GEOS having these kinds of cases in GeoPandas, just different ones than ours. I understand that strict equality is going to fail, doesn't everyone? :D @isVoid how do you feel about it after the discussion we had?

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
```diff
@@ -0,0 +1,113 @@
+/*
```
harrism, line 1: Suggest removing the name "kernel" from this file since it doesn't contain a kernel.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
```diff
@@ -0,0 +1,113 @@
...
+      y1_flag = a.y > test_point.y;
+      printf("Can intersect? : %s\n", y1_flag != y0_flag ? "True" : "False");
+      if (y1_flag != y0_flag) {
+        // Transform the following inequality to avoid division
+        //  test_point.x < (run / rise) * rise_to_point + a.x
+        auto lhs                  = (test_point.x - a.x) * rise;
+        auto rhs                  = run * rise_to_point;
+        const T INTERSECT_EPSILON = 0.00000000000001;
+        printf("lhs - rhs: %.60lf\n", lhs - rhs);
+        if (lhs - INTERSECT_EPSILON < rhs != y1_flag) {
```
harrism, line 94: I don't see why you need to add epsilons everywhere when the only concern was about the new colinearity check because it looks for exact equality. 

My main concern is that the colinearity test is basically doing the same math that the intersection check is already doing, just replacing inequality with equality. This is what led me to believe there must be a formulation of this algorithm which excludes boundaries more naturally.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,137 @@
...
+template <class Cart2dItA,
+          class Cart2dItB,
+          class OffsetIteratorA,
+          class OffsetIteratorB,
+          class OutputIt,
+          class Cart2dItADiffType = typename std::iterator_traits<Cart2dItA>::difference_type,
+          class Cart2dItBDiffType = typename std::iterator_traits<Cart2dItB>::difference_type,
+          class OffsetItADiffType = typename std::iterator_traits<OffsetIteratorA>::difference_type,
+          class OffsetItBDiffType = typename std::iterator_traits<OffsetIteratorB>::difference_type>
+__global__ void pairwise_point_in_polygon_kernel(Cart2dItA test_points_first,
```
harrism, line 43: Is pairwise PiP going to be a common case for "Big Data"? I suspect that the common case will be N points, M polys, where N >> M.  In that case, pairwise won't work. And the parallelism model used here won't be efficient for pairwise cases unless N is very large.

harrism (commented):
Other than the experimental epsilon code with printfs and my question about the utility of pairwise PiP for big data, looks good.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
isVoid: yes!

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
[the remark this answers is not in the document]
isVoid: @harrism left a convincing example that kernel does benefit from interleaved format. And I think the direction is moving to support both in the spec.

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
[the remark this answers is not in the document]
isVoid: Sure. It's a task in #649

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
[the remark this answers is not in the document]
isVoid: I later realized you need a host vector. Nevermind.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
isVoid: > I'm inclined to drop epsilons and accept that boundary cases are fuzzy and prone to error. What do you think?

As long as this fits your need for DE-9IM, that's fine. I'm trying to help you avoid corner cases in the future where some geometry lies very close to a polygon and gets misclassified as `contains`/`intersects`.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
isVoid: If you have excluded all colinear cases, then the ray crossings test here is non-ambiguous. Introducing extra epsilon here is introducing unnecessary bias into the equation.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
```diff
@@ -0,0 +1,113 @@
...
+      T rise         = b.y - a.y;
+
+      // Points on the line segment are the same, so intersection is impossible.
+      // This is possible because we allow closed or unclosed polygons.
+      if (run == 0.0 && rise == 0.0) continue;
+
+      T rise_to_point = test_point.y - a.y;
+
+      // colinearity test
+      const T EPSILON = 0.0000000000000000000000000000001;
```
isVoid, line 77: The one last thing I have with is that the epsilon value should be machine/precision dependent. Maybe a multiples of `numeric_limit<T>::epsilon()` or multiples of ULP.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
isVoid: Pairwise PIP is always N points and N polys?

isVoid (commented):
Last few comments

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
```diff
@@ -0,0 +1,113 @@
...
+      T run_to_point  = test_point.x - a.x;
+      T colinearity   = (run * rise_to_point - run_to_point * rise);
+      is_colinear     = colinearity * colinearity < EPSILON;
```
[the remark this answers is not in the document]
harrism: Why does it need to be squared, and why take the abs if it is squared?

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
harrism: Unscaled, `numeric_limits<T>::epsilon()` is by definition 1 ULP, isn't it? So should probably be like `k * numeric_limits<T>::epsilon()` for some small k.  Normally you would scale by the magnitude of the value being compared, but since that value is assumed to be close to zero, there's no need.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
harrism: Yes, that's my point. Unless N is large enough, it will underutilize the GPU.  I just couldn't see any existing use of geopandas having a very large number of polygons.  Hence this would be a case where mimicking geopandas is questionable.

harrism: I think a common case with more data would be pairwise *points* in polygon, where we test multiple points inside each polygon. And in this case, it might be a bad idea to use one thread per pair rather than one thread per point.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
isVoid: Oops, yeah it's squared. So no abs, +1 question for why sqaured?

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
isVoid: Eventually this PR will be refactored with the new range objects. And hopefully we should be able to get multipoints/multipolys for free. For now it's probably ok to leave it as is?

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
isVoid: > Unscaled, numeric_limits<T>::epsilon() is by definition 1 ULP, isn't it? 
Yes. Suggest `k=4` to align with gtest?

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
isVoid: I agree that multipoint->polygon is probably the more commonly used API. But it's probably a stretch for what this PR is trying to achieve, especially we are only looking to support a subset of binary predicates which is only single-single geometry related.

thomcom: I wrote this while working on `polygon.contains(points)` and as I moved on to linestring and polygons I remembered that pairwise doesn't apply for that case. I'm not sure of the usefulness of this PR, except for the colinearity test which I could easily separate out.

thomcom: Doing the M x N point-in-polygon seems wasteful for doing `polygon.contains(points)` but adding the code complexity may not be worth the inefficiency trade-off.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
thomcom: It needs to be squared because it can be naturally negative, which requires that I make it positive in order to test against epsilon.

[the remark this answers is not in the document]
isVoid: Oh so squared is for negative. Perhaps using `abs` instead can squeeze a few cycles?

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
isVoid: In a long run, I was hoping to expose the precision issue as a user configurable option. As this issue will soon come up else as we encounters more complicated ops.

harrism: File an issue for that, please.   We may want a way to specify global settings like this, if we don't want to add a precision argument to every API...

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
harrism: `polygon.contains(points)` presumably checks all points inside a single polygon, so it doesn't matter if it's MxN since M == 1. Or am I missing something?

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
thomcom: Done.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
thomcom: I assumed that squaring would be faster than `abs` but I really don't know. `abs` is a little cleaner and less susceptible to floating point error since small errors get much smaller when squared.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
thomcom: The other use case is `Mpolygons.contains(Mpoints)`. In the pairwise implementation this tests each point $i$ for membership in each polygon $i$. In our existing implementation this tests all points $0..M$ for membership in each polygon $0..M$, then for this algorithm discards all the results except for $M_i,i$. The pairwise operation is $O(M)$ and the all-pairs operation is $O(M^2)$, lots more expensive. I don't know how common this use case is, or if the pairwise implementation will be sufficiently faster than the quadtree implementation mess with. I'm still researching that.

harrism: OK, thanks.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon_kernel.cuh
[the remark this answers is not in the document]
harrism: Yes, abs is unquestionably the right thing here. Less error, faster, doesn't impact the relative size of your epsilon, etc.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
thomcom: Done

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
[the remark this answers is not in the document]
thomcom: This seems like a huge refactor, won't basically every file in the chain need to be modified to reduce from two iterators to one and offset?

harrism:
rerun tests

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
```diff
@@ -0,0 +1,102 @@
...
+      // Points on the line segment are the same, so intersection is impossible.
+      // This is possible because we allow closed or unclosed polygons.
+      if (run == 0.0 && rise == 0.0) continue;
+
+      T rise_to_point = test_point.y - a.y;
+
+      // colinearity test
+      T run_to_point = test_point.x - a.x;
+      T colinearity  = (run * rise_to_point - run_to_point * rise);
+      is_colinear    = abs(colinearity) < std::numeric_limits<T>::epsilon();
```
harrism, line 79: Can you change this to use `float_equal` from #773 ? (now merged).

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
```diff
@@ -0,0 +1,102 @@
...
+    bool y1_flag;
+    // for each line segment, including the segment between the last and first vertex
+    for (auto point_idx = ring_begin; point_idx < ring_end; point_idx++) {
+      Cart2d const a = poly_points_first[point_idx];
+      T run          = b.x - a.x;
+      T rise         = b.y - a.y;
+
+      // Points on the line segment are the same, so intersection is impossible.
+      // This is possible because we allow closed or unclosed polygons.
+      if (run == 0.0 && rise == 0.0) continue;
```
harrism, line 72: Also use `float_equal` for these.
```suggestion
      if (float_equal(run, 0.0) && float_equal(rise, 0.0)) continue;
```

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
```diff
@@ -0,0 +1,102 @@
...
+
+      T rise_to_point = test_point.y - a.y;
+
+      // colinearity test
+      T run_to_point = test_point.x - a.x;
+      T colinearity  = (run * rise_to_point - run_to_point * rise);
+      is_colinear    = abs(colinearity) < std::numeric_limits<T>::epsilon();
+      if (is_colinear) { break; }
```
harrism, line 80:
```suggestion
      if (float_equal(run * rise_to_point - run_to_point * rise, 0.0)) { break; } // exclude colinear points
```

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace cuspatial {
+
+/**
+ * @addtogroup spatial_relationship
+ * @{
+ */
+
+/**
+ * @brief Tests whether the specified points are inside each of their corresponding polygons.
```
harrism, line 35:
```suggestion
 * @brief Given (point, polygon pairs), tests whether the point of each pair is inside the polygon of the pair.
```

cpp/include/cuspatial/experimental/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,112 @@
...
+#pragma once
+
+#include <rmm/cuda_stream_view.hpp>
+
+namespace cuspatial {
+
+/**
+ * @ingroup spatial_relationship
+ *
+ * @brief Tests whether the specified points are inside their corresponding polygon.
```
harrism, line 26:
```suggestion
 * @brief Given (point, polygon) pairs, tests whether the point of each pair is inside the polygon of the pair.
```

cpp/include/cuspatial/experimental/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * Tests whether each point is inside a corresponding polygon.  Points on the edges of the
+ * polygon are not considered to be inside. Floating point precision limitations mean that
+ * some arithmetically excluded boundary points will not be excluded, and vice-versa.
```
harrism, line 30: It's not necessary to explain the existence of floating-point error.

```suggestion
 * polygon are not considered to be inside.
```

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,138 @@
...
```
harrism, line 43: Since this uses trivial linear indexing, this should be a device functor instead of a kernel, and call it from a `thrust::transform()` rather than launching a kernel. This way cuSpatial doesn't have to handle grid sizing, grid stride loops, etc.

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
```diff
@@ -0,0 +1,385 @@
...
+                                                {0.5, 0.0},
+                                                {0.0, -0.5},
+                                                {0.0, 0.5}};
+  auto poly_offsets      = this->make_device_offsets({0});
+  auto poly_ring_offsets = this->make_device_offsets({0});
+  auto poly_point =
+    this->make_device_points({{-1.0, -1.0}, {1.0, -1.0}, {1.0, 1.0}, {-1.0, 1.0}, {-1.0, -1.0}});
+
+  auto got      = rmm::device_vector<int32_t>(1);
+  auto expected = std::vector<bool>{false, false, false, false, true, true, true, true};
```
harrism, line 67: Avoid `vector<bool>`. https://isocpp.org/blog/2012/11/on-vectorbool

In tests where you compare against device data, just use a `thrust::device_vector` or `thrust::host_vector`.

This may be why these tests are failing in CI?

harrism:
```suggestion
  auto expected = thrust::host_vector<bool>{false, false, false, false, true, true, true, true};
```

harrism (changes requested):

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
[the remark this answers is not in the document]
thomcom: I'd like to avoid this refactor for this PR and get this merged for the current sprint.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
thomcom: Done.

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
thomcom: Done

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
thomcom: Done

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
thomcom: yikes!

thomcom: `thrust::host_vector` doesn't have the same constructor options as `std::vector`, I can't pass in a list of bools. How do you feel about me using `std::vector<int>` instead?

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
thomcom: This sounds like a big refactor. `pairwise_point_in_polygon.cuh` uses the same implementation of `is_point_in_polygon.cuh` as `point_in_polygon.cuh`.

```
__device__ inline bool is_point_in_polygon(Cart2d const& test_point,
                                           OffsetType poly_begin,
                                           OffsetType poly_end,
                                           OffsetIterator ring_offsets_first,
                                           OffsetItDiffType const& num_rings,
                                           Cart2dIt poly_points_first,
                                           Cart2dItDiffType const& num_poly_points)
                                           ```
I could be wrong, but because it requires the (multi-)polygon iterator, and rings, and points of the polygon, it doesn't seem like it would be straightforward to write a device functor to do the same thing. Like I said I could be wrong. I am assuming that you mean for me to remove the grid stride loop and instead use `thrust::transform` to do the work like:

```

  thrust::transform(rmm::exec_policy(stream),
                    test_points_first,
                    test_points_last,
                    polygon_offsets_first,
                    output,
                    [polygon_offsets_first,
                     polygon_offsets_last,
                     poly_ring_offsets_first,
                     polygon_points_first] __device__(auto const& p) {
                      is_point_in_polygon(p,
                                          polygon_offsets_first,
                                          polygon_offsets_last,
                                          poly_ring_offsets_first,
                                          num_rings,
                                          polygon_points_first,
                                          num_poly_points);
                    });
                    ```
Obviously that function doesn't work as it is written, and would require some complex thrust combinations it seems in order to combine all of the polygon components into a nested iterator that then can be passed to `thrust::transform`, unpacked, and used in `is_point_in_polygon`. Am I in the right direction here?

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
thomcom: Using `std::vector<int>`

cpp/include/cuspatial/pairwise_point_in_polygon.hpp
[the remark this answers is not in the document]
isVoid: Yes that's fine.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
harrism: Issue created to cover this: https://github.com/rapidsai/cuspatial/issues/822

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
harrism: Issue created for this: https://github.com/rapidsai/cuspatial/issues/823

cpp/include/cuspatial/experimental/detail/is_point_in_polygon.cuh
```diff
@@ -0,0 +1,104 @@
...
+    bool y1_flag;
+    // for each line segment, including the segment between the last and first vertex
+    for (auto point_idx = ring_begin; point_idx < ring_end; point_idx++) {
+      Cart2d const a = poly_points_first[point_idx];
+      T run          = b.x - a.x;
+      T rise         = b.y - a.y;
+
+      // Points on the line segment are the same, so intersection is impossible.
+      // This is possible because we allow closed or unclosed polygons.
+      T zero = 0.0;
```
isVoid, line 74:
```suggestion
      T constexpr zero = 0.0;
```

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
isVoid: This can also benefit a lot once we have `multipolygon_range`.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,139 @@
...
+                                                 OffsetItBDiffType const num_rings,
+                                                 Cart2dItB poly_points_first,
+                                                 Cart2dItBDiffType const num_poly_points,
+                                                 OutputIt result)
+{
+  using Cart2d     = iterator_value_type<Cart2dItA>;
+  using OffsetType = iterator_value_type<OffsetIteratorA>;
+  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x;
+       idx <
+       std::distance(test_points_first, thrust::prev(test_points_first + num_test_points + 1));
```
harrism, line 57: Seems silly to use `prev` with pointer arithmetic. The point of `prev` and `next` is to abstract iterator/pointer arithmetic. The `prev` negates the `+ 1`, so this is basically doing `+ 1 - 1`.  I would change to:

```
       std::distance(test_points_first, std::next(test_points_first, num_test_points));
```

Or

```
 std::distance(test_points_first, test_points_first + num_test_points);
```

harrism (approved):

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,139 @@
...
+       idx += gridDim.x * blockDim.x) {
```
isVoid, line 58:
```suggestion
  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < num_test_points; idx += gridDim.x * blockDim.x) {
```

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
```diff
@@ -0,0 +1,139 @@
...
+  CUSPATIAL_EXPECTS(num_poly_points >= num_polys * 4, "Each ring must have at least four vertices");
+
+  CUSPATIAL_EXPECTS(num_test_points == num_polys,
+                    "Must pass in an equal number of points and polygons");
+
+  // TODO: introduce a validation function that checks the rings of the polygon are
+  // actually closed. (i.e. the first and last vertices are the same)
+
+  auto constexpr block_size = 256;
+  auto const num_blocks     = (num_test_points + block_size - 1) / block_size;
```
isVoid, line 122: You can now use grid_1d in cuspatial!

```suggestion
  auto [threads_per_block, num_blocks] = grid_id(num_test_points);
```

You should `#include <cuspatial/cuda_utils.cuh>`

cpp/tests/experimental/spatial/pairwise_point_in_polygon_test.cu
```diff
@@ -0,0 +1,385 @@
...
+ public:
+  rmm::device_vector<vec_2d<T>> make_device_points(std::initializer_list<vec_2d<T>> pts)
+  {
+    return rmm::device_vector<vec_2d<T>>(pts.begin(), pts.end());
+  }
+
+  rmm::device_vector<std::size_t> make_device_offsets(std::initializer_list<std::size_t> pts)
+  {
+    return rmm::device_vector<std::size_t>(pts.begin(), pts.end());
+  }
```
isVoid, line 43: Future TODOs: All these can be replaced with `make_device_vector` in `vector_factories.cuh`.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
isVoid: This is just `num_test_points`. See https://github.com/rapidsai/cuspatial/pull/750#discussion_r1034154480

isVoid (approved):
See comments above, adopt if you sees fit.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
harrism: OMG I totally missed that in my comment. :)

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
thomcom: Done.

cpp/include/cuspatial/experimental/detail/pairwise_point_in_polygon.cuh
thomcom: Done

harrism:
@gpucibot merge

PR #752, opened 2022-10-26
Augment Cuspatial Test Utility to Allow User Specified Abs Error

isVoid:
## Description

This PR augments Cuspatial Vector Equality Test Utility with an extra parameter to allow user pass in pre-defined absolute error.
In many cases, cuspatial has to compare results with external libraries such as shapely or gdal. Since cuspatial allows user to pick computation precision based on input data (not a bug!), requiring the result to match external libraries within 4ULP is often too stringent, as external libraries often perform computation in different precisions. Adding the flexibility to specifiy abs error in test utilitiy makes it easy for developer to pick desired error range based on specific use cases.

This PR also introduces a test macro that shows lineno of failed tests.

## Checklist

### DIFF
--- a/cpp/tests/utility/vector_equality.hpp
+++ b/cpp/tests/utility/vector_equality.hpp
@@ -16,7 +16,6 @@
 
 #pragma once
 
-#include <cmath>
 #include <cuspatial/traits.hpp>
 #include <cuspatial/vec_2d.hpp>
 
--- a/cpp/tests/utility/vector_equality.hpp
+++ b/cpp/tests/utility/vector_equality.hpp
@@ -32,8 +32,14 @@
 namespace cuspatial {
 namespace test {
 
+/**
+ * @brief Compare two floats are close within N ULPs
+ *
+ * N is predefined by GoogleTest
+ * https://google.github.io/googletest/reference/assertions.html#EXPECT_FLOAT_EQ
+ */
 template <typename T>
-auto floating_eq(T val)
+auto floating_eq_by_ulp(T val)
 {
   if constexpr (std::is_same_v<T, float>) {
     return ::testing::FloatEq(val);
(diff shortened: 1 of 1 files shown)


cpp/tests/utility/vector_equality.hpp
```diff
@@ -69,6 +95,20 @@ MATCHER(float_matcher, std::string(negation ? "are not" : "are") + " approximate
   return false;
 }
 
+MATCHER_P(float_near_matcher,
```
harrism, line 98: The `float_matcher` is also a "near" matcher, so perhaps `float_near_matcher` is not the right name. The only different is that we pass abs_error. GTEST ignores the abs error if it is zero. So why not just augment the existing matchers with default abs_error of zero to avoid all the additional code?

isVoid: Hmm, it's not that simple.. If you default `abs_error` to 0, technically it's going through a different code path as opposed to using `FLOAT_EQ` matcher. `FLOAT_NEAR(abs_error==0) will go to this:

https://github.com/google/googletest/blob/3026483ae575e2de942db5e760cf95e973308dd5/googlemock/include/gmock/gmock-matchers.h#L1687

which will not use the ULP near equal check but completely resort to `abs_err` check.

But `FLOAT_EQ` matcher will use this:

https://github.com/google/googletest/blob/3026483ae575e2de942db5e760cf95e973308dd5/googlemock/include/gmock/gmock-matchers.h#L1706

which will use the ULP near equal check.

While we can default `abs_error` to -1, the default value where gtest use internally to differentiate. I believe that's an implementation detail and it's in nature an invalid input to the matcher.

isVoid: If just for naming, we can call it `float_eq_by_ulp` / `float_eq_by_abs_error`. As for the user facing API `expect_vector_equivalent`, I think we should keep it as is and use overloads for user to differentiate.

harrism (approved):

isVoid:
@gpucibot merge

PR #754, opened 2022-10-26
Support more than 31 polygons in point_in_polygon.

thomcom:
This PR drops the bitmask approach to performing 31 polygon point-in-polygon, instead preallocating a column of bools and storing the pip result there. This allows an unlimited number of polygons to be indexed at a time and increases performance with a minor memory usage increase.

This implementation uses 4x more bytes on 31 polygons (31 bytes instead of 1 int32_t) and is 37% faster than the bool mask approach. It supports more than 31 polygons, however, allocating a single `num_points x num_polygons` column for output.


thomcom:
Superceded by #981

PR #760, opened 2022-10-31
Allow None rows in `GeoSeries` and an `align` method to match `GeoPandas`

### LINKED ISSUE #761: [FEA] GeoSeries needs index alignment
GeoPandas implements pairwise comparisons in their binary operations. The default argument for all binary operations is `align=True` which index-aligns the GeoSeries before calling the binary operation. We need to support at least at a default argument level the align call so that we can match their default behavior.

See https://github.com/rapidsai/cuspatial/issues/761#issuecomment-1299170908 for more details.

Handle index alignment when `.contains_properly` is called.

**Describe alternatives you've considered**
Don't support index alignment.

### LINKED ISSUE #762: [FEA] GeoSeries needs to support `None` rows entries.
If the two GeoSeries that are passed in for a pairwise binary operation are not the same length, the default `align=true` argument will add `None` rows to make their lengths match. We should also support this.

Let it be possible for something like `GeoSeries([Point(0, 1), None])` to be created in cuSpatial.

**Describe alternatives you've considered**
Error instead of align when the input pairwise arguments are not the same length.

thomcom:
## Description
This PR allows us to have `GeoSeries` with empty rows containing `None`. This feature is used by GeoPandas frequently with binary operations on `GeoSeries` of different lengths. It also adds an `align` method which works with default arguments for aligning the indices of a `Series` and adding None where rows are missing.

I also updated the internals for `GeoColumn`, `GeoSeries`, and `GeoDataFrame` to handle native `_gather` operations from cudf correctly. This eliminates the need to use `column._data` as a reordering map when `sort_values` is used on a non-geometry column, and should support more hidden cudf operations that we haven't attempted yet.

Closes #762
Closes #761 

## Checklist


cpp/include/cuspatial/experimental/detail/point_in_polygon.cuh
```diff
@@ -114,7 +114,9 @@ __global__ void point_in_polygon_kernel(Cart2dItA test_points_first,
 {
   using Cart2d     = iterator_value_type<Cart2dItA>;
   using OffsetType = iterator_value_type<OffsetIteratorA>;
-
+  // grid-stride loop
+  // for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < multilinestrings1.num_points();
+  //     idx += gridDim.x * blockDim.x) {
```
harrism, line 119: What does commenting this out have to do with None rows?

python/cuspatial/cuspatial/core/_column/geometa.py
```diff
@@ -9,7 +9,10 @@
 import cudf
 
 
+# This causes arrow to encode NONE as =255, which I'll accept now
+# in order to keep the rest of the enums the same.
 class Feature_Enum(Enum):
+    NONE = -1
```
harrism, line 15: What is the difference between a NONE row and a null row? Could this just be implemented using nulls?

thomcom: Good idea I'm looking into this now.

thomcom: I think the short answer is
```
tests/test_empty.py:64: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
core/geoseries.py:353: in to_pandas
    return self.to_geopandas()
core/geoseries.py:343: in to_geopandas
    final_union_slice.to_shapely(),
core/geoseries.py:405: in to_shapely
    union = self.to_arrow()
core/geoseries.py:508: in to_arrow
    return pa.UnionArray.from_dense(
pyarrow/array.pxi:2308: in pyarrow.lib.UnionArray.from_dense
    ???
pyarrow/error.pxi:144: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   pyarrow.lib.ArrowInvalid: Union type ids may not have nulls

pyarrow/error.pxi:100: ArrowInvalid
```

null is not supported by `DenseUnion`, so I created an Enum type to store the null rows. The consquences are minor, the nulls are stored in the denseunion offsets and types and nowhere else.

harrism: Are we going to need to deal with the None type in C++?

isVoid: Out of curiosity, where does this null row come from? According to [spec](https://arrow.apache.org/docs/format/Columnar.html#union-layout):

```
Unlike other data types, unions do not have their own validity bitmap. Instead, the nullness of each slot is determined exclusively by the child arrays which are composed to create the union.
```

Do we actually have an use case where the null row is caused not by a child row being null (like a null point), but actually representing a type-less null row?

python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -178,6 +178,55 @@ def _slice(self: T, arg: slice) -> T:
...
+        )
+        result = GeoDataFrame(
+            {
+                name: (geo_columns[name] if mask else data_columns[name])
+                for name, mask in zip(
+                    column_names.values, geocolumn_mask.values
+                )
+            },
+        )
+        return result
```
isVoid, line 198: I'm not sure if this helper is needed if all you need is to make a dataframe from the dict returned from `_recombine_columns`.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -517,3 +509,60 @@ def to_arrow(self):
                 arrow_polygons,
             ],
         )
+
+    def _align_to_index(
```
isVoid, line 513: Nice. To align to the index we only need to align type codes and offset buffers. That's a lot cheaper than realigning all the children. Can you add a brief comment to this function describing this?

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -517,3 +509,60 @@ def to_arrow(self):
...
+                self._column.polygons,
+            ),
+            {
+                "input_types": aligned_input_types,
+                "union_offsets": aligned_union_offsets,
+            },
+        )
+        return GeoSeries(column)
+
+    def align(self, other):
```
isVoid, line 548: Public API requires docstring.

python/cuspatial/cuspatial/tests/test_align.py
```diff
@@ -0,0 +1,51 @@
...
+        [
+            Polygon(((-8, -8), (-8, 8), (8, 8), (8, -8))),
+            Polygon(((-2, -2), (-2, 2), (2, 2), (2, -2))),
+        ]
+    )
+    gpdshort = gpdpdf.iloc[0:1]
+    pdf = cuspatial.from_geopandas(gpdpdf)
+    short = cuspatial.from_geopandas(gpdshort)
+    gpdaligned = gpdshort.align(gpdpdf)
+    shortaligned = short.align(pdf)
```
isVoid, line 22: Can we rename the variable according to our convention? gpdaligned is the result computed from geopandas so should be `expected`, the other one is computed from cuspatial so should be `got`. And I suggest computing `expected` and `got` separately. It's hard to read atm. This applies to all following tests. Thanks.

python/cuspatial/cuspatial/tests/test_align.py
```diff
@@ -0,0 +1,51 @@
...
+    pd.testing.assert_series_equal(gpdaligned[0], shortaligned[0].to_pandas())
```
isVoid, line 23: I think it's probably doing the same thing.. But using the [test util from geopandas](https://geopandas.org/en/stable/docs/reference/api/geopandas.testing.assert_geodataframe_equal.html) might be better since it could handle corner cases specific to geographic use cases.

isVoid (changes requested):

python/cuspatial/cuspatial/tests/test_align.py
thomcom: Good idea, I like got and expected.

python/cuspatial/cuspatial/core/_column/geometa.py
thomcom: I tried setting nulls in the `union` values first, as you could see. I'll look at having valid `union` values but a null rows where it is referenced and share my work.

thomcom: It seems to be the latter case, a type-less null row. You can construct a test example like this:

```
polygons = gpd.GeoSeries([
        Polygon(((-8, -8), (-8, 8), (8, 8), (8, -8))),
        Polygon(((-2, -2), (-2, 2), (2, 2), (2, -2))),
    ])
polygon = gpd.GeoSeries(polygons[0]) 

print(polygon.contains(polygons))

print(polygon.align(polygons))
```
```
(0    POLYGON ((-8.00000 -8.00000, -8.00000 8.00000,...
1                                                 None
dtype: geometry, 0    POLYGON ((-8.00000 -8.00000, -8.00000 8.00000,...
1    POLYGON ((-2.00000 -2.00000, -2.00000 2.00000,...
dtype: geometry)
0     True
1    False
dtype: bool
```

thomcom: `.contains` takes an argument to align or not align the input series. If called with aligns=True, the default, then an empty row is created via `.align` for each row that doesn't exist in the other series. The missing row is not a none-polygon as far as I can tell, just a None row.

thomcom: It is less trivial to try to create a new value in the `DenseUnion` that is one of the four valid types, but is null itself, because a type needs to be chosen as the "null placeholder", then where in [geoseries](https://github.com/rapidsai/cuspatial/pull/760/files/9198f6c284e2797d7103782d71ad4c3661800d9a#diff-c522c9afb3364b1aed2b2589c0d0c260dbc634bc54844536b1d382cecb92bf29R531-R533) when None enums are created, they'll also need to be inserted into the child array, with new offsets generated and both arrays properly arranged to reflect that.

thomcom: It is not my intention that you will deal with the `None` type at all in C++, a null row to cuSpatial pandas will mean no-op.

python/cuspatial/cuspatial/core/geodataframe.py
thomcom: That's a good idea.

thomcom: Done.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: Done.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: Done.

thomcom: Done.

python/cuspatial/cuspatial/tests/test_align.py
thomcom: I'll try it.

python/cuspatial/cuspatial/core/_column/geometa.py
isVoid: Thank you for detailing the example! I definitely didn't think of the case of a null row from `contains`. Your current proposal suggest using a type code of `-1` and set all offset buffers to be `-1` for the null rows. I think for `-1` offset may lead to OOB access if not careful. How about storing an additional null type that maps to [`pa.null()`](https://arrow.apache.org/docs/python/generated/pyarrow.null.html#pyarrow.null) and store an extra `pa.NullArray` of length 1 and always use offset 0 for all null rows? The benefit is that allows you to treat the null rows homogeneously as other types in gather/scatter operations without the need to handle specially. Just a proposal - not saying this is definitely better than yours solution.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -517,3 +509,100 @@ def to_arrow(self):
...
+
+    def _align_to_index(
+        self: T,
+        index: ColumnLike,
+        how: str = "outer",
+        sort: bool = True,
+        allow_non_unique: bool = False,
+    ) -> T:
+        # The values in the newly aligned columns will not change,
+        # only their positions in the union offsets.
```
isVoid, line 521: Use `"""` for function wide comments
```suggestion
        """
        The values in the newly aligned columns will not change,
        only positions in the union offsets and type codes.
        """
```

python/cuspatial/cuspatial/io/geopandas_reader.py
```diff
@@ -30,6 +30,10 @@ def parse_geometries(geoseries: gpGeoSeries) -> tuple:
     polygon_offsets = [0]
 
     for geom in geoseries:
+        if geom is None:
+            all_offsets.append(-1)
```
isVoid, line 34: Should we define -1 as a constant?

isVoid (approved):
Thanks, lgtm.

Pending one (non-blocking) question about design here: https://github.com/rapidsai/cuspatial/pull/760#discussion_r1013183367

python/cuspatial/cuspatial/core/_column/geometa.py
thomcom: I've put a few hours into attempting this alternative. It segfaults often when I am trying to access the `DenseUnion` null array at position 0, I can't seem to keep a value inside of it. Not only is it difficult to develop for because of the segfaults but it is much more expansive of a modification, with 6+ files including the geopandas adapters needing various changes to add the null buffer and null values. I'm inclined to keep this refactoring as is as it has a relatively minor amount of code modifications. We can look at the alternative as it becomes necessary, if so.

thomcom: You can play with it [here](https://github.com/rapidsai/cuspatial/pull/780) today if you want, but I'd like to get this merged soon if not.

python/cuspatial/cuspatial/io/geopandas_reader.py
thomcom: Sure, I can just use `Feature_Enum.NONE.value`

harrism:
rerun tests

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -517,3 +509,102 @@ def to_arrow(self):
...
+        >>> polygon.align(polygons)
+        (0    POINT (-8.00000 -8.00000)
+        1                         None
+        dtype: geometry, 0    POINT (-8.00000 -8.00000)
+        1    POINT (-2.00000 -2.00000)
+        dtype: geometry)
+
+        >>> polygons_right = gpd.GeoSeries([
+            Point((-2, -2)),
+            Point((-8, -8)),
```
harrism, line 581: Why is this called polygons_right if it contains `Point`s, not `Polygon`s?

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -517,3 +509,102 @@ def to_arrow(self):
...
+            {
+                "input_types": aligned_input_types,
+                "union_offsets": aligned_union_offsets,
+            },
+        )
+        return GeoSeries(column)
+
+    def align(self, other):
+        """
+        Align the rows of two GeoSeries using outer join.
```
harrism, line 554: What does it mean to align the rows of geoseries? I think a more detailed explanation would help.

harrism (commented):
I could use more explanation of "align".

python/cuspatial/cuspatial/core/geoseries.py
thomcom: I will write one that covers this implementation.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: Copy/pasto

python/cuspatial/cuspatial/io/geopandas_reader.py
thomcom: Replaced it with a simple constant.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: I increased the docs, explaining three use cases, and provided examples for them. Please approve and merge @harrism .

harrism (approved):

harrism:
@gpucibot merge

Issue #744, opened 2022-10-20
[BUG] Point-in-polygon Primitive Should be Consistent Over Boundary Condition


isVoid:
**Describe the bug**
Consider two axis aligned rectangle:
```
[(-1.0, -1.0), (0.0, -1.0), (0.0, 1.0), (-1.0, 1.0)]
[(0.0, -1.0), (1.0, -1.0), (1.0, 1.0), (0.0, 1.0)]
```
and a test point `(0, 0)`.

![Untitled Diagram (1)](https://user-images.githubusercontent.com/13521008/196828503-2ef2768a-af5d-4a73-a251-0014e2f6fd33.jpg)

This point is on the edge of these two rectangles. Is this point in the rectangle? The answer should only be always yes, or should be always no for both cases. Not yes for one and no for another.

> Not yes for one and no for another.
This is what's happening in current PIP kernel.

**Steps/Code to reproduce bug**
See test: https://github.com/rapidsai/cuspatial/blob/930a2f33953ce1abe8f0680ebbf418675db4b362/cpp/tests/experimental/spatial/point_in_polygon_test.cu#L157

**Expected behavior**
Result should either be `b0000` or `b1111`. Better still, user definable on boundary conditions.

thomcom:
I think that #750 fixes this.

harrism:
No, the crossingsMultiply algorithm is fundamentally inconsistent. Working on it.

Issue #743, opened 2022-10-19
[FEA] Write .contains_properly for points, multipoints, linestrings, and polygons.


thomcom:
.contains is the most point-in-polygon-like binary operation.

For points, the result is the same as our point_in_polygon methods. I'll need to disambiguate between requests with 31 rows or fewer. In the case of 31 rows, use the unindexed point_in_polygon, or with more rows use quadtree. After the call is complete, the diagonal of the resulting DataFrame will need to be copied into a new Series.

For multipoints, linestrings, and polygons the behavior is the same as above, with an additional step to compute which complex geometry each point is a member of.

This will probably require building a new indexing scheme that allows any `xy` pair pulled from the coordinates buffer to be referenced back to its originating feature.

Write tests that slowly explore each of the above cases, comparing with GeoPandas results. If I'm lucky, our results will mostly match GeoPandas and all of .contains can be written in the next couple of weeks.

thomcom:
I've spent a lot of time this week trying to add a colinearity test to our point_in_polygon kernel, so that it will exclude the boundary. It has improved the results, but I'm stumped on a case where colinearity should exclude the point but is not. I don't want to dig on this forever, so I may start working on the rest of `.contains` even though I know the results aren't perfectly correct.

harrism:
Can you share / document the case that is stumping you?

Issue #742, opened 2022-10-18
Benchmark boundary exclusion


thomcom:
## Background

In https://github.com/rapidsai/cuspatial/pull/739 I introduce boundary exclusion to the `is_point_in_polygon` kernel that is used both by the original `point_in_polygon` as well as `quadtree_point_in_polygon`.

This addition adds 6 multiplies and 15 adds to the kernel per iteration. Because we added python benchmarks in `22.10` I was able to benchmark the new feature branch against the `22.12` branch.

The benchmark runs each point_in_polygon algorithm on 50m points and 30 polygons taken from the `naturalearth_lowres` feature database that is available in GeoPandas. You can run the benchmark with
```
cd cuspatial/python/cuspatial/benchmarks
pytest api/bench_api.py::bench_point_in_polygon
pytest api/bench_api.py::bench_quadtree_point_in_polygon
```
The default benchmark only tests 10m points, not 50m. You'll have to modify the tests to change to 50m. Finally, to run the benchmark using the new `is_point_in_polygon` kernel requires that you checkout my PR [branch](https://github.com/thomcom/cuspatial/tree/feature/pip_boundary_exclusion_test) and rebuild cuspatial cpp and python.

## Benchmark result

On my machine with a GV100 the boundary-exclusion branch is 30% faster, for some reason, than the original, taking 871msec on 50m points with boundary-exclusion and 1161msec without the new changes. Quadtree point in polygon is 10% slower with boundary exclusion at 36msec and 32msec with the original.

PR #770, opened 2022-11-02
`GeoSeries.contains` that matches `GeoPandas` with the exception of cases that depend on intersection.

thomcom:
## Description
Closes https://github.com/rapidsai/cuspatial/issues/738, https://github.com/rapidsai/cuspatial/issues/742, https://github.com/rapidsai/cuspatial/issues/743, and https://github.com/rapidsai/cuspatial/issues/744>

A feature $a$ is contained in a feature $b$ $\leftrightarrow$ all of the points in $a$ are in the inner region of $b$ and none of the exterior points of $a$ or $b$ intersect. This PR implements the first part of the `.contains` operation, that is, $\leftrightarrow$ all of the points in $a$ are in the inner region of $b$.

We do not have an intersection primitive available yet, so I will add the second portion of this operation when it is available. There are a small number of pytest `xfails` that demonstrate where we can't agree with GeoPandas yet.

This PR adds boundary exclusion to our point-in-polygon algorithm. Previously, point-in-polygon had inconsistent inclusion or exclusion of the polygon boundary, based on floating-point error and implementation details of the RayCrossings algorithm. Due to that inconsistency I've added explicit boundary exclusion.

Boundary exclusion in this context means that any point in $b$ that is colinear with a line segment in $a$ is treated as an intersection, which then violates the requirements of `.contains`, returning false for $a.contains(b)$.

## Checklist

### DIFF
--- a/cpp/include/cuspatial/experimental/detail/point_in_polygon.cuh
+++ b/cpp/include/cuspatial/experimental/detail/point_in_polygon.cuh
@@ -59,6 +59,7 @@ __device__ inline bool is_point_in_polygon(Cart2d const& test_point,
   using T = iterator_vec_base_type<Cart2dIt>;
 
   bool point_is_within = false;
+  bool is_colinear     = false;
   // for each ring
   for (auto ring_idx = poly_begin; ring_idx < poly_end; ring_idx++) {
     int32_t ring_idx_next = ring_idx + 1;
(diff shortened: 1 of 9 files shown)


harrism:
In the description, can you explain what "with the exception of exterior cases" means, precisely?

cpp/include/cuspatial/experimental/detail/point_in_polygon.cuh
```diff
@@ -72,22 +73,36 @@ __device__ inline bool is_point_in_polygon(Cart2d const& test_point,
...
-      if (y1_flag != y0_flag) {
-        T run           = b.x - a.x;
-        T rise          = b.y - a.y;
-        T rise_to_point = test_point.y - a.y;
+      T run          = b.x - a.x;
+      T rise         = b.y - a.y;
+
+      // Points on the line segment are the same, so intersection is impossible.
+      // This is possible because we allow closed or unclosed polygons.
+      if (run == 0.0 && rise == 0.0) continue;
```
isVoid, line 81: Can you try to replace this compare with the new utility I added in https://github.com/rapidsai/cuspatial/pull/773?
I'd use as `if (float_eq_by_ulp(b.x, a.x) && float_eq_by_ulp(b.y, a.y)) continue;`

isVoid: (No need to push to this PR for now - just want to test how usable that utility is)

thomcom: I was able to test it. Looks like it improves our accuracy for detecting those really hard corner cases like @harrism had identified.

```
            [Point([3.33, 1.11]), Polygon([[6, 2], [3, 1], [3, 4], [6, 2]]), True],
        ],
    )
    def test_float_precision_limits(point, polygon, expects):
        gpdpoint = gpd.GeoSeries(point)
        gpdpolygon = gpd.GeoSeries(polygon)
        point = cuspatial.from_geopandas(gpdpoint)
        polygon = cuspatial.from_geopandas(gpdpolygon)
        result = polygon.contains(point)
>       assert gpdpolygon.contains(gpdpoint).values == result.values_host
```

Produces a false now, previously True. It is good for us to have improved accuracy but in this case we conflict with Pandas, who produces True.

harrism: Isn't this because GeoPandas `contains` uses Shapely `contains` and not `contains_properly`? By excluding boundary points, you are now implementing `contains_properly`, not `contains`. https://shapely.readthedocs.io/en/latest/reference/shapely.contains_properly.html

thomcom:
Closing in favor of #749

PR #776, opened 2022-11-03
Better slicing via `union_offsets`

### LINKED ISSUE #771: [BUG] `GeoColumnAccessors` should respect slicing
I think I solved this, PR coming.

**Describe the bug**
The most visible interface for a `GeoSeries` respects slicing:

```
geoseries = cuspatial.from_geopandas(geopandas.read_file(geopandas.get_path("naturalearth_lowres")))
shorter = geoseries[0:3]
print(shorter)
0    MULTIPOLYGON (((180.00000 -16.06713, 180.00000...
1    POLYGON ((33.90371 -0.95000, 34.07262 -1.05982...
2    POLYGON ((-8.66559 27.65643, -8.66512 27.58948...
Name: geometry, dtype: geometry
```
However, the underlying `GeoColumnAccessor` that allows us to access the `GeoArrow` buffers, does not:
```
shorter.polygons.geometry_offset
0        0
1        3
2        4
3        5
4       35
      ... 
173    284
174    285
175    286
176    287
177    288
Length: 178, dtype: int32
```
**Expected behavior**
Users should not have to slice their `GeoSeries` and also slice the underlying buffers. When a sliced `GeoSeries` has `GeoColumnAccessors` called, they should return only the coordinates that are part of the `GeoSeries`.

thomcom:
## Description
Closes #771 

This PR modifies the production of `geometry_offset`, `part_offset`, and `ring_offset` by sampling the existing values in a `GeoSeries` before returning their various offsets. This has the effect of using `cudf.ListSeries` to re-pack any features into a new, dense `GeoColumn`, then returning the offsets based on it.

Previously, `GeoSeries` that had been modified by slicing would have the appearance of the sliced elements, but when `_offset` buffers were used they would return the full original offset buffer that the sliced `GeoSeries` had originated from. This was a problem because it made slicing useless for our algorithms.

I also modify the `core/spatial/distance.py` and `core/spatial/nearest_points.py` files to use `as_column(linestrings.lines.geometry_offset)` instead of `linestrings.lines.geometry_offset._column` because there doesn't appear to be a reason to use a `cudf.Series` to wrap the offset buffers. They are private methods essentially, don't need indexes, and will eventually be factored out so that they're hidden from the user.

I wrote a new test file `test_geocolumn_accessor.py` to exercise the new {`geometry_buffer`...} accessors for all geometry types. 

Finally I added tests for a base case, a more complicated case, and a case with noncontiugous slices to the inputs of `test_point_linestring_distance.py`, validating that the changes have exactly the effect we need.

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -148,6 +148,9 @@ class GeoSeries(cudf.Series):
             return cudf.Series(result)
 
         def _get_current_features(self, type):
+            # Resample the existing features so that the offsets returned
+            # by `_offset` methods reflect previous slicing, and match
+            # the values returned by .xy.
             existing_indices = self._meta.union_offsets[
                 self._meta.input_types == type.value
             ]
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -147,6 +147,13 @@ class GeoSeries(cudf.Series):
             result = self._col.take(indices._column).leaves().values
             return cudf.Series(result)
 
+        def _get_current_features(self, type):
+            existing_indices = self._meta.union_offsets[
+                self._meta.input_types == type.value
+            ]
+            existing_features = self._col.take(existing_indices._column)
+            return existing_features
+
     class MultiPointGeoColumnAccessor(GeoColumnAccessor):
         def __init__(self, list_series, meta):
             super().__init__(list_series, meta)
@@ -154,7 +161,7 @@ class GeoSeries(cudf.Series):
 
         @property
         def geometry_offset(self):
-            return cudf.Series(self._col.offsets.values)
+            return self._get_current_features(self._type).offsets.values
 
     class LineStringGeoColumnAccessor(GeoColumnAccessor):
         def __init__(self, list_series, meta):
@@ -163,36 +170,34 @@ class GeoSeries(cudf.Series):
 
         @property
         def geometry_offset(self):
-            return cudf.Series(self._col.offsets.values)
+            return self._get_current_features(self._type).offsets.values
 
         @property
         def part_offset(self):
-            return cudf.Series(self._col.elements.offsets.values)
+            return self._get_current_features(
+                self._type
+            ).elements.offsets.values
 
     class PolygonGeoColumnAccessor(GeoColumnAccessor):
         def __init__(self, list_series, meta):
             super().__init__(list_series, meta)
             self._type = Feature_Enum.POLYGON
 
-        def _get_current_features(self):
-            existing_indices = self._meta.union_offsets[
-                self._meta.input_types == Feature_Enum.POLYGON.value
-            ]
-            existing_features = self._col.take(existing_indices._column)
-            return existing_features
-
         @property
         def geometry_offset(self):
-            sliced_features = self._get_current_features()
-            return sliced_features.offsets.values
+            return self._get_current_features(self._type).offsets.values
 
         @property
         def part_offset(self):
-            return cudf.Series(self._col.elements.offsets.values)
+            return self._get_current_features(
+                self._type
+            ).elements.offsets.values
 
         @property
         def ring_offset(self):
-            return cudf.Series(self._col.elements.elements.offsets.values)
+            return self._get_current_features(
+                self._type
+            ).elements.elements.offsets.values
 
     @property
     def points(self):
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -141,11 +141,9 @@ class GeoSeries(cudf.Series):
 
         @property
         def xy(self):
-            types = self._meta.input_types
-            offsets = self._meta.union_offsets
-            indices = offsets[types == self._type.value]
-            result = self._col.take(indices._column).leaves().values
-            return cudf.Series(result)
+            return cudf.Series(
+                self._get_current_features(self._type).leaves().values
+            )
 
         def _get_current_features(self, type):
             # Resample the existing features so that the offsets returned
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -174,16 +174,17 @@ class GeoSeries(cudf.Series):
             super().__init__(list_series, meta)
             self._type = Feature_Enum.POLYGON
 
-        def _get_sliced_offset(self):
+        def _get_current_features(self):
             existing_indices = self._meta.union_offsets[
                 self._meta.input_types == Feature_Enum.POLYGON.value
             ]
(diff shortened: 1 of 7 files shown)


python/cuspatial/cuspatial/tests/test_geocolumn_accessor.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@pytest.mark.parametrize(
+    "range, expected",
+    [
+        [[0, 2], [0, 4, 8, 12]],
+        [[0, 1, 2], [0, 4, 8, 12, 16]],
+        [[2, 4], [0, 4, 8, 14, 18]],
+        [[4, 5], [0, 6, 10, 14]],
+        [[4, 3, 2], [0, 6, 10, 14, 18, 22]],
```
thomcom, line 223: I'm super happy with this PR except for one thing here. I don't understand why the `ring_offset` for polygon #4 is length 6. The polygon is length 5 and I don't understand why it is coming back here as length 6.

python/cuspatial/cuspatial/tests/test_geocolumn_accessor.py
```diff
@@ -0,0 +1,252 @@
...
+    shorter = gpdf[range]["geometry"]
+    expected = cp.array(expected)
+    got = shorter.polygons.geometry_offset
+    assert cp.array_equal(got, expected)
+
+
+@pytest.mark.parametrize(
+    "range, expected",
+    [
+        [[0, 2], [0, 2, 5]],
```
isVoid, line 34: I think what's tested here is gather indices but not a slice range.. IMO a range should be contiguous.

python/cuspatial/cuspatial/tests/conftest.py
```diff
@@ -204,3 +204,14 @@ def generator(n, max_num_geometries, max_num_segments):
...
+
+@pytest.fixture
+def slice_twenty():
+    return [
+        slice(0, 4),
+        slice(4, 8),
+        slice(8, 12),
+        slice(12, 16),
+        slice(16, 20),
+    ]
```
isVoid, line 217:
```suggestion
@pytest.fixture(params=[0, 4, 8, 12, 16])
def slice_twenty(request):
    return [request.param, request.param+4]
```

python/cuspatial/cuspatial/tests/test_point_linestring_distance.py
```diff
@@ -167,3 +167,121 @@ def test_mixed_geometry_series_raise(lhs, rhs):
...
+
+    # Just dropping the index for now, that is not libcuSpatial's problem
+    got.index = [1]
+    cudf.testing.assert_series_equal(got, cudf.Series(expected))
+
+
+@pytest.mark.parametrize("slice_index", [0, 1, 2, 3, 4])
+def test_multipoint_multilinestring_sliced_many(
+    multipoint_generator, multilinestring_generator, slice_twenty, slice_index
+):
```
isVoid, line 220: If you make changes according to my suggestion of fixture above, you don't need `slice_index`.

python/cuspatial/cuspatial/tests/test_point_linestring_distance.py
```diff
@@ -167,3 +167,121 @@ def test_mixed_geometry_series_raise(lhs, rhs):
...
+    glines = from_geopandas(hlines)
+
+    pslice = slice_twenty[slice_index]
+    hslicepts = hpts[pslice]
+    hslicelines = hlines[pslice]
+    slicegpts = gpts[pslice]
+    sliceglines = glines[pslice]
+    got = pairwise_point_linestring_distance(slicegpts, sliceglines)
+    expected = hslicepts.distance(hslicelines)
+    got.index = cudf.RangeIndex(pslice.start, pslice.stop)
```
isVoid, line 249: Looks like the python API is not copying the indices correctly.

isVoid (approved):
I agree that this is the solution to get the right offsets and coordinate arrays. Do you think there's a chance for optimization if the slices are simple contiguous range that can be represented by `offset` and `size`, the resulting child column slices can also be zero-copy?

python/cuspatial/cuspatial/tests/test_point_linestring_distance.py
thomcom: That's correct. I could modify the python API here, but I expect this to be an inner method on `object.distance(object)` eventually, so I didn't make those changes yet. Want me to?

python/cuspatial/cuspatial/tests/test_geocolumn_accessor.py
thomcom: Yes I'm testing both, a range and a spread out gather. Users will expect to be able to gather their `GeoSeries` in an arbitrary fashion.

python/cuspatial/cuspatial/tests/test_point_linestring_distance.py
thomcom: This reflects a pattern that I'm trying to build for some other test cases as well. There are a few tests where I use premade slicing patterns, then iterate over them to get a fair coverage of ways a column can be sliced. In those tests I just rewrite the same slice group a few different times inside of the test file, because you can't parameterize a feature in pytest. I realized while writing this that I could put the slices in `conftest.py` as a fixture, then simply iterate over the fixture using a `@pytest.mark.parametrize` over a range, as here. The goal is not fundamentally to iterate the segments of `0,4...4,8...` but to iterate over any predefined slicing range. Given our limited resources I didn't opt to refactor those tests yet, but I'd like to leave this, unless you know of a better way to iterate over a fixture in multiple tests.

isVoid: > The goal is not fundamentally to iterate the segments of 0,4...4,8... but to iterate over any predefined slicing range.

If you need other slicing range, can't you just create other fixtures?

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -141,11 +141,19 @@ def y(self):
...
-            types = self._meta.input_types
-            offsets = self._meta.union_offsets
-            indices = offsets[types == self._type.value]
-            result = self._col.take(indices._column).leaves().values
-            return cudf.Series(result)
+            return cudf.Series(
+                self._get_current_features(self._type).leaves().values
+            )
+
+        def _get_current_features(self, type):
```
harrism, line 148: What are features?

thomcom: A feature is a Point, MultiPoint, LineString, MultiLineString, Polygon, or MultiPolygon (or GeometryCollection).

harrism:
rerun tests

python/cuspatial/cuspatial/tests/test_point_linestring_distance.py
isVoid: Nope, this is out of scope of this PR. But should track with an issue.

thomcom:
@gpucibot merge

PR #793, opened 2022-11-09
Consolidate bounding box code

### LINKED ISSUE #746: [FEA] Consolidate C++ bounding box API functionality

cuSpatial has separate APIs for bounding boxes for trajectories, polygons, and linestrings.  These all duplicate code that uses `reduce_by_key` to compute the bounds of groups of points.

Ultimately, we should be able to call `.bounds()` on a GeoSeries, regardless of the contents. To enable this, we should build consistent C++ APIs for this function to call.

Start by providing a C++ function to compute the bounds of groups of points. This can be reused to compute the bounds of any geometry -- only the offset / ID computation needs to differ for different geometries.

harrism:
Fixes #746. Replaces the header-only `trajectory_bounding_boxes` with `point_bounding_boxes` and reuses it to implement `linestring_bounding_boxes` and `polygon_bounding_boxes`.

~By providing a single header-only function `point_bounding_boxes` that forms the implementation of `trajectory_bounding_boxes`, `linestring_bounding_boxes`, and `polygon_bounding_boxes`, this PR also closes closes 567 and closes 568.~

It does not, as `point_bounding_boxes` takes IDs, not polygon and ring offsets as the other APIs do. So we can't quite mark those as fixed.

## Checklist


robertmaynard (commented):
I believe that `cpp/tests/CMakeLists.txt` needs to be updated since we renamed `experimental/trajectory/trajectory_bounding_boxes_test.cu`

cpp/tests/CMakeLists.txt
```diff
@@ -143,8 +143,8 @@ ConfigureTest(POINT_IN_POLYGON_TEST_EXP
 ConfigureTest(DERIVE_TRAJECTORIES_TEST_EXP
     experimental/trajectory/derive_trajectories_test.cu)
 
-ConfigureTest(TRAJECTORY_BOUNDING_BOXES_TEST_EXP
-    experimental/trajectory/trajectory_bounding_boxes_test.cu)
+ConfigureTest(POINT_BOUNDING_BOXES_TEST_EXP
```
harrism, line 146: @robertmaynard see here, I did modify this test.

harrism:
> I believe that `cpp/tests/CMakeLists.txt` needs to be updated since we renamed `experimental/trajectory/trajectory_bounding_boxes_test.cu` 

You must have missed where I updated it? It's now called POINT_BOUNDING_BOXES_TEST_EXP

isVoid (approved):
Good cleanup!

cpp/include/cuspatial/experimental/detail/bounding_box.cuh
```diff
@@ -46,27 +60,32 @@ struct box_minmax {
...
-                                              IdInputIt ids_last,
-                                              PointInputIt points_first,
-                                              BoundingBoxOutputIt bounding_boxes_first,
-                                              rmm::cuda_stream_view stream)
+template <typename IdInputIt, typename PointInputIt, typename BoundingBoxOutputIt, typename T>
+BoundingBoxOutputIt point_bounding_boxes(IdInputIt ids_first,
+                                         IdInputIt ids_last,
+                                         PointInputIt points_first,
+                                         BoundingBoxOutputIt bounding_boxes_first,
+                                         T expansion_radius,
```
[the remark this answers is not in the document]
harrism: Discussed offline with @isVoid. Doing this would require doing the addition of the expansion radius to each bounding box after the reduction. We looked at a few of approaches.

1. `thrust::transform` after the `reduce_by_key`. This incurs an additional kernel launch. The cost of this may be larger than the cost of a couple of adds to each input coordinate, since the reduction should be bandwidth bound (so the adds are "free").
2. Use a `transform_output_iterator`. This doesn't work because the output type of a `reduce_by_key` bust be constructable from an instance of the input type. 
3. Leave the code as is (expand every point into a box using the expansion radius).

We chose option 3.

cpp/include/cuspatial/experimental/bounding_box.cuh
```diff
@@ -16,22 +16,29 @@
...
+ * @addtogroup spatial_relationship
  * @{
  */
 
 /**
- * @brief Compute the spatial bounding boxes of trajectories.
+ * @brief Compute the spatial bounding boxes of sequences of points.
+ *
+ * Computes a bounding box around all points within each group (consecutive points with the same
+ * ID).
```
harrism, line 34:
```suggestion
 * ID). This function can be applied to trajectory data, polygon vertices, linestring vertices, or any grouped point data.
```

trxcllnt:
rerun tests

harrism:
Ugh, I wish pre-commit would run for github-ui commits.

robertmaynard (approved):
Aprroving CMake changes

trxcllnt (approved):

harrism:
@gpucibot merge

PR #797, opened 2022-11-14
[FIX] Update quadtree spatial join docstrings

### LINKED ISSUE #397: [DOC/QST] Scale parameter in quadtree methods

### LINKED ISSUE #733: [BUG] quadtree_point_in_polygon.cu, bug and fix
### 1. Bug description
Version: 22.12 and 22.10: will use 22.12 to walk through'
We will show the cause of this bug and how we fixed it.
**Note: to trigger the bug, one needs to set the scale appropriately to create deep quad-tree**

We are using the quad-tree based point to polygon intersection, 
cuspatial.quadtree_on_points
cuspatial.join_quadtree_and_bounding_boxes
cuspatial.**quadtree_point_in_polygon**

and the last step, quadtree_point_in_polygon, returns wrong indices for points. 
 
https://github.com/rapidsai/cuspatial/blob/c4b1440fbac1b7a186ef9a4765ced1d1f1d14312/cpp/include/cuspatial/spatial_join.hpp#L100-L109

quadtree_point_in_polygon supposes to return **point_offset**, which should be the indices into (point_x, point_y).
(point_x, point_y) are column data from the original table, so point_offset is just the row_ID of the original table.
And we use the indices to join tables, etc.

### 2. The cause of the bug
**The bug is that the point_offset is wrong, it actually returns the indices to point_indices instead of (point_x, point_y)**
point_indices is parted of the quad-tree leaf node storage (sorted by morton code), which points to(point_x, point_y)

The bug is in cpp/src/join/quadtree_point_in_polygon.cu
here's the relevant parts of the code

The copy_if copies from global_to_poly_and_point_indices to poly_and_point_indices (which is the final output), if points intersect with polygon
https://github.com/rapidsai/cuspatial/blob/c4b1440fbac1b7a186ef9a4765ced1d1f1d14312/cpp/src/join/quadtree_point_in_polygon.cu#L181-L190

global_to_poly_and_point_indices iterator is created by apply compute_poly_and_point_indices 
https://github.com/rapidsai/cuspatial/blob/c4b1440fbac1b7a186ef9a4765ced1d1f1d14312/cpp/src/join/quadtree_point_in_polygon.cu#L170-L176

and compute_poly_and_point_indices
https://github.com/rapidsai/cuspatial/blob/c4b1440fbac1b7a186ef9a4765ced1d1f1d14312/cpp/src/join/quadtree_point_in_polygon.cu#L53-L69

quad_point_offsets contains the quad tree's leaf nodes' offset to the **point_indices**, so this functor just returns indices to  **point_indices**
(https://github.com/rapidsai/cuspatial/blob/c4b1440fbac1b7a186ef9a4765ced1d1f1d14312/cpp/src/join/quadtree_point_in_polygon.cu#L65)

as one can see ,
compute_poly_and_point_indices returns indices to  **point_indices**
so global_to_poly_and_point_indices returns indices to  **point_indices**
and finally it just copies global_to_poly_and_point_indices to poly_and_point_indices, which is the final output, that's how we get indices to point_indices instead of point_x y.

### 3. The fix
`pass point_indices to compute_poly_and_point_indices`
`then return the content of point_indices (indices to point x y ) `

template <typename T, typename QuadOffsetsIter>
struct compute_poly_and_point_indices {
  QuadOffsetsIter quad_point_offsets;
  uint32_t const* point_offsets;
  uint32_t const* point_offsets_end;
  cudf::column_device_view const poly_indices;
  **cudf::column_device_view const point_indices;**

  inline thrust::tuple<uint32_t, uint32_t> __device__
  operator()(cudf::size_type const global_index) const
  {
    // uint32_t quad_poly_index, local_point_index;
    auto const [quad_poly_index, local_point_index] =
      get_quad_poly_and_local_point_indices(global_index, point_offsets, point_offsets_end);
    **uint32_t const point_idx = point_indices.element<uint32_t>(quad_point_offsets[quad_poly_index] + local_point_index);**
    uint32_t const poly_idx  = poly_indices.element<uint32_t>(quad_poly_index);
    return thrust::make_tuple(poly_idx, point_idx);
  }
};

`change point_xys_iter into iterator on point x y`
 auto point_xys_iter =  thrust::make_zip_iterator(point_x.begin<T>(), point_y.begin<T>());

`pass new arg to compute_poly_and_point_indices`

    auto global_to_poly_and_point_indices = detail::make_counting_transform_iterator(
        0,
        compute_poly_and_point_indices<T, decltype(quad_offsets_iter)>{
          quad_offsets_iter,
          local_point_offsets.begin(),
          local_point_offsets.end(),
          *cudf::column_device_view::create(poly_indices, stream),
          *cudf::column_device_view::create(point_indices, stream)
          });

this works because global_to_poly_and_point_indices now returns indices to point x y directly, which is passed to test_poly_point_intersection, then we changed **PointIter points;** to simple iterator on point x,y, so this gives us the actual data for point testing

https://github.com/rapidsai/cuspatial/blob/c4b1440fbac1b7a186ef9a4765ced1d1f1d14312/cpp/src/join/quadtree_point_in_polygon.cu#L71-L92

and global_to_poly_and_point_indices is if_copied back to user, which suppose to have indices to point x y. so this is the correct output.

We verified it with our spatial join cases, the above fix works. It produces the correct results compared to geopandas sjoin

trxcllnt:
* Aligns docstrings between C++ and Python
* Adds missing methods to Python docs
* Fixes #733 
* Fixes #397

### DIFF
--- a/cpp/include/cuspatial/spatial_join.hpp
+++ b/cpp/include/cuspatial/spatial_join.hpp
@@ -38,7 +38,8 @@ namespace cuspatial {
  * eventual number of levels may be less than `max_depth` if the number of points is small or
  * `max_size` is large.
  *
- * @param quadtree: cudf table representing a quadtree (key, level, is_quad, length, offset).
+ * @param quadtree: cudf table representing a quadtree (key, level, is_internal_node, length,
+ * offset).
  * @param bbox: cudf table of bounding boxes as four columns (x_min, y_min, x_max, y_max).
  * @param x_min The lower-left x-coordinate of the area of interest bounding box.
  * @param x_max The upper-right x-coordinate of the area of interest bounding box.
@@ -85,7 +86,8 @@ std::unique_ptr<cudf::table> join_quadtree_and_bounding_boxes(
  *
  * @param poly_quad_pairs cudf table of (polygon, quadrant) index pairs returned by
  * `cuspatial::join_quadtree_and_bounding_boxes`
- * @param quadtree cudf table representing a quadtree (key, level, is_quad, length, offset).
+ * @param quadtree cudf table representing a quadtree (key, level, is_internal_node, length,
+ * offset).
  * @param point_indices Sorted point indices returned by `cuspatial::quadtree_on_points`
  * @param point_x x-coordinates of points to test
  * @param point_y y-coordinates of points to test
@@ -132,7 +134,8 @@ std::unique_ptr<cudf::table> quadtree_point_in_polygon(
  *
  * @param linestring_quad_pairs cudf table of (linestring, quadrant) index pairs returned by
  * `cuspatial::join_quadtree_and_bounding_boxes`
- * @param quadtree cudf table representing a quadtree (key, level, is_quad, length, offset).
+ * @param quadtree cudf table representing a quadtree (key, level, is_internal_node, length,
+ * offset).
  * @param point_indices Sorted point indices returned by `cuspatial::quadtree_on_points`
  * @param point_x x-coordinates of points to test
  * @param point_y y-coordinates of points to test
--- a/cpp/include/cuspatial/spatial_join.hpp
+++ b/cpp/include/cuspatial/spatial_join.hpp
@@ -32,9 +32,11 @@ namespace cuspatial {
 /**
  * @brief Search a quadtree for polygon or linestring bounding box intersections.
  *
- * @note `scale` is applied to (x - x_min) and (y - y_min) to convert coordinates into a Morton code
- * in 2D space.
- * @note `max_depth` should be less than 16, since Morton codes are represented as `uint32_t`.
+ * @note 2D coordinates are converted into a 1D Morton code by dividing each x/y by the `scale`:
+ * (`(x - min_x) / scale` and `(y - min_y) / scale`).
+ * @note `max_depth` should be less than 16, since Morton codes are represented as `uint32_t`. The
+ * eventual number of levels may be less than `max_depth` if the number of points is small or
+ * `max_size` is large.
  *
  * @param quadtree: cudf table representing a quadtree (key, level, is_quad, length, offset).
  * @param bbox: cudf table of bounding boxes as four columns (x_min, y_min, x_max, y_max).
@@ -104,8 +106,8 @@ std::unique_ptr<cudf::table> join_quadtree_and_bounding_boxes(
  * polygon_offset - UINT32 column of polygon indices
  *   point_offset - UINT32 column of point indices
  *
- * @note The returned polygon and point indices are offsets into the poly_quad_pairs inputs and
- * (point_x, point_y), respectively.
+ * @note The returned polygon and point indices are offsets into the `poly_quad_pairs` inputs and
+ * `point_indices`, respectively.
  *
  **/
 std::unique_ptr<cudf::table> quadtree_point_in_polygon(
@@ -153,8 +155,8 @@ std::unique_ptr<cudf::table> quadtree_point_in_polygon(
  *   distance          - FLOAT or DOUBLE column (based on input point data type) of distances
  *                       between each point and linestring
  *
- * @note The returned point and linestring indices are offsets into the (point_x, point_y) and
- * linestring_quad_pairs inputs, respectively.
+ * @note The returned point and linestring indices are offsets into the `point_indices` and
+ * `linestring_quad_pairs` inputs, respectively.
  *
  **/
 std::unique_ptr<cudf::table> quadtree_point_to_nearest_linestring(
--- a/python/cuspatial/cuspatial/core/spatial/indexing.py
+++ b/python/cuspatial/cuspatial/core/spatial/indexing.py
@@ -61,20 +61,19 @@ def quadtree_on_points(
                 An int32 column of quadrant keys
             level       : cudf.Series(dtype=np.int8)
                 An int8 column of quadtree levels
-            is_quad     : cudf.Series(dtype=np.bool_)
-                A boolean column indicating whether the node is a parent (true)
-                or leaf (false) node
+            is_internal_node : cudf.Series(dtype=np.bool_)
+                A boolean column indicating whether the node is a quad or leaf
             length      : cudf.Series(dtype=np.int32)
-                If this is a non-leaf quadrant (i.e. ``is_quad`` is ``True``),
-                this column's value is the number of children in the non-leaf
-                quadrant.
+                If this is a non-leaf quadrant (i.e. ``is_internal_node`` is
+                ``True``), this column's value is the number of children in
+                the non-leaf quadrant.
 
                 Otherwise this column's value is the number of points
                 contained in the leaf quadrant.
             offset      : cudf.Series(dtype=np.int32)
-                If this is a non-leaf quadrant (i.e. ``is_quad`` is ``True``),
-                this column's value is the position of the non-leaf quadrant's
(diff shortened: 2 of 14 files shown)


harrism:
@trxcllnt since #733 is a long discussion and recommends a different fix rather than just documentation, can you please comment in the description on how specifically this PR fixes that issue?

python/cuspatial/cuspatial/core/spatial/indexing.py
```diff
@@ -51,7 +62,8 @@ def quadtree_on_points(
             level       : cudf.Series(dtype=np.int8)
                 An int8 column of quadtree levels
             is_quad     : cudf.Series(dtype=np.bool_)
```
harrism, line 64: Should we call this `is_parent` or `is_parent_quad` instead, for clarity? After all, leaves are quads too, right?

Alternatively, call it `is_leaf` and invert the values.

trxcllnt:
Commit https://github.com/rapidsai/cuspatial/pull/797/commits/37ca193cf4e6fcbd2ae9b4e9b3483a435e5fed38 fixes #733 by updating the docstring to indicate the resulting indices are offsets into the input `point_indices`, and not the original `point_x` and `point_y` columns.

python/cuspatial/cuspatial/core/spatial/indexing.py
```diff
@@ -147,6 +172,15 @@ def quadtree_on_points(
         Length: 120, dtype: int32
     """
 
+    if min_size is not None and max_size is None:
+        max_size = min_size
+        min_size = None
+        warnings.warn(
```
harrism, line 178: https://github.com/rapidsai/cuspatial/issues/798 files to remove this in 23.02 after we deprecate it in 22.12.

python/cuspatial/cuspatial/core/spatial/join.py
```diff
@@ -238,7 +239,8 @@ def quadtree_point_in_polygon(
         polygon_index : cudf.Series
             Indices of each polygon with which a point intersected.
         point_index : cudf.Series
-            Indices of each point that intersects with a polygon.
+            Indices of each point (in ``point_indices``) that intersects with a
```
harrism, line 242:
```suggestion
            Indices of points (in ``point_indices``) that are contained in a
```

harrism: Also, I find the "(in ``point indices``) a bit unclear.  Are the actual indices of a contained polygon returned, or are the indices of the points in `point_indices` returned? i.e. is it this:

```
point_indices = [0, 5, 9, 21, 45]
contained_points = [0, 9, 45]
return contained_points
```

or this:

```
point_indices = [0, 5, 9, 21, 45]
contained_points = [0, 2, 4]
return contained_points
```

harrism (changes requested):
Thanks for the fixes. Some questions.

python/cuspatial/cuspatial/core/spatial/join.py
trxcllnt: The second one. The returned indices can be used as a gather map for `point_indices`, which is itself a gather map for `point_x` and `point_y`:

```python
points = cudf.DataFrame({"x": [-5, 5], "y": [-5, 5]})

indices, quadtree = cuspatial.quadtree_on_points(points["x"], points["y"], ...)
bboxes = cuspatial.polygon_bounding_boxes(...)
intersections = cuspatial.join_quadtree_and_bounding_boxes(quadtree, bboxes, ...)
polygons_and_points = cuspatial.quadtree_point_in_polygon(
  intersections, quadtree, indices, points["x"], points["y"], ...
)

intersecting_points = points[indices[polygons_and_points["point_index"]]]
```

python/cuspatial/cuspatial/core/spatial/indexing.py
trxcllnt: I thought about renaming it but didn't. It'd be a breaking change and wasn't sure if we wanted to introduce one.

python/cuspatial/cuspatial/core/spatial/join.py
```diff
@@ -238,7 +239,8 @@ def quadtree_point_in_polygon(
         polygon_index : cudf.Series
             Indices of each polygon with which a point intersected.
```
harrism, line 240:
```suggestion
            Index of containing polygon.
```

python/cuspatial/cuspatial/core/spatial/join.py
harrism: OK, so these are indices of indices. That is very hard to explain succinctly. I think an example that makes this clear should be added to the docstring.  And I would word it something like:

```suggestion
            Index of contained point. This index refers to ``point_indices``, so it is an index to an index.
```

Note I suggested changing the working of the polygon index above to match. Is the polygon index an index to an index as well?

python/cuspatial/cuspatial/core/spatial/join.py
```diff
@@ -311,7 +313,8 @@ def quadtree_point_to_nearest_linestring(
         between the two.
 
         point_index : cudf.Series
-            Indices of each point that intersects with a linestring.
+            Indices of each point (in ``point_indices``) that intersects with
+            a linestring.
         linestring_index : cudf.Series
             Indices of each linestring with which a point intersected.
         distance : cudf.Series
```
harrism, line 320: I'm providing the whole suggested return documentation here because it runs outside the range of lines I can comment on in this PR.  Please make analogous changes to the PiP one as mentioned. Also, please double check that the right double quotes are being used. I'm not that familiar with Python docstring syntax.

```
    result : cudf.DataFrame
        Indices for each point and its nearest linestring, and the distance
        between the two.

        point_index : cudf.Series
            Index of point. This index refers to ``point_indices``, so it is an index to an index.
        linestring_index : cudf.Series
            Index of the nearest linestring to the point.
        distance : cudf.Series
            Distance between point and its nearest linestring.
```

trxcllnt:
rerun tests

cpp/include/cuspatial/spatial_join.hpp
```diff
@@ -32,11 +32,14 @@ namespace cuspatial {
 /**
  * @brief Search a quadtree for polygon or linestring bounding box intersections.
  *
- * @note `scale` is applied to (x - x_min) and (y - y_min) to convert coordinates into a Morton code
- * in 2D space.
- * @note `max_depth` should be less than 16, since Morton codes are represented as `uint32_t`.
- *
- * @param quadtree: cudf table representing a quadtree (key, level, is_quad, length, offset).
+ * @note 2D coordinates are converted into a 1D Morton code by dividing each x/y by the `scale`:
```
harrism, line 35: I think this explanation fixes #397 .

harrism (approved):

harrism:
Also Fixes #397

trxcllnt:
rerun tests

thomcom (approved):

trxcllnt:
@gpucibot merge

PR #799, opened 2022-11-15
Update pre-commit configuration and CI

bdice:
## Description
This PR updates the pre-commit configuration and uses pre-commit for all CI style checks. This aligns the configuration closer to that of cudf, but does not add all of the hooks from cudf's list to keep this PR limited in scope.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #803, opened 2022-11-16
Update Dependency List with `dependencies.yaml` [skip gpuci]

isVoid:
## Description

This PR updates dependency lists with `dependencies.yaml`.

## Checklist


dependencies.yaml
```diff
@@ -0,0 +1,100 @@
...
+          - geopandas>=0.11.0
+  cudatoolkit:
+    specific:
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.5"
+            packages:
+              - cudatoolkit=11.5
+  develop:
```
bdice, line 56: Most of this section should probably disappear with #799. Developers should only need pre-commit in this section, and use it to invoke the proper versions of all linters.

bdice: I’d move the docs dependencies into their own section.

dependencies.yaml
```diff
@@ -0,0 +1,100 @@
...
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - black=22.3.0
+          - cmakelang=0.6.13
+          - flake8=3.8.3
+          - gcovr>=5.0
```
bdice, line 63: Is gcovr actually used? I was only aware of it being used with rmm.

dependencies.yaml
```diff
@@ -0,0 +1,100 @@
...
+              - python=3.9
+  run:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - numba>=0.49
+          - numpy>=1.19
+  test_cpp:
+    common:
+      - output_types: [conda]
```
bdice, line 91:
```suggestion
      - output_types: conda
```

dependencies.yaml
isVoid: Probably not, the list is derived from rmm.

isVoid: Removed all linters.

conda/environments/all_cuda-115_arch-x86_64.yaml
```diff
@@ -0,0 +1,27 @@
...
+- conda-forge
+- nvidia
+dependencies:
+- clang-tools=11.1.0
+- clang=11.1.0
+- cmake>=3.23.1
+- cudatoolkit=11.5
+- cudf=22.12.*
+- cython>=0.29,<0.30
+- gdal>=3.0.2
```
harrism, line 15: Don't we need 3.5?

dependencies.yaml
```diff
@@ -0,0 +1,101 @@
...
+          - matrix:
+              cuda: "11.5"
+            packages:
+              - cudatoolkit=11.5
+  develop:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - pre-commit 
+      - output_types: conda
```
bdice, line 62: These clang packages can also be removed after #799, which uses pre-commit to handle clang-format.

dependencies.yaml
```diff
@@ -0,0 +1,101 @@
...
+    output: none
+    includes:
+      - cudatoolkit
+      - test_cpp 
+  checks:
+    output: none
+    includes:
+      - build
+      - develop
+      - py_version
```
bdice, line 31: Style checks should not be dependent on a specific Python version being specified in the matrix (if no value is specified, the matrix match will fail). The Python requirement in the `build` list should be sufficient. However, we may want a default (empty matrix in the last entry) of `py_version` defining `>=3.8,<3.10` instead of putting it in `build`.
```suggestion
```

dependencies.yaml
```diff
@@ -0,0 +1,101 @@
...
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - numba>=0.49
+          - numpy>=1.19
+  test_cpp:
+    common:
+      - output_types: conda
+        packages:
+          - gtest=1.10.0
```
bdice, line 94: Why are these pinned?

dependencies.yaml
```diff
@@ -0,0 +1,101 @@
...
+              - python=3.8
+          - matrix:
+              py: "3.9"
+            packages:
+              - python=3.9
+  run:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - numba>=0.49
```
bdice, line 88: Have we reviewed these >= values recently? These versions of Numba and NumPy are getting kind of old. Do the minimum versions still work?

isVoid: Cuspatial doesn't really depend on the use of numba as I don't think it hosts any jitted kernel. Should remove probably.

dependencies.yaml
isVoid: Are these used for `clang-format` or used to setup language servers for intellisense?

dependencies.yaml
isVoid: No reason I'm currently aware of.

dependencies.yaml
isVoid: Any reason doing it just for py_version? Why not also cython?

isVoid: And are you suggesting removing `python` entry from `build` and use `py_version` in `all`?

conda/environments/all_cuda-115_arch-x86_64.yaml
isVoid: Updated to reflect https://github.com/rapidsai/cuspatial/pull/806

dependencies.yaml
isVoid: Removed

dependencies.yaml
isVoid: Removed

dependencies.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - rapidsai-nightly
+  - conda-forge
+  - nvidia
+dependencies:
+  build:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - cudf=22.12.*
+          - cmake>=3.23.1
```
galipremsagar, line 42:
```suggestion
          - cmake>=3.23.1,!=3.25.0
```
To sync with: https://github.com/rapidsai/cuspatial/pull/805/files

conda/environments/all_cuda-115_arch-x86_64.yaml
```diff
@@ -0,0 +1,23 @@
+# This file is generated by `rapids-dependency-file-generator`.
+# To make changes, edit dependencies.yaml and run `rapids-dependency-file-generator`.
+channels:
+- rapidsai
+- rapidsai-nightly
+- conda-forge
+- nvidia
+dependencies:
+- cmake>=3.23.1
```
galipremsagar, line 9:
```suggestion
- cmake>=3.23.1,!=3.25.0
```

dependencies.yaml
bdice: I think these are just for `clang-format`. The language servers are independently managed (in rapids-compose, at least), and I think this clang version is too old for that anyway.

dependencies.yaml
bdice: I think you did what I wanted -- remove `py_version` from the style check dependency list.

As for removing `python` from `build` -- is there a C++ library that can be built without Python? If so, I would remove `python`, `cython`, ... from `build` and put them in a separate list like `build_python`. We're trying to achieve a level of modularity/granularity that is useful for minimizing dependencies at each stage of a build.

dependencies.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - rapidsai
+  - rapidsai-nightly
+  - conda-forge
+  - nvidia
+dependencies:
+  build:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - cudf=22.12.*
```
bdice, line 41: I would have expected `libcudf` and `librmm` / `rmm` in this list, too? Avoid making those transitive dependencies if cuspatial directly relies on them via `#include` or `cimport`. (If it is only `import`ed in Python, it is a runtime and not a build time dependency.) Also combine this advice with my other comment to split out the C++ and Python build dependencies (we have yet to do this for cudf, still work in progress).

dependencies.yaml
```diff
@@ -0,0 +1,84 @@
...
+        packages:
+          - pydata-sphinx-theme
+          - myst-parser
+  py_version:
+    specific:
+      - output_types: conda
+        matrices:
+          - matrix:
+            packages:
+              - python>=3.8,<3.10
```
bdice, line 72: Use a specific Python version if requested in the matrix, and fall back to the broader pinning of `python>=3.8,<3.10` if `py` is not specified in the matrix.
```suggestion
  py_version:
    specific:
      - output_types: conda
        matrices:
          - matrix:
              py: "3.8"
            packages:
              - python=3.8
          - matrix:
              py: "3.9"
            packages:
              - python=3.9
          - matrix:
            packages:
              - python>=3.8,<3.10
```

ajschmidt8:
> I would have expected `libcudf` and `librmm` / `rmm` in this list, too? Avoid making those transitive dependencies if cuspatial directly relies on them via `#include` or `cimport`. (If it is only `import`ed in Python, it is a runtime and not a build time dependency.) Also combine this advice with my other comment to split out the C++ and Python build dependencies (we have yet to do this for cudf, still work in progress).

@bdice, I just pushed some changes. I'll defer to you to add these necessary dependencies. Otherwise, I think this PR looks good to me!

dependencies.yaml
```diff
@@ -0,0 +1,157 @@
...
+            packages:
+              - python=3.9
+          - matrix:
+            packages:
+              - python>=3.8,<3.10
+  run_python:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - cudf=22.12.*
```
ajschmidt8, line 143: @bdice, I see we have `cudf=22.12.*` twice in this file. any objections to using YAML anchors to keep it DRY?

bdice: It doesn't really matter since these are handled by `ci/release/update-version.sh`. No objections - either way is fine with me.

bdice: Updated in 56151d16831d16ed49d2e8e637e387a796986384.

bdice (approved):

bdice:
@ajschmidt8 Feel free to merge if you think this is ready, then we can start the GitHub Actions migration. I checked the current CI workflows and I think we have already handled all the features that the cuSpatial repo will need.

ajschmidt8 (approved):

PR #806, opened 2022-11-17
Set max version pin for `gdal`

ajschmidt8:
## Description

This PR sets a maximum version pin for `gdal`. Recently, `gdal` version `3.6.0` was released, which has a different version number for the `libgdal` shared object (`libgdal.so.32` vs. `libgdal.so.31`).

Setting the max pin here will ensure that we use a consistent `libgdal` shared object version for building and at runtime.

This should solve the `ImportError: libgdal.so.32: cannot open shared object file: No such file or directory` errors that have been popping up in recent Jenkins logs.

## Checklist

### DIFF
--- a/conda/environments/cuspatial_dev_cuda11.5.yml
+++ b/conda/environments/cuspatial_dev_cuda11.5.yml
@@ -9,7 +9,7 @@ dependencies:
   - clang-tools=11.1.0
   - cudf=22.12.*
   - cudatoolkit=11.5
-  - gdal>=3.0.2
+  - gdal>3.5.0,<3.6.0
   - geopandas>=0.11.0
   - cmake>=3.23.1
   - cython>=0.29,<0.30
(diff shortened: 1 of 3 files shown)


raydouglass (approved):

galipremsagar (approved):

thomcom (approved):

thomcom:
@gpucitbot merge

harrism:
Note that there is a interaction with tensorflow  and gdal. We had a past PR (can't find, on mobile now) that set / pinned our gdal version in the conda recipe to avoid conflicts with AI packages.

ajschmidt8:
> Note that there is a interaction with tensorflow and gdal. We had a past PR (can't find, on mobile now) that set / pinned our gdal version in the conda recipe to avoid conflicts with AI packages.

Good to know. I think we should be okay here though since we're setting a max pin. The max pin will just exclude the `3.6.0` version of `gdal` which was released in the last day.

ajschmidt8:
@gpucibot merge

ajschmidt8:
(there was a typo in Thomson's merge comment)

PR #813, opened 2022-11-22
Add `pairwise_linestring_intersection_with_duplicates` API

### LINKED ISSUE #785: Header-only Implementation of `intersection_with_duplicates`
Implement `pairwise_linestring_intersection_with_duplicates`, returns intersections between two linestrings and the indices corresponding to the intersecting linestring and multilinestrings. The result may contain duplicating points and mergable point on segments.

https://github.com/rapidsai/cuspatial/issues/785#tasklist-block-4f88f4ad-a9db-4b00-a00c-c0796c672e36

isVoid:
## Description

This PR adds internal function: `pairwise_linestring_intersection_with_duplicates` API. closes #785,

The API returns two structs: `linestring_intersection_intermediates<vec_2d>` and `linestring_intersection_intermediates<segment>`. This struct is important for following steps when removing duplicate geometries. Because the geometries and underlying look-back ids needs to be updated in tandem. 

Notice I didn't go too hard on the test cases - result from intersection with duplicates is often hard to hand craft. I would rely on future integrated tests with end-to-end intersection to test the correctness of this function.

## Checklist


robertmaynard (approved):
Approving CMake changes

cpp/include/cuspatial/experimental/ranges/range.cuh
```diff
@@ -0,0 +1,47 @@
...
+
+#include <cuspatial/cuda_utils.hpp>
+#include <cuspatial/traits.hpp>
+
+#include <thrust/distance.h>
+
+namespace cuspatial {
+
+template <typename IteratorType>
+class range {
```
harrism, line 27: docs

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,297 @@
...
+                                                        OutputIt1 points_first,
+                                                        OutputIt2 segments_first)
+{
+  using T       = typename MultiLinestringRange1::element_t;
+  using types_t = uint8_t;
+  using count_t = iterator_value_type<Offsets1>;
+  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < multilinestrings1.num_points();
+       idx += gridDim.x * blockDim.x) {
+    auto const part_idx = multilinestrings1.part_idx_from_point_idx(idx);
+    if (!multilinestrings1.is_valid_segment_id(idx, part_idx)) continue;
```
harrism, line 177: Another example of these two functions called together.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,297 @@
...
+    auto const lhs_linestring_idx = multilinestrings1.intra_part_idx(part_idx);
+    auto const lhs_segment_idx    = multilinestrings1.intra_point_idx(idx);
+    auto [a, b]                   = multilinestrings1.segment(idx);
+    auto const geometry_idx       = multilinestrings1.geometry_idx_from_part_idx(part_idx);
+    auto const multilinestring2   = multilinestrings2[geometry_idx];
+    auto const geometry_collection_offset =
+      num_points_offsets_first[geometry_idx] + num_segments_offsets_first[geometry_idx];
+
+    for (auto rhs_linestring_idx = 0; rhs_linestring_idx < multilinestring2.size();
+         ++rhs_linestring_idx) {
```
harrism, line 187: Now that you have a `range` type, could be?

```suggestion
    for (auto rhs_linestring_idx : cuspatial::range(0, multilinestring2.size() - 1) {
```

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,297 @@
...
+      auto const linestring2 = multilinestring2[rhs_linestring_idx];
+      for (auto rhs_segment_idx = 0; rhs_segment_idx < linestring2.num_segments();
+           ++rhs_segment_idx) {
+        auto [c, d]                   = linestring2.segment(rhs_segment_idx);
+        auto [point_opt, segment_opt] = segment_intersection(segment<T>{a, b}, segment<T>{c, d});
+
+        // Writes geometry and origin IDs to output. Note that for each pair, intersecting
+        // points always precedes overlapping segments (arbitrarily).
+        if (point_opt.has_value()) {
+          auto r              = cuda::atomic_ref<count_t>{*(n_points_stored + geometry_idx)};
```
harrism, line 197: Do you need the pointer arithmetic here?

```suggestion
          auto r              = cuda::atomic_ref<count_t>{n_points_stored[geometry_idx]};
```

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,297 @@
...
+          points_first[num_points_offsets_first[geometry_idx] + next_point_idx] = point_opt.value();
+          auto union_column_idx = geometry_collection_offset + next_point_idx;
+          auto point_column_idx = num_points_offsets_first[geometry_idx] + next_point_idx;
+          point_ids_range.set(point_column_idx,
+                              lhs_linestring_idx,
+                              lhs_segment_idx,
+                              rhs_linestring_idx,
+                              rhs_segment_idx);
+        } else if (segment_opt.has_value()) {
+          auto r                = cuda::atomic_ref<count_t>{*(n_segments_stored + geometry_idx)};
```
harrism, line 208: And here?

harrism (changes requested):

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: Ooh, looks nice

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: No. I don't think they are to be combined. `part_idx` is the index to the linestring in the multilinestring. Here `idx` happens to be interpreted as the index of the segment, and for each linestring there is always an invalid segment idx at the end of the linestring. `is_valid_segment_id` detects those segment index and removes them. 

In theory the above are two logically separate steps. In reality, the second API depends on the first one. So if you want to use the second API, you always need to compute the first one. But you can have a kernel that only calls the first API and not the second.

cpp/include/cuspatial/experimental/ranges/range.cuh
isVoid: Done

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: No specific reason, I think they are the same.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: Hang on... You can't do that. `Range` is composed of iterators, not raw values. You wouldn't want to dereference `0` as an address.

I think what you are looking for is `range(counting_iterator(0), counting_iterator(multilinestring2.size()-1))`, but I'm not sure which is more succinct. Also, in c++ 20, this is what we need: https://en.cppreference.com/w/cpp/ranges/iota_view

harrism: Actually, you are looking for https://github.com/harrism/ranger :)

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * @brief Compute intersections between multilnestrings with duplicates.
+ */
+template <typename index_t,
+          typename T,
+          typename MultiLinestringRange1,
+          typename MultiLinestringRange2>
+std::pair<linestring_intersection_intermediates<vec_2d<T>, index_t>,
+          linestring_intersection_intermediates<segment<T>, index_t>>
```
isVoid, line 234: These are owning objects, so should be wrapped with `unique_ptr` to avoid double free bugs.
```suggestion
std::pair<std::unique_ptr<linestring_intersection_intermediates<vec_2d<T>, index_t>>,
          std::unique_ptr<linestring_intersection_intermediates<segment<T>, index_t>>>
```

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
harrism: I think clang-tidy would complain if and when we set it up. ;)

isVoid: I see - here I updated it with array accessor

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: In fact, I think the common practice (at least in RAPIDS) is to set each of the owning member as unique pointers. This is partly because the cudf's column utilities is built around `std::unique_ptr<cudf::column>` not `cudf::column`. The effect kinda propagates all the way to all internals.

isVoid: I think a more common option here is to use `std::unique_ptr` for all members to provide flexibility on partial moves.

isVoid: Done. I wrapped all non-temporary, owning objects with `unique_ptr`.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,301 @@
...
+  id_ranges(IntegerRange lhs_linestring_ids,
+            IntegerRange lhs_segment_ids,
+            IntegerRange rhs_linestring_ids,
+            IntegerRange rhs_segment_ids)
+    : lhs_linestring_ids(lhs_linestring_ids),
+      lhs_segment_ids(lhs_segment_ids),
+      rhs_linestring_ids(rhs_linestring_ids),
+      rhs_segment_ids(rhs_segment_ids)
+  {
+  }
```
thomcom, line 56: Is this something that should be written into another utility range object?

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,301 @@
...
+                                                        IdRanges point_ids_range,
+                                                        IdRanges segment_ids_range,
+                                                        OutputIt1 points_first,
+                                                        OutputIt2 segments_first)
+{
+  using T       = typename MultiLinestringRange1::element_t;
+  using types_t = uint8_t;
+  using count_t = iterator_value_type<Offsets1>;
+  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < multilinestrings1.num_points();
+       idx += gridDim.x * blockDim.x) {
```
thomcom, line 176: Isn't there a utility function for this?

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,301 @@
...
+                                rhs_linestring_idx,
+                                rhs_segment_idx);
+        }
+      }
+    }
+  }
+}
+
+/**
+ * @brief Compute intersections between multilnestrings with duplicates.
```
thomcom, line 228:
```suggestion
 * @brief Compute intersections between multilinestrings with duplicates.
```

thomcom: What does it mean for a multilinestring to have duplicates?

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,301 @@
...
+ */
+template <typename index_t,
+          typename T,
+          typename MultiLinestringRange1,
+          typename MultiLinestringRange2>
+std::pair<linestring_intersection_intermediates<vec_2d<T>, index_t>,
+          linestring_intersection_intermediates<segment<T>, index_t>>
+pairwise_linestring_intersection_with_duplicate(MultiLinestringRange1 multilinestrings1,
```
thomcom, line 236: Suggest renaming this to `_with_duplicates` to match filename.

thomcom (changes requested):

cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
```diff
@@ -0,0 +1,411 @@
...
+                                      *std::move(points.rhs_segment_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_lhs_linestring_ids,
+                                      *std::move(segments.lhs_linestring_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_lhs_segment_ids,
+                                      *std::move(segments.lhs_segment_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_rhs_linestring_ids,
+                                      *std::move(segments.rhs_linestring_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_rhs_segment_ids,
+                                      *std::move(segments.rhs_segment_ids));
+}
```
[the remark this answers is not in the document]
isVoid: Sounds good. I will add a few more tests on simpler cases.

cpp/include/cuspatial/experimental/ranges/range.cuh
```diff
@@ -0,0 +1,66 @@
...
+  range(IteratorType begin, IteratorType end) : _begin(begin), _end(end) {}
+
+  /// Return the start iterator to the range
+  auto CUSPATIAL_HOST_DEVICE begin() { return _begin; }
+  /// Return the end iterator to the range
+  auto CUSPATIAL_HOST_DEVICE end() { return _end; }
+  /// Return the size of the range
+  auto CUSPATIAL_HOST_DEVICE size() { return thrust::distance(_begin, _end); }
+
+  /// Access the `i`th element in the range
```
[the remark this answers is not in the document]
isVoid: Yes, this is one of the ways to document member functions.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: No, this is something very specific to this internal function. `id_ranges` is just a convenience accessor for `intersection_intermediates` struct. You can represent this as a range of ranges, but readability I think what we have now is sufficient.

cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
```diff
@@ -0,0 +1,411 @@
...
+    {0, 2, 5, 7, 12, 16, 18, 20},
+    {P{1, 0},       P{0, 1},     P{0.5, 0},    P{0, 0.5},     P{1, 0.5},
+     P{0.5, 0.5},   P{1.5, 1.5}, P{-1, -1},    P{0.25, 0.25}, P{0.25, 0.0},
+     P{0.75, 0.75}, P{1.5, 1.5}, P{0.25, 0.0}, P{0.25, 0.5},  P{0.75, 0.75},
+     P{1.5, 1.5},   P{2, 2},     P{3, 3},      P{1, 0},       P{2, 0}});
+
+  auto [points, segments] = detail::pairwise_linestring_intersection_with_duplicate<index_t, T>(
+    multilinestrings1.range(), multilinestrings2.range(), this->mr(), this->stream());
+
+  auto expected_points_offsets   = make_device_vector<index_t>({0, 1, 3, 3, 5, 7, 7, 7});
```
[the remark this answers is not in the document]
isVoid: Think of this as a multipoint series. Each row is the intersection result of the corresponding multilinestring. See `intersection_intermediates` docstring for clarification on output format: https://github.com/rapidsai/cuspatial/blob/8534f1704b7892dc3f3cf769a73e497b4e2f25b0/cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh#L81-L89

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: Ha! Mark asked the similar question here: https://github.com/rapidsai/cuspatial/pull/813#discussion_r1034163231. Short answer is no. But we can create a short hand for it with `counting_iterator` and `range`. Want me to do so?

isVoid: Or introduce https://github.com/harrism/ranger to the code base.

isVoid: I think it's worth making an issue to track this. But probably outside of the scope of this PR.

cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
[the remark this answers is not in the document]
isVoid: Done.

thomcom (approved):

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
harrism: But do you have a kernel that only calls the first?

cpp/include/doxygen_groups.h
```diff
@@ -114,10 +114,12 @@
  *      @}
  *      @defgroup ranges Ranges
  *      @{
- *          @brief Non owning object of geometry arrays
+ *          @brief Abstract Data Type that Represents any containers represented by a start and end
+ *          iterator
  *
  *          Cuspatial header only APIs accepts ranges that are composed of flattened view of
  *          offsets and coordinates. Various accessors are provided for flexible access to the data.
```
harrism, line 121:
```suggestion
 *          cuSpatial header only APIs accept ranges that provide flattened views of
 *          offsets and coordinates. Various accessors are provided for flexible access to the data.
```

cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
```diff
@@ -0,0 +1,812 @@
...
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_point_rhs_segment_ids,
+                                      *std::move(points.rhs_segment_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_lhs_linestring_ids,
+                                      *std::move(segments.lhs_linestring_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_lhs_segment_ids,
+                                      *std::move(segments.lhs_segment_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_rhs_linestring_ids,
+                                      *std::move(segments.rhs_linestring_ids));
+  CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(expected_segment_rhs_segment_ids,
+                                      *std::move(segments.rhs_segment_ids));
```
harrism, line 522: A lot of duplication in these tests. Perhaps a test function that parameterizes the things that change between tests and then call that function from each test.

harrism (changes requested):
Looks good. This is pretty complicated, so I tried to provide a few ideas for simplification / DRY.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,302 @@
...
+    auto const geometry_idx       = multilinestrings1.geometry_idx_from_part_idx(part_idx);
+    auto const multilinestring2   = multilinestrings2[geometry_idx];
+    auto const geometry_collection_offset =
+      num_points_offsets_first[geometry_idx] + num_segments_offsets_first[geometry_idx];
+
+    for (auto rhs_linestring_idx = 0; rhs_linestring_idx < multilinestring2.size();
+         ++rhs_linestring_idx) {
+      auto const linestring2 = multilinestring2[rhs_linestring_idx];
+      for (auto rhs_segment_idx = 0; rhs_segment_idx < linestring2.num_segments();
+           ++rhs_segment_idx) {
```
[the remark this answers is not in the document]
isVoid: Yes, unfortunately. I have thought about adding additional information in `segment<T>` struct to store the source ids. But I don't think that's a common enough use case..

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: > But do you have a kernel that only calls the first?

Currently no. In fact, giving this a second thought. The only case where we need to call `part_idx_from_point_idx` is when the threads are launched at the bottom of the geometry hierarchy and need to traverse the hierarchy bottom-up. When this happens we almost always need to call both function. So perhaps yes, we can merge these two helpers.

cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
harrism: OK, but you still get line numbers, and combined with the failing test name you can easily jump to the test case that failed and which vector mismatched within it. 

But even if you duplicate the EXPECT macro calls, there seems to be a lot of unnecessary duplication in the declaration of all the test and expected vectors.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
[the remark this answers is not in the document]
harrism:
```
for (auto rhs_segment: multilinestring2.segments_with_indices()) { ... }
```

`segments_with_indices` range getter would return `tuple<segment, linestring_idx, segment_idx>`?

Or even just provide `segment_indices` range getter that just returns a `pair` of the indices.  Just brainstorming.

cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
isVoid: That is because the tests are... duplicated! Yeah I agree the duplication is a bit hard to read. I'll work on some refactor on the next round of update.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: Actually, I would add to the above: although the thread working in the geometry hierarchy bottom up, it only needs to call `is_valid_segment_id` when the thread is treating the linestring as per segment-basis. 

There are, however, other situation where we might want to treat the treat the linestring as a mere collection of points. (The recent `contains_properly` is an example for that!) In these situation we don't need to call `is_valid_segment_id` after the searching the `part_id`. It's just up to this point, APIs that uses the new `multilienstring_range` objects hasn't come across these API requirements yet.

On that note, I would recommend adding a new `part_idx_from_segment_idx`, implying that the argument is indeed id for the segment, not points. It can then return a `nullopt` if the segment id is invalid in the multilinestring.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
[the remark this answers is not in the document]
isVoid: What you suggested above would require using a `permutation_iterator` to flatten a 2-level nested loop into a single level, with the range size of the level becoming `N*M`, where `N` is the length of the first level and `M` is the second. This is promising work to take on but I think it's a bit out of scope of this PR rn.

To simplify things a bit. Creating something like a `enumerate_range` (referring to python's [enumerate](https://docs.python.org/3/library/functions.html#enumerate) builtin) would be an interesting task.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,302 @@
...
+
+  // Allocate a temporary vector for each thread to keep track of the number of results
+  // from the current multilinestring pair has written.
+  rmm::device_uvector<index_t> num_points_stored_temp(num_pairs, stream);
+  rmm::device_uvector<index_t> num_segments_stored_temp(num_pairs, stream);
+
+  thrust::uninitialized_fill_n(
+    rmm::exec_policy(stream), num_points_stored_temp.begin(), num_pairs, 0);
+  thrust::uninitialized_fill_n(
+    rmm::exec_policy(stream), num_segments_stored_temp.begin(), num_pairs, 0);
```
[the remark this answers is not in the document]
isVoid: Why isn't `thrust::unintialized_fill_n` using `cudaMemsetAsync` under the hood?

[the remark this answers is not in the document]
isVoid: By saying policy, do you mean stream here?

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -0,0 +1,301 @@
...
+                                                        IdRanges point_ids_range,
+                                                        IdRanges segment_ids_range,
+                                                        OutputIt1 points_first,
+                                                        OutputIt2 segments_first)
+{
+  using T       = typename MultiLinestringRange1::element_t;
+  using types_t = uint8_t;
+  using count_t = iterator_value_type<Offsets1>;
+
+  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < multilinestrings1.num_points();
```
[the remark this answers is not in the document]
isVoid: Similarly:
https://github.com/rapidsai/cuspatial/blob/138dc69cf8745b0e51380ee984bf4065006d96eb/cpp/include/cuspatial/experimental/detail/linestring_intersection_count.cuh#L41

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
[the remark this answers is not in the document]
harrism: I guess because not everything you might fill is just an array of bytes. And you can't use `cudaMemsetAsync` to fill anything that is not the same value in every *byte*. E.g. if you wanted to fill an array of ints with the value 1, then you can't use `cudaMemsetAsync`.

cpp/include/cuspatial/experimental/detail/ranges/enumerate_range.cuh
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @brief Given Any Range, zip the iterator with the index of the element.
```
harrism, line 18: Does this standalone brief work without @file?

cpp/include/cuspatial/detail/utility/zero_data.cuh
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @brief Zero set a data range on device
+ *
+ * @tparam Iterator type of iterator to the range
+ * @param begin Start iterator to the range
+ * @param end End iterator to the range
+ * @param stream CUDA stream used for device memory operations and kernel launches
+ */
+template <typename Iterator>
+void zero_data(Iterator begin, Iterator end, rmm::cuda_stream_view stream)
```
harrism, line 34: This function should have `async` in the name, I think.  Also consider calling it `zero_range_async`, but I'm fine with either.

```suggestion
void zero_data_async(Iterator begin, Iterator end, rmm::cuda_stream_view stream)
```

harrism (approved):
I have one naming request but otherwise approve. Love the enumerated_range stuff!

cpp/include/cuspatial/experimental/detail/ranges/enumerate_range.cuh
isVoid: I don't think so, but at the same time, I'm not sure if we should expose this to public API? See I left it in `detail`.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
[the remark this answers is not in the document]
isVoid: Maybe not a bug.. I might confused myself here. Both kernels are launched with number of threads equal to the number of segments, so it's ok to call `num_points()` here.

cpp/include/cuspatial/experimental/detail/ranges/enumerate_range.cuh
harrism: I think delete this standalone comment block. The one on the function should be sufficient.

isVoid: Removed

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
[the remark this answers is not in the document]
isVoid: Oh I see. That's why memset is more efficient for zero-setting since every byte is the same.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: `Intersection_intermediate` itself isn't wrapped with unique_ptr. But all of its member variables are.

thomcom:
rerun tests

isVoid:
rerun tests

isVoid:
@gpucibot merge

PR #819, opened 2022-11-23
Add `find_point_on_segment` internal API

### LINKED ISSUE #817: Create Internal API to Merge Points to Segments if Point is on the Segment
As part of #765 , given a list of points and a list of segments, remove the points if the points are on the segment.

isVoid:
## Description

This PR adds internal API to find if point can be merged on corresponding segment sets. Closes #817.

Depending on other internal API PRs to add `range` and update `range` objects to make this pass compile.

## Checklist


cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
```diff
@@ -0,0 +1,95 @@
...
+          typename SegmentsRange,
+          typename OutputIt1>
+void find_point_on_segments(MultiPointRange multipoints,
+                            OffsetsRange segment_offsets,
+                            SegmentsRange segments,
+                            OutputIt1 mergeable_flag,
+                            rmm::cuda_stream_view stream)
+{
+  using index_t     = typename MultiPointRange::index_t;
+  auto num_segments = segment_offsets.size() - 1;
```
harrism, line 82: Seems like something called `SegmentsRange` would be able to tell you the number of segments without having to subtract one...

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
```diff
@@ -0,0 +1,95 @@
...
+  uint8_t __device__ operator()(IndexType i)
+  {
+    auto point        = thrust::raw_reference_cast(multipoints.point_begin()[i]);
+    auto geometry_idx = multipoints.geometry_idx_from_point_idx(i);
+
+    for (auto segment_idx = segment_offsets[geometry_idx];
+         segment_idx < segment_offsets[geometry_idx + 1];
+         segment_idx++) {
+      auto const& segment = thrust::raw_reference_cast(segments[segment_idx]);
+      if (point_on_segment(segment, point)) return 1;
```
harrism, line 61: Why not true/false instead of 1/0?

```suggestion
      return is_point_on_segment(segment, point);
```

harrism (changes requested):
Some small suggestions.

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
isVoid: I thought about introducing something like this, as this will also simplify the input to this method. Later I punted to doing so because I don't want the type of ranges to explode. 

Also, I think a more generalized introduction is a templated `NestedRange`. We can use this to generalize anything that's list like - and that would also include the existing `multipoint_ranges/multilinestring_ranges` etc..

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
isVoid: Hmm, wonder if it's because of output type casting. Let me see if returning boolean works.

isVoid: Wait, this isn't the same logic. For each point we loop over all segments and if point is on any of them you early exit the thread and return 1.

isVoid: But still, can try returning boolean here.

cpp/include/cuspatial/detail/utility/linestring.cuh
```diff
@@ -213,5 +213,21 @@ segment_intersection(segment<T> const& segment1, segment<T> const& segment2)
...
+ * @brief Given a point and a segment, returns true if point is on the segment.
+ */
+template <typename T>
+bool __device__ point_on_segment(segment<T> const& segment, vec_2d<T> const& c)
+{
+  auto [a, b] = segment;
+  auto ab     = b - a;
+  auto ac     = c - a;
+
+  if (not_float_equal(det(ab, ac), T{0})) return false;
```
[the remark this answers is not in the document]
isVoid: Removed.

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
harrism: That's not really my point. What is `SegmentsRange`? If it's actually a range of segments, then `segments.size()` should be the number of segments. So you could just delete this line and use `segments.size()` on the RHS of your CUSPATIAL_EXPECTS.

isVoid: Oh, I see. Yeah using `segments.size()` is better here.

harrism (approved):

thomcom (approved):
Looks good.

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
isVoid: Has to use `segment_offsets.size() -  1`. Here it is comparing the number of multipoints and the number of "multi-segments".

isVoid:
@gpucibot merge

Issue #722, opened 2022-10-03
[FEA] Implement binary predicates that depend only on point-in-polygon


thomcom:
**Is your feature request related to a problem? Please describe.**
Some spatial predicates depend only on the outcome of point-in-polygon.

**Describe the solution you'd like**
I can work on these predicates in python while the `equals` and `intersects` basic predicates are being worked on.

Equality/Intersection/PIP table
Legend:
x: False/impossible
e: uses equality
i: uses intersection
p: uses pip

  | x(p, p) | x(p, l) | x(p, P) | x(l, p) | x(l, l) | x(l, P) | x(P, p) | x(P, l) | x(P, P)
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Equals | e | x | x | x | e | e | e | e | e
Disjoint | e | e | e | i | i | ip | ep | ep | ep
Touches | e | e | p | i | i | ip | ip | ip | ip
Contains | e | x | x | i | e | x | p | p | p
Covers | e | x | x | i | e | x | ip | ip | ip
Intersects | e | e | p | i | i | i | ip | ip | ip
Within | e | e | p | x | e | p | x | x | p
CoveredBy | e | e | p | x | e | p | x | x | p
Crosses | e | x | x | i | i | i | i | i | i
Overlaps | e | x | x | i | i | i | p | i | i

https://github.com/rapidsai/cuspatial/issues/722#tasklist-block-68684f15-c10b-4537-8c90-e2423e25a899

harrism:
@thomcom please add a tasklist and use this issue as an epic.

thomcom:
I implemented them all yesterday morning. They are implemented insofar as only point-in-polygon, with the exception of boundary testing, is required.

harrism:
Goal of task lists is for each task to become an issue to give one place.to track an epic. If these are all done in one go, probably didn't need a task list. :)

thomcom:
It was shorter and easier than I expected - I didn't end up making any separate issues out of it, just writing them up and testing them as I went. I'm working on a refactor that will close this issue hopefully soon.

PR #834, opened 2022-12-01
Implement all binary operations that depend only on point-in-polygon

### LINKED ISSUE #835: Special `contains_properly` cases polygon overlaps point, and polygon overlaps polygon.
The listed two cases of `.contains_properly` use a different counting method: _any_ point in polygon instead of _all_ points in polygon.

This is the first unique case in the effort to refactor cuSpatial binops to handle all parameterizations, not just `.contains_properly`. In order to handle it, the refactoring needs to include three steps:

- 1. Preprocess the lhs and rhs. This includes alignment and reordering, as some cases are inversions of each other. This also includes restructuring the lhs and rhs data for calling the actual binop, which is the last step before hitting Cython and may have a significantly different call signature than `GeoSeries, GeoSeries`.
- 2. Call the binop with the preprocessed lhs and rhs.
- 3. Postprocess the result. The PR for this issue implements a disambiguation that uses  _any_ point in polygon to compute true results, instead of _all_ points in polygon.

thomcom:
Closes https://github.com/rapidsai/cuspatial/issues/722
Closes #835 

## Description
This PR begins the refactoring of binary operations code to support every binary operation and combination of inputs. A new test file is created, `tests/binops/test_pip_only_binpreds.py` that contains a size 1 and a size 31 case for each binop combination that should work with only `.contains_properly`. Three binary predicates are partially implemented in this PR: overlaps, intersects, and within.

The following tests are implemented in this binpred:
- test_polygon_overlaps_point
- test_max_polygons_overlaps_max_points
- test_polygon_overlaps_polygon
- test_max_polygon_overlaps_max_points
- test_point_intersects_polygon_interior
- test_max_points_intersects_max_polygons_interior
- test_point_within_polygon
- test_max_points_within_max_polygons
- test_linestring_within_polygon
- test_max_linestring_within_max_polygon
- test_polygon_within_polygon
- test_max_polygons_within_max_polygons
- test_polygon_overlaps_linestring
- test_max_polygons_overlaps_max_linestrings

## Checklist


thomcom:
rerun tests

python/cuspatial/cuspatial/tests/binops/test_pip_only_binops.py
```diff
@@ -0,0 +1,190 @@
+import geopandas as gpd
```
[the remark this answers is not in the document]
isVoid: Also, needs license.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -168,9 +170,12 @@ def _get_current_features(self, type):
         def point_indices(self):
             # Return a cupy.ndarray containing the index values that each
             # point belongs to.
+            """
             offsets = cp.arange(0, len(self.xy) + 1, 2)
             sizes = offsets[1:] - offsets[:-1]
             return cp.repeat(self._series.index, sizes)
+            """
+            return self._series.index
```
isVoid, line 178: The length of index can be different from the same as number of points right? Unless I misunderstood, this seems wrong to me.

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+from cuspatial.utils.column_utils import (
+    contains_only_linestrings,
+    contains_only_multipoints,
+    contains_only_points,
+    contains_only_polygons,
+    has_same_dimension,
+)
+
+
+class _binop(ABC):
```
isVoid, line 18: Rename this to `_BinaryPredicate`?

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+
+        Returns
+        -------
+        cudf.Series
+            The output of the post processing, True/False results for
+            the specified binary op.
+        """
+        pass
+
+    def __init__(self, op, lhs, rhs, align=True):
```
isVoid, line 73: It comes strange to me that you would perform the heavy computation inside the init function. Usually in init you would just store the necessary state for the object and do compute in `__call__`.

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+        - `.touches`
+        - `.contains`
+        - `.contains_properly`
+        - `.covers`
+        - `.intersects`
+        - `.within`
+        - `.crosses`
+        - `.overlaps`
+
+        There are twenty-five ordering combinations of `lhs` and `rhs`:
```
isVoid, line 88: There are 36 and I don't think it's necessary to list all of them.

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+            lhs.polygons.ring_offset[:-1],
+            lhs.polygons.x,
+            lhs.polygons.y,
+        )
+        return point_result
+
+    def equals(self, lhs, rhs):
+        """Compute the equals relationship between two GeoSeries."""
+        if contains_only_points(lhs):
+            result = lhs.points.xy.equals(rhs.points.xy)
```
isVoid, line 192: Are we assuming `rhs` are also points?

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+        elif contains_only_linestrings(lhs):
+            result = lhs.lines.xy.equals(rhs.lines.xy)
```
isVoid, line 194: Also, the points in both geometries may not be in the same order...

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+            result = lhs.polygons.xy.equals(rhs.polygons.xy)
+        elif contains_only_multipoints(lhs):
+            result = lhs.multipoints.xy.equals(rhs.multipoints.xy)
+        return result
+
+    def touches(self, lhs, rhs):
+        """Compute the touches relationship between two GeoSeries:
+        Two objects touch if they share at least one point in common, but their
+        interiors do not intersect."""
+        return self.equals(lhs, rhs)
```
isVoid, line 205: Eh, I don't see preprocess and postprocess to touches. I think it takes more than just `equals`...

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+        point_result = cudf.Series(
+            df_result["pip"], index=cudf.RangeIndex(0, len(df_result))
+        )
+        point_result.name = None
+        return point_result
+
+
+class EqualsBinop(_binop):
+    def preprocess(self, op, lhs, rhs):
+        return super().preprocess(op, lhs, rhs)
```
isVoid, line 405:
```suggestion
```
If it does nothing, you don't need to write it :)

python/cuspatial/cuspatial/core/binops/binops.py
```diff
@@ -0,0 +1,409 @@
...
+        Postprocess converts the raw results of the binary operation into
+        the final result. This is where the discrete math rules are applied.
+
+        Parameters
+        ----------
+        op : str
+            The binary operation to post process. Determines for example the
+            set operation to use for computing the result.
+        processed : cudf.Series
+            The data after applying the fundamental binary operation.
```
isVoid, line 63: Not the right documentation for inputs.

isVoid (changes requested):
Initial pass to the code. Strategy pattern does makes sense here.

python/cuspatial/cuspatial/core/binops/binops.py
harrism: Agree. binop typically means arithmetic operations.

python/cuspatial/cuspatial/utils/column_utils.py
```diff
@@ -97,3 +97,20 @@ def contains_only_polygons(gs: GeoSeries):
     """
 
     return contain_single_type_geometry(gs) and len(gs.polygons.xy) > 0
+
+
+def has_same_dimension(lhs: GeoSeries, rhs: GeoSeries):
```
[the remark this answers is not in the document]
thomcom: I will change it.

python/cuspatial/cuspatial/tests/binops/test_pip_only_binops.py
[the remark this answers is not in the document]
thomcom: I think the test names are highly descriptive:

`test_polygon_overlaps_point`
`test_max_polygons_overlaps_max_points`
`test_polygon_overlaps_polygon`
`test_linestring_within_polygon`
etc...
Do you think it needs to have these listed in a comment at the beginning of the file?

python/cuspatial/cuspatial/tests/binops/test_pip_only_binops.py
```diff
@@ -0,0 +1,190 @@
...
+    gpdpolygon = gpd.GeoSeries(Polygon([[0, 0], [0, 1], [1, 1], [0, 0]]))
+    point = cuspatial.from_geopandas(gpdpoint)
+    polygon = cuspatial.from_geopandas(gpdpolygon)
+    got = point.intersects(polygon).values_host
+    expected = gpdpoint.intersects(gpdpolygon).values
+    assert (got == expected).all()
+"""
+
+
+# Needs any instead of all.
```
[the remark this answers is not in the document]
thomcom: I'm going to delete the commented tests instead. They are just mis-entries from the Basic Predicates Table.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: That's a good point. I should do something that is more obvious instead of something weird, which is to compute all the results in the `__init__` and then just return them in `__call__`. There may have been a design decision there that I'll encounter when I refactor it.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: I was thinking of using the list to track their implementations. You're right they certainly don't need to be enumerated in the public docstring! Thanks. I'll revisit this doc.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: Yes, by definition if the lhs and rhs are different types they are not equal. I have added a default `False` return value which I previously overlooked.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: This is essentially crust - at one time I thought I would write some cases for `touches` but found that equality doesn't work for any cases of `touches`. I left it in because I'll be writing it soon.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: In that case it had to be written because it is an `@abstractmethod` and has to have one in order to inherit from `_binop(ABC)`. For now I've refactored it to inherit from `ContainsProperlyBinpred` instead, but I will iterate it again, especially in the next PRs as it gets implemented more thoroughly.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: What this was doing was duplicating the index, so that for a buffer of points the `point_indices` would be `[0, 0, 1, 1, 2, 2, ..., n, n]`, then in the final logic I would `groupby` those indices, then do some comparison on that. It was very inefficient for points since they always have the same offset. For `pip` at least, only one result is returned for each point, each pair of xy coords stored in interleaved format.

thomcom: I don't think we'll ever have an operation that works on a level more granular than points, so keeping two indices, `[0, 0]` for example for a single point, is necessary. Two values in points = 1 index.

python/cuspatial/cuspatial/tests/binops/test_pip_only_binops.py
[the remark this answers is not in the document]
thomcom: It was just a note to myself as I was writing the tests, I removed it.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: I am writing a test for this now, what orderings of LineStrings GeoPandas considers to be equal.

thomcom: Hmm. Now that I dig into this deeper I'm reminded that `equals` is in the next PR, not this one. I think I got ahead of myself on the PR and wrote functionality that isn't intended to be tested in this branch.

thomcom: Equals functionality is in https://github.com/rapidsai/cuspatial/issues/838. Do you want  me to carefully remove the equals functionality from this branch, or do you mind ignoring it until the next PR?

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: Also this will appear correctly supported in the next PR https://github.com/rapidsai/cuspatial/issues/838

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: Ok, I moved the computation into `__call__` and cleaned up the factoring another iteration.

python/cuspatial/cuspatial/tests/binops/test_pip_only_binops.py
```diff
@@ -0,0 +1,190 @@
...
+    gpdpoint = gpd.GeoSeries([*point_generator(31)])
+    polygon = cuspatial.from_geopandas(gpdpolygon)
+    point = cuspatial.from_geopandas(gpdpoint)
+    got = polygon.overlaps(point).values_host
+    expected = gpdpolygon.overlaps(gpdpoint).values
+    assert (got == expected).all()
+
+
+# Needs any instead of all.
+def test_polygon_overlaps_polygon():
```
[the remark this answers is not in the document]
thomcom: I added tests for no overlap, and complete overlap and already had one for partial overlap, and found a bug.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: Removed.

python/cuspatial/cuspatial/core/binops/binops.py
thomcom: Removed

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -85,43 +87,12 @@ def __init__(self, op, lhs, rhs, align=True):
...
-        - multipolygon linestring
-        - multipolygon multilinestring
-        - multipolygon polygon
-        - multipolygon multipolygon
+        There are thirty-six ordering combinations of `lhs` and `rhs`, the
+        unordered pairs of each `point`, `multipoint`, `linestring`,
+        `multilinestring`, `polygon`, and `multipolygon`. The ordering of
+        `lhs` and `rhs` is important because the result of the binary
+        operation is not symmetric. For example, `A.contains(B)` is not
+        necessarily the same as `B.within(A)`.
```
isVoid, line 95: Why not? Shapely says yes: https://shapely.readthedocs.io/en/stable/manual.html#object.contains

isVoid (changes requested):
Where's `.contains_properly`?

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -324,35 +225,33 @@ def postprocess(self, op, point_indices, point_result):
...
-    def postprocess(self, op, point_indices, point_result):
-        """Postprocess the output GeoSeries to ensure that they are of the
-        correct type for the operation."""
-        return super().postprocess(op, point_indices, point_result)
+        return super().preprocess(lhs, rhs)
 
 
-class WithinBinop(ContainsProperlyBinop):
-    def preprocess(self, op, lhs, rhs):
+class WithinBinpred(ContainsProperlyBinpred):
```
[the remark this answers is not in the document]
isVoid: Ah I realized `.within` may not be the inverse of `.contains_properly`. Nevermind about this comment.

python/cuspatial/cuspatial/tests/binops/test_pip_only_binops.py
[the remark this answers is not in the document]
isVoid: Yup, that's also fine.

python/cuspatial/cuspatial/core/geoseries.py
isVoid: Ok, thanks for explaining. I did not notice this is inside `GeoColumnAccessor`, and that `self._series` is the point series. But can you remind me where is this index computed?

thomcom: As the index for the `GeoSeries`, I think it will typically just be a `RangeIndex`. As we discuss this I realize that this approach may only work for homogeneous `Point` `GeoSeries`. If it is a heterogeneous `GeoSeries` then the index values won't all refer to `Point`s, and will produce undesirable results.

isVoid: I thought what you want here is not the RangeIndex, but the index of the point that can be used as group-by keys? For example:
```
Offset: [0, 2, 3, 4]
Points: [(0.0, 0.0), (1.0, 1.0), (2.0, 2.0), (3.0, 3.0)]
Index:  [0, 0, 1, 2]
```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -279,7 +162,28 @@ def preprocess(self, op, lhs, rhs):
...
 
-    def postprocess(self, op, point_indices, point_result):
+    def _op(self, lhs, points):
+        """Compute the contains_properly relationship between two GeoSeries.
+        A feature A contains another feature B if no points of B lie in the
+        exterior of A, and at least one point of the interior of B lies in the
+        interior of A. This is the inverse of `within`."""
+        if not contains_only_polygons(lhs):
+            raise TypeError(
+                "`.contains` can only be called with polygon series."
```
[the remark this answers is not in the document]
thomcom: I'm interested in this use case - I definitely want to support 1-to-many operations in the future. For now just using `NotImplementedError` is a good idea.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: Yeah I'll fix this up.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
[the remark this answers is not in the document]
thomcom: Ok, I'll need to think about this. At this level, the `lhs` has to be polygon(s) because I'm about to call `lhs.polygons...` and the lhs must contain polygons. `.contains/.contains_properly` will be implemented for the other types as well, but they will be shortcutted during `preprocess` to another Binpred that doesn't require point-in-polygon for its result.

This involves part of the refactoring decision in the strategy pattern. The binpreds need to be able to call each other at different stages of other `preprocess, _op, or postprocess`. I haven't worked out how to design that in software most simply yet.

For now, `TypeError` here remains appropriate because in order to each this point the `lhs` must be polygon(s). Ok?

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # Postprocess: Apply discrete math rules to identify relationships.
+        return self.postprocess(indices, point_result)
+
+
+class ContainsProperlyBinpred(BinaryPredicate):
+    def preprocess(self, lhs, rhs):
+        """Preprocess the input GeoSeries to ensure that they are of the
+        correct type for the operation."""
+        # Type determines discrete math recombination outcome
```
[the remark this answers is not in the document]
thomcom: That makes sense, I don't think the comment fit well with the next section of code. I've updated the description to be more clear.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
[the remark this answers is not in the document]
thomcom: Actually I'll move the same check up a level higher because I understand that this suggests that maybe the other combinations will not be implemented. They are not implemented yet because they do not depend on point-in-polygon.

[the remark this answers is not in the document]
thomcom: When intersect is merged, my plan is to iterate over each of the binary predicates, filling out the types matrix for each one. I'll create new documentation that details progress through each.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: Yes, the RangeIndex becomes a regular series when it becomes a column of `result` before the groupby. Points don't need to be grouped, however, their indices are all singular. A pip result that uses a geometry other than `Point` needs to have the `n` points grouped into a single result, for point the result is always a single.

isVoid: Got it, thanks for explaining.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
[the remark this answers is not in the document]
isVoid: From offline: let's add a docstring to `BinaryPredicate` to allow user to track the supported geometries.

isVoid (approved):

thomcom:
rerun tests

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Parameters
+        ----------
+        op : str
+            The binary operation to perform.
+        lhs : GeoSeries
+            The left-hand-side of the GeoSeries-level binary operation.
+        rhs : GeoSeries
+            The right-hand-side of the GeoSeries-level binary operation.
+
```
harrism, line 36:
```suggestion
            The left-hand-side of the GeoSeries-level binary predicate.
        rhs : GeoSeries
            The right-hand-side of the GeoSeries-level binary predicate.

```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -0,0 +1,272 @@
...
+        Returns
+        -------
+        GeoSeries
+            The left-hand-side of the internal binary operation, may be
+            reordered.
+        GeoSeries
+            The right-hand-side of the internal binary operation, may be
```
harrism, line 43: predicate

harrism: And everywhere else in the docs...

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -768,63 +771,66 @@ def contains_properly(self, other, align=True):
...
-                result.groupby("idx").sum().sort_index()
-                == result.groupby("idx").count().sort_index()
-            )
-            point_result = cudf.Series(
-                df_result["pip"], index=cudf.RangeIndex(0, len(df_result))
-            )
-            point_result.name = None
-        return point_result
+    def intersects(self, other, align=True):
+        """Compute the intersections of two GeoSeries.
```
harrism, line 777: This one just says what it computes, doesn't define intersection.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -768,63 +771,66 @@ def contains_properly(self, other, align=True):
...
+        Returns
+        -------
+        result : cudf.Series
+            A Series of boolean values indicating whether the geometries of
+            each row intersect.
+        """
+        return IntersectsBinpred(self, other, align)()
+
+    def within(self, other, align=True):
+        """An object is said to be within other if at least one of its points
```
harrism, line 795: This one just defines within, doesn't start with "computes".

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -768,63 +771,66 @@ def contains_properly(self, other, align=True):
...
+        Returns
+        -------
+        result : cudf.Series
+            A Series of boolean values indicating whether each feature falls
+            within the corresponding polygon in the input.
+        """
+        return WithinBinpred(self, other, align)()
+
+    def overlaps(self, other, align=True):
+        """Returns True for all aligned geometries that overlap other, else
```
harrism, line 815: This one starts with saying what it returns, then defines overlap.

For all of my comments that start with "this one", please make the documentation consistent. Please just emulate the GeoPandas docs. :)

harrism (changes requested):

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: Good idea.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: I updated them to have the same short and long text as GeoPandas.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -768,63 +774,73 @@ def contains_properly(self, other, align=True):
...
-                df_result["pip"], index=cudf.RangeIndex(0, len(df_result))
-            )
-            point_result.name = None
-        return point_result
+    def intersects(self, other, align=True):
+        """Returns a `Series` of `dtype('bool')` with value `True` for each
+        aligned geometry that intersects _other_.
+
+        An object is said to intersect _other_ if its _boundary_ and
+        _interior_ intersects in any way with those of other.
```
harrism, line 784: "boundary and interior", or "boundary or interior"? Also typo.

```suggestion
        An object is said to intersect _other_ if its _boundary_ and
        _interior_ intersect in any way with those of other.
```

harrism: I don't like referring to  the parameter _other_ for this relationship definition since it's actually individual geometries within other that get tested.

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -768,63 +774,73 @@ def contains_properly(self, other, align=True):
...
+        """
+        return IntersectsBinpred(self, other, align)()
+
+    def within(self, other, align=True):
+        """Returns a `Series` of `dtype('bool')` with value `True` for each
+        aligned geometry that is within _other_.
+
+        An object is said to be within other if at least one of its points
+        is located in the interior and no points are located in the exterior
+        of the other.
```
harrism, line 807:
```suggestion
        A geometry is within another if at least one of its points
        is located in the interior of the other geometry and no points are located in the exterior
        of the other geometry.
```

python/cuspatial/cuspatial/core/geoseries.py
harrism:
```suggestion
        A geometry intersects another geometry if its boundary or
        interior intersect in any way with those of the other geometry.
```

harrism (approved):
More doc suggestions, but pre-approving.

python/cuspatial/cuspatial/core/geoseries.py
thomcom: Done

python/cuspatial/cuspatial/core/geoseries.py
thomcom: Done

thomcom:
@gpucibot merge

PR #836, opened 2022-12-02
Add GitHub Actions Workflows

bdice:
## Description

This PR adds GitHub Actions workflows to `cuspatial`.

**Wait until after the 22.12 release is complete (and potentially until other repos are ready to migrate to GHA) to merge.**

### Task list

Coverage required for this PR:
- [x] C++ tests
- [x] Python tests
- [x] Codecov
- [x] Style checks

Possibly in scope for this PR:
- [ ] Java tests (not currently tested in gpuCI?)
- [ ] Notebooks (not currently tested in gpuCI?)

Future work:
- Build Metrics Report (copy from cudf when it is added)
- JUnit output annotations? https://github.com/marketplace/actions/junit-report-action
- Reduce exit code handling verbosity (use global variable with `trap` function)
- Set up `rapids-dependency-file-generator` pre-commit hook
- Docs builds

## Checklist


.github/workflows/pr.yaml
```diff
@@ -1,12 +1,43 @@
...
 jobs:
+  pr-builder:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@main
   checks:
     secrets: inherit
     uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@main
+  conda-cpp-build:
+    #TODO: temporarily skip for speed
+    #needs: checks
```
bdice, line 21:
```suggestion
    needs: checks
```

cpp/CMakeLists.txt
```diff
@@ -110,7 +110,10 @@ rapids_find_package(
 rapids_cpm_init()
 # find or add cuDF
 include(cmake/thirdparty/CUSPATIAL_GetCUDF.cmake)
-
+# find or install GoogleTest
```
bdice, line 113: I'd like a critical CMake eye here. I saw failures like the following:
```
CMake Error at tests/CMakeLists.txt:33 (target_link_libraries):
  Target "QUADTREE_LINESTRING_FILTERING_TEST" links to:

    GTest::gtest_main

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  tests/CMakeLists.txt:101 (ConfigureTest)
```
I adapted this code to run `get_gtest.cmake` from libcudf.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -54,16 +58,18 @@ outputs:
     requirements:
       build:
         - cmake {{ cmake_version }}
+        - ninja
       run:
         - cudatoolkit {{ cuda_spec }}
         - libcudf {{ minor_version }}.*
         - librmm {{ minor_version }}.*
+        - libcusparse=11.7.0.107 # This version corresponds to CUDA 11.5
```
bdice, line 66: Previously, gpuCI's images included a full CTK installation. This package build was implicitly using some of the system CTK (cusparse) which is not included in `conda-forge::cudatoolkit`. This package comes from the `nvidia` channel. It isn't exactly ideal to mix `conda-forge::cudatoolkit` with `nvidia::libcusparse` but it's probably no worse than relying on the system CTK as in the previous CI images. (We have similar dependencies on `nvidia::cuda-nvtx` in `libcudf`.)

.github/workflows/pr.yaml
```diff
@@ -1,12 +1,42 @@
...
+      - "pull-request/[0-9]+"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
+  pr-builder:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@main
```
ajschmidt8, line 15: we'll need to add the `needs` section here

bdice:
> should we also add a codeowner section for dependencies.yaml?

Already exists! :) https://github.com/rapidsai/cuspatial/blob/886e1c98280aee2e1fbc7e078df38dcaaf904200/.github/CODEOWNERS#L17

harrism:
>  Java tests (not currently tested in gpuCI?)

Not sure cuSpatial Java even builds in CI currently. @jlowe do you know?

jlowe:
> Not sure cuSpatial Java even builds in CI currently. @jlowe do you know?

Unfortunately I do not know, but I suspect it is not.  I was only tangentially involved in cuSpatial Java, mostly helping to answer questions from @byin001 about the cudf Java bindings when cuSpatial Java was first created.

ci/check_style.sh
```diff
@@ -0,0 +1,25 @@
...
+rapids-dependency-file-generator \
+  --output conda \
+  --file_key checks \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee env.yaml
+
+rapids-mamba-retry env create --force -f env.yaml -n checks
+conda activate checks
+
+set +e
+
```
bdice, line 18:
```suggestion
```

conda/recipes/cuspatial/meta.yaml
```diff
@@ -1,9 +1,9 @@
-# Copyright (c) 2018, NVIDIA CORPORATION.
+# Copyright (c) 2018-2022, NVIDIA CORPORATION.
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
 {% set minor_version =  version.split('.')[0] + '.' + version.split('.')[1] %}
-{% set py_version=environ.get('CONDA_PY', 36) %}
-{% set cuda_version = '.'.join(environ.get('CUDA', '11.5').split('.')[:2]) %}
+{% set py_version=environ.get('CONDA_PY', 38) %}
+{% set cuda_version = '.'.join(environ.get('RAPIDS_CUDA_VERSION', '11.5.2').split('.')[:2]) %}
```
bdice, line 6:
```suggestion
{% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -1,8 +1,8 @@
-# Copyright (c) 2018, NVIDIA CORPORATION.
+# Copyright (c) 2018-2022, NVIDIA CORPORATION.
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
 {% set minor_version = version.split('.')[0] + '.' + version.split('.')[1] %}
-{% set cuda_version = '.'.join(environ.get('CUDA', '11.5').split('.')[:2]) %}
+{% set cuda_version = '.'.join(environ.get('RAPIDS_CUDA_VERSION', '11.5.2').split('.')[:2]) %}
```
bdice, line 5:
```suggestion
{% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
```

cpp/CMakeLists.txt
vyasr: cudf exports the gtest targets in the case where cpm downloads gtest. My suspicion is that with GHA we're doing a better job of ensuring that gtest is already present in the conda environment than we were before (or in some other way we are ensuring that it's available), and as a result cudf isn't exporting the target and it's revealing the fact that cuspatial wasn't finding it on its own. Assuming that diagnosis is correct, though, I am not sure how best to handle this going forward. I have a feeling that we're going to need to revisit this anyway re: some of the ClobberWarnings that we discussed, it's not clear that we should be packaging up some pieces that we currently are.

@robertmaynard what do you think? Do we really want to be in a situation where cuspatial could _also_ install a copy of gtest if it didn't find one? That's what copying in `get_gtest.cmake` will do AFAICT.

bdice: Aha. This sounds like it should be solved by adding gtest packages and avoiding the CPM fetch entirely. Maybe adding the _logic_ for fetching gtest is fine, in case a user is missing those packages, but at CI build/test time I think we should be avoiding CPM fetches and use conda packages instead.

robertmaynard: No we should fix the underlying issue here instead of calling `get_gtest.cmake`. It would be good to back out the change and see what the full error message is, so we can find the root issue

conda/recipes/cuspatial/meta.yaml
```diff
@@ -1,9 +1,9 @@
-# Copyright (c) 2018, NVIDIA CORPORATION.
+# Copyright (c) 2018-2022, NVIDIA CORPORATION.
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
 {% set minor_version =  version.split('.')[0] + '.' + version.split('.')[1] %}
-{% set py_version=environ.get('CONDA_PY', 36) %}
-{% set cuda_version = '.'.join(environ.get('CUDA', '11.5').split('.')[:2]) %}
+{% set py_version=environ.get('CONDA_PY', 38) %}
```
ajschmidt8, line 5:
```suggestion
{% set py_version=environ['CONDA_PY'] %}
```

ajschmidt8 (approved):

cpp/CMakeLists.txt
bdice: In commit a629048, I fixed an issue with the specification of gtest in the conda recipe. The gtest/gmock conda packages are now available at build time and test run time, so CPM is not needed and this is no longer an issue for CI. As for source build correctness and whether CPM fetching should be possible outside the CI or conda context, I don’t have a concrete answer. I plan to revisit CPM/conda conflicts and ClobberWarnings from excessive CPM fetching in conda builds for all RAPIDS packages at some point in the future and may follow up with more questions. For now, we should not hold up the GitHub Actions migration on this packaging question.

trxcllnt: The C++ projects should configure correctly without conda. I'll clone this PR and see what's going on.

dependencies.yaml
```diff
@@ -100,17 +100,25 @@ dependencies:
       - output_types: conda
         matrices:
           - matrix:
-              cuda: "11.2"
+              cuda: "11.5"
             packages:
-              - cudatoolkit=11.2
+              - cudatoolkit=11.5
+              - libcusparse-dev>=11.7.0.31,<=11.7.0.107
+              - libcusparse>=11.7.0.31,<=11.7.0.107
```
trxcllnt, line 107: Do we need both `libcusparse` and `libcusparse-dev`? Shouldn't `libcusparse-dev` include the runtime lib from `libcusparse` (or at least depend on it)?

Also what's up w/ the version? Is this versioned independently from `cudatoolkit`?

cpp/CMakeLists.txt
trxcllnt: This configures fine for me outside conda:
```shell
# Configure RMM
cmake -S /rmm -B /rmm/build -GNinja -DBUILD_TESTS=ON -DBUILD_BENCHMARKS=ON
​
# Configure cuDF
cmake -S /cudf/cpp -B /cudf/cpp/build -GNinja -DBUILD_TESTS=ON -DBUILD_BENCHMARKS=ON -Drmm_ROOT=/rmm/build
​
# Compile the jitify headers so `/cudf/cpp/build/include` dir exists for cuSpatial
# (this is a bug in cuDF's cmake, we shouldn't have to compile anything for the CMake package to be a valid target)
ninja -C /cudf/cpp/build jitify_preprocess_run
​
# Configure cuSpatail
cmake -S /cuspatial/cpp -B /cuspatial/cpp/build -GNinja -DBUILD_TESTS=ON -DBUILD_BENCHMARKS=ON -Drmm_ROOT=/rmm/build -Dcudf_ROOT=/cudf/cpp/build
```

dependencies.yaml
bdice: Great questions! I struggled with both of these. I'll answer these in reverse order.

First, you are correct that the versions of individual CUDA Toolkit packages like `libcusparse[-dev]` are independent of the CUDA Toolkit version. In some GitHub Actions PRs, we have added comments about what CUDA Toolkit versions we're aiming for when pinning individual components -- I haven't checked that on this PR yet, we may need to add comments explaining that for cuSpatial if they aren't there yet.

Second, `libcusparse-dev` depends on `libcusparse`. We must pin both `libcusparse-dev` and `libcusparse` because the pinning is `>=`, meaning that pinning only `libcusparse-dev` to the versions that correspond to CUDA 11.5 means that the system could pull `libcusparse` runtime libraries for CUDA 12.0. Crossing CTK major versions is not a good thing, and we need to pin these in `dependencies.yaml` so that we are actually testing CTK components that we claim (e.g. CI tests for 11.2, 11.4, 11.5, ...)

For an example, see https://anaconda.org/nvidia/libcusparse-dev/files `linux-64/libcusparse-dev-11.7.0.107-h338262b_0.tar.bz2`.
image

trxcllnt (approved):

ajschmidt8 (approved):

bdice (commented):
I did a final pass of review and all looks good from my side.

ajschmidt8:
@gpucibot merge

PR #837, opened 2022-12-02
Enable copy_prs. [skip gpuci]

bdice:
## Description
Enables copying PRs so that GitHub Actions CI can run.

## Checklist

### DIFF
--- a/.github/ops-bot.yaml
+++ b/.github/ops-bot.yaml
@@ -6,3 +6,5 @@ branch_checker: true
 label_checker: true
 release_drafter: true
 external_contributors: false
+copy_prs: true
+rerun_tests: true
--- a/.github/ops-bot.yaml
+++ b/.github/ops-bot.yaml
@@ -7,4 +7,3 @@ label_checker: true
 release_drafter: true
 external_contributors: false
 copy_prs: true
-rerun_tests: true


.github/ops-bot.yaml
```diff
@@ -6,3 +6,5 @@ branch_checker: true
 label_checker: true
 release_drafter: true
 external_contributors: false
+copy_prs: true
+rerun_tests: true
```
ajschmidt8, line 10:
```suggestion
```

We should remove this. The current `rerun_tests` functionality provided by the ops-bot was intended to be used on Jenkins when we were refining our Jenkins process. It's not configured to work with GH Actions. If we keep this enabled, it will start sending invalid HTTP requests to our Jenkins server whenever someone types `rerun tests`.

Eventually I'll probably rewrite the `rerun_tests` functionality to work on GH Actions, but I think it's still in use by the `nv-morpheus` org, who was piloting our refined Jenkins build process. I believe they are in the process of migrating to GH Actions as well, but I haven't followed up with them in a while.

ajschmidt8 (approved):

.github/ops-bot.yaml
bdice: Thank you for the context and fix!

PR #850, opened 2022-12-09
Fix a bug preventing `align` from working with a basic slice.

### LINKED ISSUE #848: [BUG]: `align` fails on a pretty simple index alignment
### Version

22.12

Rapids-Compose

### Describe the issue

Using the demo code from the blog post, but dropping `align=False` from the final call, results in an error:

```shell
mixed_geoseries = cuspatial.GeoSeries([
     Polygon(
         [[-0.5, -0.5], [-0.5, 0.5], [0.5, 0.5], [0.5, -0.5], [-0.5, -0.5]]
     ),
     Polygon(
         [[-0.5, -0.5], [-0.5, 0.5], [0.5, 0.5], [0.5, -0.5], [-0.5, -0.5]]
     ),
     Polygon(
         [[-0.5, -0.5], [-0.5, 0.5], [0.5, 0.5], [0.5, -0.5], [-0.5, -0.5]]
     ),
     Point(-1, -1), Point(0, 0), Point(1, 1)
])
print(
    mixed_geoseries[0:3].contains_properly(mixed_geoseries[3:6])
)
```
```


### Relevant log output

```shell
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[23], line 32
     18 print(points.within(polygons))
     20 mixed_geoseries = cuspatial.GeoSeries([
     21     Polygon(
     22         [[-0.5, -0.5], [-0.5, 0.5], [0.5, 0.5], [0.5, -0.5], [-0.5, -0.5]]
   (...)
     30     Point(-1, -1), Point(0, 0), Point(1, 1)
     31 ])
---> 32 print(mixed_geoseries[0:3].contains_properly(mixed_geoseries[3:6]))

File ~/cuspatial/python/cuspatial/cuspatial/core/geoseries.py:778, in GeoSeries.contains_properly(self, other, align)
    701 def contains_properly(self, other, align=True):
    702     """Compute from a GeoSeries of points and a GeoSeries of polygons which
    703     points are properly contained within the corresponding polygon. Polygon
    704     A contains Point B properly if B intersects the interior of A but not
   (...)
    776         within the corresponding polygon in the input.
    777     """
--> 778     return ContainsProperlyBinop("contains_properly", self, other, align)()

File ~/cuspatial/python/cuspatial/cuspatial/core/binops/binops.py:143, in _binop.__init__(self, op, lhs, rhs, align)
     72 def __init__(self, op, lhs, rhs, align=True):
     73     """Compute the binary operation `op` on `lhs` and `rhs`.
     74 
     75     There are ten binary operations supported by cuspatial:
   (...)
    141         A GeoSeries containing the result of the binary operation.
    142     """
--> 143     (self.lhs, self.rhs) = lhs.align(rhs) if align else (lhs, rhs)
    144     self.align = align
    146     # Type disambiguation
    147     # Type disambiguation has a large effect on the decisions of the
    148     # algorithm.

File ~/cuspatial/python/cuspatial/cuspatial/core/geoseries.py:685, in GeoSeries.align(self, other)
    683 aligned_right.index = index
    684 aligned_right = aligned_right.sort_index()
--> 685 aligned_left.index = (
    686     self.index
    687     if len(self.index) == len(aligned_left)
    688     else aligned_right.index
    689 )
    690 aligned_left = aligned_left.sort_index()
    691 return (
    692     aligned_left,
    693     aligned_right.loc[aligned_left.index],
    694 )

File ~/cudf/python/cudf/cudf/core/indexed_frame.py:526, in IndexedFrame.index(self, value)
    524 # A DataFrame with 0 columns can have an index of arbitrary length.
    525 if len(self._data) > 0 and new_length != old_length:
--> 526     raise ValueError(
    527         f"Length mismatch: Expected axis has {old_length} elements, "
    528         f"new values have {len(value)} elements"
    529     )
    530 self._index = Index(value)

ValueError: Length mismatch: Expected axis has 6 elements, new values have 3 elements
```
```


### Environment details

_No response_

### Other/Misc.

_No response_

### PR DESCRIPTION
**Pull request by thomcom**
Closes #848 

## Description
This PR fixes a bug with `align` so that slicing a `GeoSeries`, then calling a binpred with slices of the `GeoSeries` with unmatching indexes works. It also adds a few more valid tests.

As of this PR one test still fails, I'm working on it but it feels like a rabbit hole. I'm inclined to leave the case for later:
The case that doesn't work is the hardest case wherein two `GeoSeries` have different length as well as indexes in one that is not present in the other.

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -668,19 +668,28 @@ class GeoSeries(cudf.Series):
         dtype: geometry)
 
         """
-        index = (
+
+        index_old = (
             other.index if len(other.index) >= len(self.index) else self.index
         )
+        idx1 = cudf.DataFrame({"idx": self.index})
+        idx2 = cudf.DataFrame({"idx": other.index})
+        index = idx1.merge(idx2, how="outer").sort_values("idx")["idx"]
+        index.name = None
         aligned_left = self._align_to_index(index)
         aligned_right = other._align_to_index(index)
-        aligned_right.index = index
+        breakpoint()
+        if len(other) == len(aligned_right):
+            aligned_right.index = other.index
+        else:
+            aligned_right.index = index
+        if len(self) == len(aligned_left):
+            aligned_left.index = self.index
+        else:
+            aligned_left.index = index
         aligned_right = aligned_right.sort_index()
-        aligned_left.index = (
-            self.index
-            if len(self.index) == len(aligned_left)
-            else aligned_right.index
-        )
         aligned_left = aligned_left.sort_index()
+        aligned_left.index = index
         return (
             aligned_left,
             aligned_right.loc[aligned_left.index],
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -670,6 +670,12 @@ class GeoSeries(cudf.Series):
         """
 
         # Merge the two indices and sort them
+        if (
+            len(self.index) == len(other.index)
+            and (self.index == other.index).all()
+        ):
+            return self, other
+
         idx1 = cudf.DataFrame({"idx": self.index})
         idx2 = cudf.DataFrame({"idx": other.index})
         index = idx1.merge(idx2, how="outer").sort_values("idx")["idx"]
--- a/python/cuspatial/cuspatial/core/geoseries.py
+++ b/python/cuspatial/cuspatial/core/geoseries.py
@@ -669,16 +669,18 @@ class GeoSeries(cudf.Series):
 
         """
 
-        index_old = (
-            other.index if len(other.index) >= len(self.index) else self.index
-        )
+        # Merge the two indices and sort them
         idx1 = cudf.DataFrame({"idx": self.index})
         idx2 = cudf.DataFrame({"idx": other.index})
         index = idx1.merge(idx2, how="outer").sort_values("idx")["idx"]
         index.name = None
+
+        # Align the two GeoSeries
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by harrism** [comment]
Does this fix #848? Should this be added to the current sprint and estimated?

**Comment by thomcom** [comment]
Yes, it is done for Sprint 2.

**Comment by harrism** [comment]
What about #849?

**Comment by thomcom** [comment]
#849 is fixed in https://github.com/rapidsai/cuspatial/pull/856

File: python/cuspatial/cuspatial/core/geoseries.py [hunk 1]
```diff
@@ -668,22 +668,34 @@ def align(self, other):
...
         """
-        index = (
-            other.index if len(other.index) >= len(self.index) else self.index
-        )
+
+        # Merge the two indices and sort them
+        idx1 = cudf.DataFrame({"idx": self.index})
+        idx2 = cudf.DataFrame({"idx": other.index})
+        index = idx1.merge(idx2, how="outer").sort_values("idx")["idx"]
+        index.name = None
```
isVoid, line 676: I noticed when `len(other) == len(aligned_right) and len(self) == len(aligned_left)` this computation (which is expensive), is actually not needed. Can we rearrange the logic a bit to only compute `index` when needed?

thomcom: Possibly, I'll try it out.

thomcom: Great suggestion, thanks.

thomcom: Done.

isVoid (approved):

PR #851, opened 2022-12-10
Adds `remove_if` Function to `linestring_intersection_intermediates`

isVoid:
## Description

https://github.com/rapidsai/cuspatial/pull/813 adds a struct to hold the intermediate result from `linestring_intersection_with_duplicates`. After discovering the duplicates from the result, we need a way to remove these geometries from the struct, as well as updating the offsets and look-back ids. This PR adds a `remove_if` function to the struct, which models after `thrust::remove_if` that accepts a range of flags, and removes the geometries in the intermediates where the flag is set to 1.

Closes #853 
Depends on #813. Contributes to #765.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #852, opened 2022-12-10
Add Header Only API `pairwise_linestring_intersection`

### LINKED ISSUE #765: header-only implementation of intersection
Implement header only API for `intersection_count` and `intersection`. 

```c++
template<typename Range1, typename Range2, typename OutputIt>
OutputIt pairwise_linestring_intersection(Range1 multilinestrings1, Range2 multilinestring2, OutputIt output_begin, stream);

template<typename Range1, typename Range2, typename OutputIt>
OutputIt pairwise_linestring_intersection_count(Range1 multilinestring1, Range2 multilinestring2, OutputIt output_begin, stream);
```

Depending on #763 and #764. With the help of these primitive and data structures, I posturizes that the kernel should be parallelized on 1-segment per thread and iterate on the multilinestring in the other multilinestring.

Kernel pseudo code without needing to write indices 

```c++
__global__ void intersection_count_kernel(range1, range2, output):
	tidx = get_threadidx()
	Segment seg1 = range1.segment(tidx)
	For (mls : range2)
		For (seg2 : mls)
			Point_opt, line_opt = segment_intersection(seg1, seg2)
			If (point_opt != nullopt)
				atomicInc(Output[tidx].num_intersecting_points)
			If (line_opt != nullopt)
				atomicInc(Output[tidx].num_overlapping_segments)
```

```c++
__global__ void intersection_kernel(range1, range2, temp_storage, output):
	tidx = get_threadidx()
	Segment seg1 = range1.segment(tidx)
	For (mls : range2)
		For (seg2 : mls)
			point_opt, line_opt = segment_intersection(seg1, seg2)
			If (point_opt != nullopt)
				Npoint = atomicReadAndInc(temp_storage.npoint)
				Output[npoint] = point_opt.value()
			If (line_opt != nullopt)
				NLine  = atomicReadAndInc(temp_storage.nline)
				Output[nLine] = line_opt.value()
```

[UPDATE: 11-8]
`count` API is only counting the upper bound of the number of intersection/overlaps - it does not give any promises to the number of results. So `count` will not be exposed as a public API.

https://github.com/rapidsai/cuspatial/issues/765#tasklist-block-52d7347e-d26a-4c76-968c-5e78b76839c8

isVoid:
## Description

This PR adds header only API `pairwise_linestring_intersection`. Closes #765 .

Depends on #813 #818 #819 #851

## Checklist


thomcom:
Waiting for the dependencies to be merged before reviewing this.

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * @brief Given a multipoint and a set of segments, for each point, if the point is
+ * on any of the segments, set the `mergeable_flag` of the point to `1`.
+ */
+template <typename MultiPointRange,
+          typename OffsetsRange,
+          typename SegmentsRange,
+          typename OutputIt1>
+void find_point_on_segments(MultiPointRange multipoints,
```
harrism, line 75: Should be `find_points_on_segments`? multipoints (plural), and a set of segments (plural)

cpp/include/cuspatial/experimental/linestring_intersection.cuh
```diff
@@ -0,0 +1,95 @@
...
+ * @tparam T Type of coordinate
+ * @tparam index_t Type of the look-back index in result
+ * @tparam MultiLinestringRange1 Multilinestring Range of the first operand
+ * @tparam MultiLinestringRange2 Multilinestring Range of the second operand
+ *
+ * @param multilinestrings1 Range to the first multilinestring in the pair
+ * @param multilinestrings2 Range to the second multilinestring in the pair
+ * @param mr The resource to use to allocate the returned data
+ * @param stream The CUDA stream on which to perform computations
+ * @return Results of intersection with duplicates
```
harrism, line 81: Does this result have duplicates? I thought this one had no duplicates.

harrism (approved):
This all came together really well. I can't see anything major wrong, just a couple of comments.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
```diff
@@ -0,0 +1,447 @@
...
+    CUSPATIAL_EXPECT_VECTORS_EQUIVALENT(*std::move(expected.rhs_segment_id),
+                                        *std::move(got.rhs_segment_id));
+  }
+};
+
+// float and double are logically the same but would require seperate tests due to precision.
+using TestTypes = ::testing::Types<float, double>;
+TYPED_TEST_CASE(LinestringIntersectionTest, TestTypes);
+
+// // TODO: sort the points in the intersection result since the result order is arbitrary.
```
thomcom, line 117: Is this for this PR? Seems hard to compare `got` and `expected` if results are arbitrary?

cpp/tests/experimental/spatial/linestring_intersection_test.cu
```diff
@@ -0,0 +1,447 @@
...
+  using T = TypeParam;
+  using P = vec_2d<T>;
+
+  using index_t = typename linestring_intersection_result<T, std::size_t>::index_t;
+  using types_t = typename linestring_intersection_result<T, std::size_t>::types_t;
+
+  auto multilinestrings1 = make_multilinestring_array({0, 1}, {0, 2}, {P{0.0, 0.0}, P{1.0, 1.0}});
+
+  auto multilinestrings2 =
+    make_multilinestring_array({0, 1}, {0, 4}, {P{0.0, 1.0}, P{1.0, 0.0}, P{0.5, 0.0}, P{0.5, 1}});
```
thomcom, line 171: :thumb: No duplicates!

cpp/tests/experimental/spatial/linestring_intersection_test.cu
```diff
@@ -0,0 +1,447 @@
...
+  auto expected =
+    make_linestring_intersection_result<T, index_t, types_t>({0, 1},
+                                                             {1},
+                                                             {0},
+                                                             {},
+                                                             {segment<T>{P{0.0, 0.0}, P{1.5, 1.5}}},
+                                                             {0},
+                                                             {0},
+                                                             {0},
+                                                             {0},
```
thomcom, line 204: The `ids` don't seem correct to me here, as they're all 0. `lhs` looks right, there's only one linestring and only one segment. On `rhs` however I'm confused at the output. There's only one linestring so 0 is correct for line #203, but line 204 is confusing. All three segments overlap, which you've combined into the result segment, but only segment 0 is returned.

If the user wanted for example to display each intersecting segment from `rhs` in different colors, this output would not enable that, I think? I'm ok with leaving the API as it is but this might need consideration.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
```diff
@@ -0,0 +1,447 @@
...
+  auto multilinestrings1 = make_multilinestring_array({0, 1}, {0, 2}, {P{0.0, 0.0}, P{3.0, 3.0}});
+
+  auto multilinestrings2 = make_multilinestring_array(
+    {0, 1}, {0, 4}, {P{0.0, 1.0}, P{1.0, 0.0}, P{2.0, 2.0}, P{0.0, 0.0}});
+
+  auto expected =
+    make_linestring_intersection_result<T, index_t, types_t>({0, 1},
+                                                             {1},
+                                                             {0},
+                                                             {},
```
thomcom, line 231: Does this not contain anything because the merged point exists in the input data, so the user can look it up?

cpp/tests/experimental/spatial/linestring_intersection_test.cu
```diff
@@ -0,0 +1,447 @@
...
+
+  auto multilinestrings1 = make_multilinestring_array({0, 1}, {0, 2}, {P{0.0, 0.0}, P{3.0, 3.0}});
+
+  auto multilinestrings2 = make_multilinestring_array(
+    {0, 1}, {0, 4}, {P{0.0, 1.0}, P{1.0, 0.0}, P{2.0, 2.0}, P{0.0, 0.0}});
+
+  auto expected =
+    make_linestring_intersection_result<T, index_t, types_t>({0, 1},
+                                                             {1},
+                                                             {0},
```
thomcom, line 230: A similar problem exists here where the user cannot recover which points in `rhs` intersect.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
thomcom: Actually shouldn't there be a return point, `P{0.5, 0.5}`  caused by the intersection when `({0.0, 1.0}, {1.0, 0.0})` crosses `({0.0, 0.0}, {3.0, 3.0})`? I see that got merged into `P{0.0, 2.0}`.

cpp/tests/experimental/spatial/linestring_intersection_intermediates_remove_if_test.cu
```diff
@@ -0,0 +1,185 @@
...
+    else
+      expect_segment_equivalent(*intermediates.geoms, *expected.geoms);
+    expect_vector_equivalent(*intermediates.lhs_linestring_ids, *expected.lhs_linestring_ids);
+    expect_vector_equivalent(*intermediates.lhs_segment_ids, *expected.lhs_segment_ids);
+    expect_vector_equivalent(*intermediates.rhs_linestring_ids, *expected.rhs_linestring_ids);
+    expect_vector_equivalent(*intermediates.rhs_segment_ids, *expected.rhs_segment_ids);
+  }
+};
+
+// float and double are logically the same but would require seperate tests due to precision.
```
thomcom, line 92:
```suggestion
// float and double are logically the same but would require separate tests due to precision.
```

cpp/tests/experimental/find/find_point_on_segments_test.cu
```diff
@@ -0,0 +1,207 @@
...
+  using P       = vec_2d<T>;
+  using S       = segment<T>;
+
+  auto multipoints = make_multipoints_array({{P{0.0, 0.0}, P{1.0, 0.0}}});
+
+  auto segment_offsets = make_device_vector<index_t>({0, 1});
+  auto segments        = make_device_vector<segment<T>>({S{P{1.0, 1.0}, P{1.0, -1.0}}});
+
+  rmm::device_vector<uint8_t> flags(multipoints.range().num_points());
+  std::vector<uint8_t> expected_flags{0, 1};
```
thomcom, line 53: All the `Simple` `FindPointOnSegmentTest`s test for flags `[0, 1]`. I'd like to see the basic set enumerated with these tests: `[0, 0]`, `[0, 1]`, `[1,0]`, `[1, 1]`. Ideally the same breakdown for `multipoints`

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -35,6 +35,21 @@ auto make_device_vector(std::initializer_list<T> inl)
   return rmm::device_vector<T>(inl.begin(), inl.end());
 }
 
+template <typename T>
+auto make_device_uvector(std::initializer_list<T> inl,
```
thomcom, line 39: I thought that `device_uvector` meant uninitialized. If so, `make_device_uvector` from an initialization list doesn't make sense to me as a name.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -123,6 +194,63 @@ struct linestring_intersection_intermediates {
                            thrust::next(offsets->begin()));
   }
 
+  /// Given a flag array, remove the ith geometry if `flags[i] == 1`.
+  /// @pre flag array must have the same size as the geometry array.
+  template <typename FlagRange>
+  void remove_if(FlagRange flags, rmm::cuda_stream_view stream)
```
thomcom, line 200: I'd love to see more detailed comments about the steps in this function.

thomcom (changes requested):
Looks really nice. Just a few comments about testing strategy and one about internal documentation. Most of my comments are just trying to understand the output of the final API. Do we have a concern about dropping duplicates and making it harder for the user to identify which points intersect or their count?

cpp/tests/experimental/spatial/linestring_intersection_intermediates_remove_if_test.cu
```diff
@@ -0,0 +1,185 @@
...
+  rmm::mr::device_memory_resource* mr() { return rmm::mr::get_current_device_resource(); }
+
+  template <typename FlagType, typename IntermediateType>
+  void run_single(IntermediateType& intermediates,
+                  std::initializer_list<FlagType> flags,
+                  IntermediateType const& expected)
+  {
+    using GeomType = typename IntermediateType::geometry_t;
+    auto d_flags   = make_device_uvector<FlagType>(flags, this->stream(), this->mr());
+    intermediates.remove_if(range(d_flags.begin(), d_flags.end()), this->stream());
```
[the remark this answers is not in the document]
isVoid: The name of the method is modeled after thrust: https://nvidia.github.io/thrust/api/groups/group__stream__compaction.html#function-remove-if with similar input.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: Yeah, this still needs to be done. Otherwise the test maybe flaky.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: Done in #851

cpp/include/cuspatial/experimental/linestring_intersection.cuh
isVoid: No, it's a typo.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: I think the original requirement for intersection is a ambiguous and this is certainly not considered in original design. Can file a separate issue to track if desired.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: Yes, that's the purpose of https://github.com/rapidsai/cuspatial/pull/819

isVoid:
> Do we have a concern about dropping duplicates and making it harder for the user to identify which points intersect or their count?

Unfortunately, the original requirement for the output of intersection is unclear about what the user really expects here. I went afar to return the most cleaned-up output that the algorithm can return. (So does geopandas) It could be an early-optimization. It could be likely that it's not what the user wants. If you think the design is not meeting the need, feel free to file an issue to track. 

My question is, does this intersection function help your binary predicate use case?

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: Hmm, Shapely seems to merge the points, but not the segments:
```python
In [1]: from shapely.geometry import *

In [2]: l = LineString([(0.0, 0.0), (3.0, 3.0)])

In [3]: l2 = LineString([(0.0, 1.0), (1.0, 0.0), (2.0, 2.0), (0.0, 0.0)])

In [4]: l.intersection(l2)
Out[4]: <MULTILINESTRING ((0 0, 0.5 0.5), (0.5 0.5, 2 2))>
```

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: To support your use case, you would need to group all merged geometry ids, this requires an additional level of offsets into the id arrays. I think that's a pretty big change and would take up a separate effort.

From a different angle, can't we pose this as a separate step from intersection? You have a intersection, finding where the intersection comes from is a search problem.

harrism: We may want to factor out the intersection metadata (segment indices) as a search problem in the future anyway.

cpp/include/cuspatial_test/vector_factories.cuh
harrism: I think it's fine. a device_uvector is a container that is never *default* initialized. But you can still put data in it. This factory function just creates a uvector and then copies data into it.  

Note this is only available in tests, it is not intended for use in the library or by users of the library.  I kind of agree that we could probably get away with just `make_device_vector` in tests...

isVoid: I think the main reason why I added this helper is because `intersection_result` consists of `device_uvector`. And I tried to keep things simple by constructing `device_uvector` in test and just copy/move it to the struct. Using `device_vector` will require some extra steps.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
thomcom: I'm in favor of creating a new issue to track this, it doesn't need to block this PR.

thomcom:
I'm sure this helps with implementing a great deal of binary predicates! I'm not exactly sure how, yet. I'm a little worried about a case of $\forall p in \lambda_1 \cap \lambda_2$ if it is required. This definitely covers `None` and `any` cases, which, without looking closely at dozens of intersection predicate calls yet, probably covers most of them.

thomcom:
My questions are mostly general and for the purpose of understanding what's coming my way soon. :) I see @harrism already approved, I will approve if you want to get this merged before the week ends.

cpp/include/cuspatial_test/vector_factories.cuh
thomcom: I don't require a change, but was mostly referring to the fact that the function name doesn't compute well for me. Since `uvector` takes an initializer list and that users are going to understand that it takes an initializer list, I'm ok with the name.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
thomcom: This result is really interesting to me. It seems that shapely combines the point intersection with the segment intersection in this example. It also seems like that's something that could be accomplished in python, though perhaps not as efficiently.

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: Ok, I think sorting the result is finally added!

cpp/tests/experimental/spatial/linestring_intersection_test.cu
```diff
@@ -30,12 +34,126 @@
...
+ * [
+ *  [Point(0.0, -0.3), Point(1.0, 1.5), Segment((0.0, 0.0), (1.0, 1.0))]
+ *   ^                 ^
+ *  [Point(-3, -5)]
+ * ]
+ *
+ * This function invalidates the input @p result and return a copy of sorted results.
+ */
+template <typename T, typename IndexType, typename type_t>
+linestring_intersection_result<T, IndexType> segment_sort_intersection_result(
```
thomcom, line 70: I can see us not wanting this to be only a test utility down the road.

cpp/include/cuspatial_test/vector_factories.cuh
isVoid: I believe in c++ `make_*` is a convention for factory functions of classes. In standard c++, they were really meant to be a hack to make classes before CTAD. In cuDF, there are examples of building columns from various input specs that's non-standard to `cudf::column` ctor. I'm kinda abusing that idea here so it makes sense that it feels off. I'll add a line of comment to remind for better naming here.

thomcom (changes requested):

cpp/tests/experimental/find/find_point_on_segments_test.cu
isVoid: Done

cpp/include/cuspatial/experimental/detail/find/find_point_on_segments.cuh
isVoid: renamed

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: #858

cpp/tests/experimental/spatial/linestring_intersection_test.cu
isVoid: #860

thomcom (approved):

isVoid:
@gpucibot merge

PR #857, opened 2022-12-14
Fix: Sort Intersection Result by Keys in `intersection_with_duplicate` Test

isVoid:
## Description

This PR sorts the intersection result in `linestring_intersection_with_duplicate` test. If a pair has more than 1 intersection result (e.g. 2 intersecting points) the order of these geometries is non-deterministic. While this doesn't affect the semantic of the result, this makes tests flaky and may break CI. Thus in the tests we have to sort the result before testing. This PR adds the infrastructure to sort the results, as well as rearranging the order to some of the hard coded results.

## Checklist

### DIFF
--- a/cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
+++ b/cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
@@ -55,7 +55,7 @@ bool CUSPATIAL_HOST_DEVICE operator<(segment<T> lhs, segment<T> rhs)
  *
  * Using a label array and a geometry array as keys, this functor defines that
  * all keys with smaller labels should precede keys with larger labels; and that
- * the order with the same label should be determined by the natrual order of the
+ * the order with the same label should be determined by the natural order of the
  * geometries.
  *
  * Example:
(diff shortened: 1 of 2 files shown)


cpp/tests/experimental/spatial/linestring_intersection_with_duplicates_test.cu
```diff
@@ -24,18 +24,136 @@
...
+{
+  return lhs.first < rhs.first || (lhs.first == rhs.first && lhs.second < rhs.second);
+}
+
+/**
+ * @brief Functor for segmented sorting a geometry array
+ *
+ * Using a label array and a geometry array as keys, this functor defines that
+ * all keys with smaller labels should precede keys with larger labels; and that
+ * the order with the same label should be determined by the natrual order of the
```
harrism, line 58:
```suggestion
 * the order with the same label should be determined by the natural order of the
```

harrism (approved):
Impressed by how fast you whipped this up. 

However, would it have been possible (and simpler) to use some thrust set operation (e.g. thrust::symmetric_distance) to do an unordered comparison between the expected results and the actual (non-deterministic) results?  Could also be the `by_key` version as needed.

https://thrust.github.io/doc/group__set__operations.html

isVoid:
@harrism Didn't think about that. Good suggestion, although we would have to rewrite the test utilities to make use of thrust set operations. And producing the same error reports - difference report and error logs, would require more infrastructure.

On another note, I wonder if gmock has set-like expects..

isVoid:
`set_difference_by_key` seems to require the range to be sorted first...

harrism:
I don't think you need new utilities. In the test you would just do a set_symmetric_difference and expect the result to be empty. :)

isVoid:
Sounds good, I'm trying this approach in https://github.com/rapidsai/cuspatial/pull/852.

Issue #565, opened 2022-06-15
[FEA] Refactor Spatial Join APIs to header-only


harrism:
Create header-only APIs for the following and refactor existing APIs / tests on top of the header-only API:

See https://github.com/rapidsai/cuspatial/blob/branch-22.08/cpp/include/cuspatial/spatial_join.hpp

`cuspatial::join_quadtree_and_bounding_boxes()` completed by #861 

```[tasklist]

- [x] Header-only join_quadtree_and_bounding_boxes
- [ ] #985
- [ ] #986
- [ ] https://github.com/rapidsai/cuspatial/issues/1006
- [ ] https://github.com/rapidsai/cuspatial/issues/1007
```

jarmak-nv:
Completed with #861

trxcllnt:
@jarmak-nv there's two more methods this issue describes, so we've been keeping it open until they're all done.

jarmak-nv:
Oh apologies! I just saw status done. I'll reach out to confirm in the future.

PR #861, opened 2022-12-20
Add header-only `cuspatial::join_quadtree_and_bounding_boxes`

trxcllnt:
Contributes to #565. Add header-only `cuspatial::join_quadtree_and_bounding_boxes`.


cpp/include/cuspatial/experimental/detail/quadtree_bbox_filtering.cuh
```diff
@@ -0,0 +1,199 @@
...
+  rmm::device_uvector<uint8_t> cur_types(num_pairs, stream);
+  rmm::device_uvector<uint8_t> cur_levels(num_pairs, stream);
+  rmm::device_uvector<uint32_t> cur_node_idxs(num_pairs, stream);
+  rmm::device_uvector<uint32_t> cur_bbox_idxs(num_pairs, stream);
+
+  // Vectors for found pairs of bbox and leaf node indices
+  rmm::device_uvector<uint32_t> out_node_idxs(num_pairs, stream, mr);
+  rmm::device_uvector<uint32_t> out_bbox_idxs(num_pairs, stream, mr);
+
+  auto make_current_level_iter = [&]() {
```
isVoid, line 85: Any reason to use a lambda here? It seems we should create the iterator as a variable instead.

cpp/include/cuspatial/experimental/detail/join/intersection.cuh
```diff
@@ -0,0 +1,146 @@
...
+      T const node_x_max  = x_min + (key_x + 1) * level_scale;
+      T const node_y_max  = y_min + (key_y + 1) * level_scale;
+
+      if ((node_x_min > poly_x_max) || (node_x_max < poly_x_min) || (node_y_min > poly_y_max) ||
+          (node_y_max < poly_y_min)) {
+        // if no overlap, return type = none_indicator
+        return thrust::make_tuple(none_indicator, level, node_idx, bbox_idx);
+      }
+      // otherwise return type = leaf_indicator (0) or quad_indicator (1)
+      return thrust::make_tuple(is_internal_node, level, node_idx, bbox_idx);
```
isVoid, line 134: It would be nice to have a `box.overlap()` helper that we can just call 
```c++
return bbox.overlaps(node_bbox) ? tuple(is_internal_node, ..) : tuple(none_indicator, ..);
```
This is a really high level thought, so no need to implement.

cpp/include/cuspatial/experimental/detail/join/intersection.cuh
```diff
@@ -0,0 +1,146 @@
...
+
+namespace cuspatial {
+namespace detail {
+
+static __device__ uint8_t const leaf_indicator = 0;
+static __device__ uint8_t const quad_indicator = 1;
+static __device__ uint8_t const none_indicator = 2;
+
+template <typename InputIterator, typename OutputIterator>
+inline int32_t copy_leaf_intersections(InputIterator input_begin,
```
isVoid, line 44: Documentation for this function.

cpp/include/cuspatial/experimental/detail/join/intersection.cuh
```diff
@@ -0,0 +1,146 @@
...
+  return thrust::distance(
+    output_begin,
+    thrust::copy_if(
+      rmm::exec_policy(stream), input_begin, input_end, output_begin, [] __device__(auto const& t) {
+        return thrust::get<0>(t) == leaf_indicator;
+      }));
+}
+
+template <typename InputIterator, typename OutputIterator>
+inline int32_t remove_non_quad_intersections(InputIterator input_begin,
```
isVoid, line 58: Also documentation.

cpp/include/cuspatial/experimental/detail/quadtree_bbox_filtering.cuh
trxcllnt: Yes, this lambda closes over variables that are updated each iteration of the loop.

harrism (changes requested):
Looks good, thanks Paul! Please update all copyrights in new / modified files to 2023.

Couple of minor comments.  There are some places with really long lambdas that you could consider turning into functor classes for tidyness, but I'm not going to press the issue.

cpp/include/cuspatial/experimental/detail/quadtree_bbox_filtering.cuh
isVoid: Do you mean that the iterators gets invalidated after every iteration of the loop?

isVoid (approved):
Thanks!

cpp/include/cuspatial/experimental/detail/quadtree_bbox_filtering.cuh
trxcllnt: `make_current_level_iter()` returns an zip iterator over the `cur_{types,levels,node_idxs,bbox_idxs}` vectors, which represent the current level in the tree descent. The `descend_quadtree()` call [here](https://github.com/rapidsai/cuspatial/blob/adb747a4b4bb2d677b3d908cf3175879816bb27e/cpp/include/cuspatial/experimental/detail/quadtree_bbox_filtering.cuh#L143-L151) descends one level in the tree and returns vectors representing the next "current level." 

Similarly, `make_output_values_iter()` returns a zip iterator of `num_results + out_{node,bbox}_idxs`. `num_results` is updated by the `detail::find_intersections()` call [here](https://github.com/rapidsai/cuspatial/blob/adb747a4b4bb2d677b3d908cf3175879816bb27e/cpp/include/cuspatial/experimental/detail/quadtree_bbox_filtering.cuh#L154-L172).

The iterators returned by these functions are used both in setting up the initial state of the the top-level quadrants, and in the loop that does the tree traversal. An earlier implementation tracked and updated these two iterators each iteration of the loop, but [they were refactored](https://github.com/rapidsai/cuspatial/commit/d618baf989cd455ae65d90ddf102f455ca4eb90c#diff-c3817f6937f1738ced51e1fa383142ce02976d629b91fe5f19f4203eb63dc5b7) into these lambdas in a cleanup PR.

harrism (approved):

Issue #261, opened 2020-08-06
[FEA] Use cudf List types for polygon and polyline geometry


trxcllnt:
**Is your feature request related to a problem? Please describe.**
With the addition of  List and Struct types in cuDF, we can update cuSpatial's polygon and polyline geometry representations to use nested "list of lists of point structures".

**Describe the solution you'd like**
The logical structure of a list of polygons looks like this:
```js
polygons = [
    [ // <-- polygon 1
        [ // <-- ring 1
            [40, 40], [20, 45], [45, 30], [40, 40] // <-- points
        ]
    ],
    [ // <-- polygon 2
        [ // <-- ring 1
            [20, 35], [10, 30], [10, 10], [30, 5], [45, 20], [20, 35] // <-- points
        ],
        [ // <-- ring 2
            [30, 20], [20, 15], [20, 25], [30, 20] // <-- points
        ]
    ]
]
```

We can represent these data in the same 4 buffers we use today (`poly_offsets`, `ring_offsets`, `points_x`, `points_y`) with the following Arrow types:

```python
import pyarrow as pa

polygons = pa.list(pa.field(
  "rings", pa.list(pa.field(
    "points", pa.struct([
      pa.field("x", pa.float32()),
      pa.field("y", pa.float32())
    ])
  ))
))

# The above example's underlying buffers:
# polygon_offsets: [0, 1, 3]
# ring_offsets: [0, 4, 10, 14]
# xs: [40, 20, 45, 40, 20, 10, 10, 30, 45, 20, 30, 20, 20, 30]
# ys: [40, 45, 30, 40, 35, 30, 10, 5, 20, 35, 20, 15, 25, 20]
```

**Additional context**
https://github.com/geopandas/geo-arrow-spec/issues/3
https://github.com/geopandas/geo-arrow-spec/issues/4

trxcllnt:
This should be possible now that we have list and struct types in cuDF.

PR #862, opened 2022-12-21
Add C++ Column API and Python API for `pairwise_linestring_intersection`

### LINKED ISSUE #648: [FEA] linestring-linestring intersection (pairwise)

A function to return, for each specified pair of linestrings, the point at which they intersect. We may also need the segment index within each linestring of the intersection.

https://github.com/rapidsai/cuspatial/issues/648#tasklist-block-d3e21ca2-3c8a-4251-990f-97fe808b41bf

isVoid:
## Description

This PR adds column/python API for `pairwise_linestring_intersection`. Closes #648 

This PR is also the first attempt of #261 , establishing `geometry_column_view` that inherits `cudf::column` and adds additional information to identify the underlying nested levels. Note that it cannot directly inherit from `cudf::lists::list_column_view`, because a point array isn't a `List<T>` array.

Also note that the type codes defined in libcuspatial and cuspatial-python may be different, so there is a type code mapping after computing the result from libcuspaital.

Finally, since `List<Union>` is currently unsupported in libcudf, the python API returns the geometries in two parts: `offset + GeoColumn`.

## Checklist


cpp/include/cuspatial/column/geometry_column_view.hpp
```diff
@@ -0,0 +1,75 @@
...
+                       geometry_type_id geometry_type);
+  geometry_column_view(geometry_column_view&&)      = default;  ///< Move constructor
+  geometry_column_view(const geometry_column_view&) = default;  ///< Copy constructor
+  ~geometry_column_view()                           = default;
+
+  /**
+   * @brief Copy assignment operator
+   *
+   * @return The reference to this lists column
+   */
```
harrism, line 50: The docs for the two `operator=` seem unnecessary. They are obvious.

cpp/include/cuspatial/experimental/ranges/multilinestring_range.cuh
```diff
@@ -132,7 +133,8 @@ class multilinestring_range {
   template <typename IndexType>
   CUSPATIAL_HOST_DEVICE auto operator[](IndexType multilinestring_idx);
 
- protected:
+  //  protected:
+ public:
```
harrism, line 137: Is this meant to be temporary?

cpp/include/cuspatial/traits.hpp
```diff
@@ -26,6 +26,19 @@
 
 namespace cuspatial {
 
+/**
+ * @brief Convenience macro for SFINAE as an unnamed template parameter.
+ *
+ * Example:
+ * \code{cpp}
+ * // This function will participate in overload resolution only if T is an integral type
+ * template <typename T, CUDF_ENABLE_IF(std::is_integral_v<T> )>
```
harrism, line 35:
```suggestion
 * template <typename T, CUSPATIAL_ENABLE_IF(std::is_integral_v<T> )>
```

cpp/src/column/geometry_column_view.cu
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
```
harrism, line 2:
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.
```

python/cuspatial/cuspatial/_lib/types.pyx
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
```
harrism, line 1:
```suggestion
# Copyright (c) 2023, NVIDIA CORPORATION.
```

python/cuspatial/cuspatial/core/binops/intersection.py
```diff
@@ -0,0 +1,123 @@
...
+def pairwise_linestring_intersection(
+    linestrings1: GeoSeries, linestrings2: GeoSeries
+):
+    """
+    Compute the intersection of two GeoSeries of linestrings.
+
+    Note
+    ----
+    Note that the result should be interpreted as type list<union>, since
+    cuDF does not support union type, the result is returned as a sperate list
```
harrism, line 23:
```suggestion
    cuDF does not support union type, the result is returned as a separate list
```

python/cuspatial/cuspatial/core/binops/intersection.py
```diff
@@ -0,0 +1,123 @@
...
+    linestrings1 : GeoSeries
+        A GeoSeries of linestrings.
+    linestrings2 : GeoSeries
+        A GeoSeries of linestrings.
+
+    Returns
+    -------
+    Tuple[cudf.Series, GeoSeries, DataFrame]
+        A tuple of three elements:
+        - An integral cudf serial to indicate the offsets of the geoseries.
```
harrism, line 37:
```suggestion
        - An integral cuDF series of offsets to each intersection result in the GeoSeries.
```

python/cuspatial/cuspatial/core/binops/intersection.py
```diff
@@ -0,0 +1,123 @@
...
+                    "lhs_segment_id": [],
+                    "rhs_linestring_id": [],
+                    "rhs_segment_id": [],
+                }
+            ),
+        )
+
+    if not contains_only_linestrings(
+        linestrings1
+    ) or not contains_only_linestrings(linestrings2):
```
harrism, line 59: This is very weird formatting. Can it not be:

```suggestion
    if not contains_only_linestrings(linestrings1) or 
       not contains_only_linestrings(linestrings2):
```

harrism (changes requested):
Looks good. This probably could have been two PRs (C++ and Python). Please update the copyright in all added/changed files to 2023.

python/cuspatial/cuspatial/core/binops/intersection.py
thomcom: Probably just what the `black` autoformatter did with it.

thomcom (approved):

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

python/cuspatial/cuspatial/core/binops/intersection.py
isVoid: Yes, this is due to `black`. And the reason why black does this is because of [PEP-8](https://peps.python.org/pep-0008/#maximum-line-length):

> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

harrism (changes requested):
Looks great. A few small comments.

cpp/include/cuspatial/experimental/ranges/multilinestring_range.cuh
```diff
@@ -226,6 +227,67 @@ auto make_multilinestring_range(IntegerRange1 geometry_offsets,
                                points.end());
 }
 
+/**
+ * @ingroup ranges
+ * @brief Create a range object of multilinestring from cuspatial::geometry_column_view
+ * Specialization for linestrings column.
+ *
+ * @pre linestrings_column must be a cuspatial::geometry_column_view
```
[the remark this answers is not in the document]
isVoid: If we don't, then the header only API would need to depend on cuDF (as a transitive dependency in geoemtry_column_view.hpp).

cpp/src/spatial/linestring_intersection.cu
```diff
@@ -0,0 +1,187 @@
...
+namespace detail {
+
+/**
+ * @brief Move the ownership to a buffer from a `device_uvector` to a `cudf::column`
+ */
+template <typename T>
+std::unique_ptr<cudf::column> move_uvector(std::unique_ptr<rmm::device_uvector<T>> ptr)
+{
+  return std::unique_ptr<cudf::column>(new cudf::column(std::move(*ptr)));
+}
```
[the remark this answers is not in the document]
isVoid: I think I ran into some issue with the suggested. Let me do some testing.

python/cuspatial/cuspatial/core/binops/intersection.py
isVoid: Found a better way to express this.

cpp/src/spatial/linestring_intersection.cu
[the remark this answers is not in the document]
isVoid: Can't reproduce issue, helper is removed.

harrism (approved):

isVoid:
Although CI is passing, @thomcom finds a memory error when running all the tests locally which I can also repro. Holding until I figure out what's going on.

thomcom (changes requested):
Just following up that I get a critical `bad_access` error when I run this will the full suite of tests. It is clearly not happening with CI but it happens on @isVoid and my local machines so he's looking into it.

thomcom (approved):

PR #910, opened 2023-02-08
Use quadtree for `.contains_properly`

### LINKED ISSUE #889: [BUG]: `.contains_properly` fails with single MultiPolygons of length greater than 31
### Version

23.04

Rapids-Compose

### Describe the issue

Even knowing about the 31 polygon limit, a user might expect a single `MultiPolygon` to work with `.contains_properly`. It does not. `.contains_properly` should be updated to use quadtree point in polygon to defeat this limit and improve usability, even if the quadtree implementation also has limitations.

This is primarily to support the GTC demo, which should have a clean and pleasant user experience with `.contains_properly`. We will still need to work on our overall design of point in polygon for use with large datasets.

```shell
import cuspatial
from shapely.geometry import MultiPolygon, Polygon, Point
def get_polygon():
    return Polygon([[0, 0], [1, 0], [1, 1]])
series = cuspatial.GeoSeries([MultiPolygon([get_polygon() for i in range(32)])])
series.contains_properly(cuspatial.GeoSeries([Point(0, 0)]))
```

### Relevant log output

```shell
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[16], line 7
      5     return Polygon([[0, 0], [1, 0], [1, 1]])
      6 series = cuspatial.GeoSeries([MultiPolygon([get_polygon() for i in range(32)])])
----> 7 series.contains_properly(cuspatial.GeoSeries([Point(0, 0)]))

File ~/cuspatial/python/cuspatial/cuspatial/core/geoseries.py:874, in GeoSeries.contains_properly(self, other, align)
    794 def contains_properly(self, other, align=True):
    795     """Returns a `Series` of `dtype('bool')` with value `True` for each
    796     aligned geometry that contains _other_.
    797 
   (...)
    872         within the corresponding polygon in the input.
    873     """
--> 874     return ContainsProperlyBinpred(self, other, align)()

File ~/cuspatial/python/cuspatial/cuspatial/core/binpreds/binpreds.py:125, in BinaryPredicate.__call__(self)
    122 (lhs, rhs, indices) = self.preprocess(self.lhs, self.rhs)
    124 # Binpred call
--> 125 point_result = self._op(lhs, rhs)
    127 # Postprocess: Apply discrete math rules to identify relationships.
    128 return self.postprocess(indices, point_result)

File ~/cuspatial/python/cuspatial/cuspatial/core/binpreds/binpreds.py:173, in ContainsProperlyBinpred._op(self, lhs, points)
    168     raise TypeError(
    169         "`.contains` can only be called with polygon series."
    170     )
    172 # call pip on the three subtypes on the right:
--> 173 point_result = contains_properly(
    174     points.x,
    175     points.y,
    176     lhs.polygons.part_offset[:-1],
    177     lhs.polygons.ring_offset[:-1],
    178     lhs.polygons.x,
    179     lhs.polygons.y,
    180 )
    181 return point_result

File ~/cuspatial/python/cuspatial/cuspatial/core/binpreds/contains.py:79, in contains_properly(test_points_x, test_points_y, poly_offsets, poly_ring_offsets, poly_points_x, poly_points_y)
     70     pip_result = cpp_pairwise_point_in_polygon(
     71         test_points_x,
     72         test_points_y,
   (...)
     76         poly_points_y,
     77     )
     78 else:
---> 79     pip_result = cpp_point_in_polygon(
     80         test_points_x,
     81         test_points_y,
     82         poly_offsets_column,
     83         poly_ring_offsets_column,
     84         poly_points_x,
     85         poly_points_y,
     86     )
     88 result = Series(pip_result, dtype="bool")
     89 return result

File point_in_polygon.pyx:32, in cuspatial._lib.point_in_polygon.point_in_polygon()

RuntimeError: cuSpatial failure at: /home/tcomer/mnt/NVIDIA/rapids-docker/cuspatial/cpp/src/spatial/point_in_polygon.cu:140: Number of polygons cannot exceed 31
```

### Environment details

```shell
rapids-compose 23.04
```

### Other/Misc.

_No response_

### LINKED ISSUE #896: [BUG]: Unexpected failure in `.contains_properly`.
### Version

23.04

Rapids-Compose

### Describe the issue

Using the `naturalearth_lowres` dataset produces an unexpected error with contains properly. We expect that a `GeoSeries` can be used as its own argument to `.contains_properly`, as in `x.contains_properly(x)` and will return False. With the `naturalearth_lowres` dataframe, an error is thrown.

```shell
import geopandas as gpd
import cuspatial
from shapely.geometry import Polygon
tpoly = cuspatial.GeoSeries([Polygon([[0, 1], [2, 3], [4, 5]])])
tpoly.contains_properly(tpoly)

df = cuspatial.GeoDataFrame(
    gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
)
df['geometry'].contains_properly(df['geometry'])
```

### Relevant log output

```shell
0    False
dtype: bool

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[2], line 7
      3 import cuspatial
      4 df = cuspatial.GeoDataFrame(
      5     gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
      6 )
----> 7 df['geometry'].contains_properly(df['geometry'])

File ~/cuspatial/python/cuspatial/cuspatial/core/geoseries.py:874, in GeoSeries.contains_properly(self, other, align)
    794 def contains_properly(self, other, align=True):
    795     """Returns a `Series` of `dtype('bool')` with value `True` for each
    796     aligned geometry that contains _other_.
    797 
   (...)
    872         within the corresponding polygon in the input.
    873     """
--> 874     return ContainsProperlyBinpred(self, other, align)()

File ~/cuspatial/python/cuspatial/cuspatial/core/binpreds/binpreds.py:122, in BinaryPredicate.__call__(self)
    118 """Return the result of the binary predicate."""
    119 # Type disambiguation
    120 # Type disambiguation has a large effect on the decisions of the
    121 # algorithm.
--> 122 (lhs, rhs, indices) = self.preprocess(self.lhs, self.rhs)
    124 # Binpred call
    125 point_result = self._op(lhs, rhs)

File ~/cuspatial/python/cuspatial/cuspatial/core/binpreds/binpreds.py:154, in ContainsProperlyBinpred.preprocess(self, lhs, rhs)
    150 xy_points = geom.xy
    152 # Arrange into shape for calling point-in-polygon, intersection, or
    153 # equals
--> 154 point_indices = geom.point_indices()
    155 from cuspatial.core.geoseries import GeoSeries
    157 final_rhs = GeoSeries(
    158     GeoColumn._from_points_xy(xy_points._column)
    159 ).points

File ~/cuspatial/python/cuspatial/cuspatial/core/geoseries.py:242, in GeoSeries.PolygonGeoColumnAccessor.point_indices(self)
    240 offsets = cp.array(self.ring_offset)
    241 sizes = offsets[1:] - offsets[:-1]
--> 242 return cp.repeat(self._series.index, sizes)

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/site-packages/cupy/_manipulation/tiling.py:70, in repeat(a, repeats, axis)
     56 def repeat(a, repeats, axis=None):
     57     """Repeat arrays along an axis.
     58 
     59     Args:
   (...)
     68 
     69     """
---> 70     return a.repeat(repeats, axis)

File ~/cudf/python/cudf/cudf/core/index.py:809, in RangeIndex.repeat(self, repeats, axis)
    808 def repeat(self, repeats, axis=None):
--> 809     return self._as_int_index().repeat(repeats, axis)

File ~/cudf/python/cudf/cudf/core/index.py:1492, in GenericIndex.repeat(self, repeats, axis)
   1490 def repeat(self, repeats, axis=None):
   1491     return self._from_columns_like_self(
-> 1492         Frame._repeat([*self._columns], repeats, axis), self._column_names
   1493     )

File ~/cudf/python/cudf/cudf/core/frame.py:2810, in Frame._repeat(columns, repeats, axis)
   2807 if not is_scalar(repeats):
   2808     repeats = as_column(repeats)
-> 2810 return libcudf.filling.repeat(columns, repeats)

File ~/compose/etc/conda/cuda_11.6/envs/notebooks/lib/python3.8/contextlib.py:75, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     72 @wraps(func)
     73 def inner(*args, **kwds):
     74     with self._recreate_cm():
---> 75         return func(*args, **kwds)

File filling.pyx:57, in cudf._lib.filling.repeat()

File filling.pyx:68, in cudf._lib.filling._repeat_via_column()

RuntimeError: CUDF failure at: /home/tcomer/mnt/NVIDIA/rapids-docker/cudf/cpp/src/filling/repeat.cu:109: in and count must have equal size
```

### Environment details

```shell
rapids-compose 23.04
```

### Other/Misc.

_No response_

thomcom:
## Description
Closes #889
Closes #896 

This PR makes a number of changes to the binpred system, including adding boundary exclusion for `.contains_properly` to the quadtree line-crossings algorithm.

I've closed and then reopened this PR due to benchmark results that showed that using `pairwise` and `column-limited` forms of point-in-polygon don't offer any performance improvement but would just require more complicated code paths.

## Checklist


python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -183,25 +186,64 @@ def _op(self, lhs, points):
...
+            geometry_map = cudf.Series(
+                cp.arange(len(geometry_sizes)), name="polygon_index"
+            ).repeat(geometry_sizes)
+            geom_df = geometry_map.reset_index(drop=True)
+            geom_df.index.name = "part_index"
+            geom_df = geom_df.reset_index()
+            # Replace the part index with the polygon index
+            part_result = parts_df.merge(point_result, on="part_index")
+            # Replace the polygon index with the row index
+            result = geom_df.merge(part_result, on="part_index")
```
thomcom, line 213: Maybe you can reduce this to a couple of `.take`s like you did in the `point_indices` pr, @isVoid?

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -183,25 +186,64 @@ def _op(self, lhs, points):
...
+                    self.lhs.index, index=cp.arange(len(self.lhs.index))
+                )
             )
-        point_result = cudf.Series(
-            df_result["pip"], index=cudf.RangeIndex(0, len(df_result))
-        )
-        point_result.name = None
-        return point_result
+            return result
+        else:
```
thomcom, line 223:
```suggestion
        else:
        # pairwise
```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -183,25 +186,65 @@ def _op(self, lhs, points):
...
+        # pairwise
```
thomcom, line 224:
```suggestion
            # pairwise
```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -183,25 +186,65 @@ def _op(self, lhs, points):
...
+            parts_map = cudf.Series(
+                cp.arange(len(part_sizes)), name="part_index"
+            ).repeat(part_sizes)
+            # Mapping of parts to polygons
+            parts_df = parts_map.reset_index(drop=True).reset_index()
+            parts_to_geoms = cp.array(self.lhs.polygons.geometry_offset)
+            geometry_sizes = parts_to_geoms[1:] - parts_to_geoms[:-1]
+            geometry_map = cudf.Series(
+                cp.arange(len(geometry_sizes)), name="polygon_index"
+            ).repeat(geometry_sizes)
```
isVoid, line 206: Simplify
```suggestion
            pcol = self.lhs.polygons
            offsets = pcol.ring_offset.take(pcol.part_offset).take(self.geometry_offset)
            sizes = offsets[1:] - offsets[:-1]
            geometry_map = self.Series(cp.arange(len(pcol.xy) // 2).repeat(sizes)
```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -183,25 +186,65 @@ def _op(self, lhs, points):
...
+            geom_df = geometry_map.reset_index(drop=True)
+            geom_df.index.name = "part_index"
+            geom_df = geom_df.reset_index()
```
isVoid, line 209: Why do you have to reset index twice?

python/cuspatial/cuspatial/tests/binpreds/test_contains_properly.py
```diff
@@ -349,3 +350,210 @@ def test_max_polygons_max_multipoints(multipoint_generator, polygon_generator):
...
+        ),
+    ],
+)
+def test_self_contains(object):
+    gpdobject = gpd.GeoSeries(object)
+    object = cuspatial.from_geopandas(gpdobject)
+    got = object.contains_properly(object).values_host
+    expected = gpdobject.contains(gpdobject).values
+    assert not got
+    assert expected
```
isVoid, line 373: We must use `np.testing.assert_arrays_equal` to test here (or other explicit tests). Because Python will recognize all non Falsy object as True. And that means `assert expected` will always be true. And `assert not got` might still be true even if you return `None` for `contains_properly`.

```suggestion
    np.testing.assert_arrays_equal(got, np.array([False]))
    np.testing.assert_arrays_equal(expected, np.array([True]))
```

python/cuspatial/cuspatial/tests/binpreds/test_contains_properly.py
```diff
@@ -349,3 +350,210 @@ def test_max_polygons_max_multipoints(multipoint_generator, polygon_generator):
...
+            Point(0.5, 0.25),
+            Point(0.5, 0.25),
+            Point(0.5, 0.25),
+            Point(0.5, 0.25),
+            Point(0.5, 0.25),
+            Point(0.5, 0.25),
+        ]
+    )
+    got = lhs.contains_properly(rhs).values_host
+    assert (got == [True, True, True, True, True, True, True, True]).all()
```
isVoid, line 446: `.all` isn't needed since python doesn't return an element wise compare result.

python/cuspatial/cuspatial/core/binpreds/contains.py
```diff
@@ -50,6 +47,13 @@ def contains_properly(
         within its corresponding polygon.
     """
 
+    scale = (
+        (test_points_x.std() + test_points_y.std()) / 2.0
+        if len(test_points_x) > 1
+        else 1.0
+    )
```
isVoid, line 54: Some comment explaining this?

python/cuspatial/cuspatial/core/geoseries.py
```diff
@@ -794,7 +796,7 @@ def reset_index(
             self.index = cudf_series.index
             return None
 
-    def contains_properly(self, other, align=True):
+    def contains_properly(self, other, align=True, allpairs=False):
```
isVoid, line 799: Update the document for the new argument, the new `polygon_index` column in the result. And better, give an example.

isVoid (changes requested):

cpp/src/join/quadtree_point_in_polygon.cu
```diff
@@ -268,7 +268,7 @@ std::unique_ptr<cudf::table> quadtree_point_in_polygon(cudf::table_view const& p
                     "number of rings must be no less than number of polygons");
   CUSPATIAL_EXPECTS(poly_points_x.size() == poly_points_y.size(),
                     "numbers of vertices must be the same for both x and y columns");
-  CUSPATIAL_EXPECTS(poly_points_x.size() >= 3 * ring_offsets.size(),
+  CUSPATIAL_EXPECTS(poly_points_x.size() >= 3 * (ring_offsets.size() - 1),
```
harrism, line 271: Why is this -1? If there are 2 rings, ring_offsets.size() would be 2 and there should be at least 6 vertices, not at least 3 vertices, right?

python/cuspatial/cuspatial/core/binpreds/contains.py
harrism: What happens if this heuristic is not good for the provided points? Do we still get correct results?

python/cuspatial/cuspatial/core/geoseries.py
harrism, line 799: Looking at geopandas APIs, I see a lot of named mode parameters rather than boolean. how is commonly used, e.g.

lanes_districts = gpd.sjoin(districts, bike_lane, how='inner', predicate='intersects')
lanes_districts

So should we use how="allpairs" (and how="pairwise") instead of a Boolean parameter? Perhaps how might be needed also for other things, so maybe we need a new name for this, e.g. parallelism="allpairs" with options of "all_pairs", "pairwise", "one_to_all", etc.

Thoughts?

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: The first time I discard the index, the second time I move the new index to become a dataframe column. It is the equivalent to saying `geom_df['part_index'] = cp.arange(len(geom_df))` I think, which is simpler than the above.

cpp/src/join/quadtree_point_in_polygon.cu
thomcom: This is a bug that was revealed by the `test_contains_properly.py` tests, which test a lot of single-polygons against single points and other geometries. The reason for the change is that the test was failing on very simple cases due to this precondition, tests that shouldn't have been failing. It turns out that `ring_offsets` is in `GeoArrow` format at this point, so a polygon with 1 ring with 3 vertices in it will have a `ring_offsets` = [0, 3] == length 2, but poly_points_x is only length 3, so it was `< 3 * ring_offsets.size()`. Because it is a GeoArrow structure with an extra end-offset, that should be subtracted from the length before comparing the number of vertices against the number of rings.

python/cuspatial/cuspatial/tests/binpreds/test_contains_properly.py
thomcom: In this case it does, I guess numpy is raising the type of the boolean list to an ndarray?

python/cuspatial/cuspatial/core/binpreds/contains.py
thomcom: After chatting with @trxcllnt decided to just set this to -1.

thomcom:
Closing this in favor of #981

python/cuspatial/cuspatial/core/geoseries.py
thomcom: I will rename from `mode` to `how`.

thomcom: I've actually renamed the parameter back to `allpairs=True` for now. I will file an issue to rename it to `how=` but I'd like to leave it as the boolean because that is the API for the GTC talk. Let's give the GTC talk then give it a week or two before renaming the API.

isVoid (changes requested):

cpp/src/utility/point_in_polygon.cuh
```diff
@@ -54,10 +57,26 @@ inline __device__ bool is_point_in_polygon(T const x,
...
+      // This can happen in polygon vertices when the first and last vertex of the ring are
+      // the same. In this scenario, do not attempt ray casting on a degenerate point.
+      T constexpr zero = 0.0;
+      if (float_equal(run, zero) && float_equal(rise, zero)) continue;
+
+      T rise_to_point = y - y0;
+      T run_to_point  = x - x0;
+
+      is_colinear = float_equal(run * rise_to_point, run_to_point * rise);
+      if (is_colinear) { break; }
```
trxcllnt, line 71:
```suggestion
      if (float_equal(run * rise_to_point, run_to_point * rise)) { return false; }
```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -168,23 +185,71 @@ def _op(self, lhs, points):
...
+        ).repeat(part_sizes)
+        parts_df = parts_map.reset_index(drop=True).reset_index()
+        # Mapping of parts to polygons
+        parts_to_geoms = cp.array(self.lhs.polygons.geometry_offset)
+        geometry_sizes = parts_to_geoms[1:] - parts_to_geoms[:-1]
+        geometry_map = cudf.Series(
+            cp.arange(len(geometry_sizes)), name="polygon_index"
+        ).repeat(geometry_sizes)
+        geom_df = geometry_map.reset_index(drop=True)
+        geom_df.index.name = "part_index"
```
harrism, line 219: Put this complex block of code in a clearly named function with clearly named parameters.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -168,23 +185,71 @@ def _op(self, lhs, points):
...
+        # Replace the polygon index with the original index
+        result["polygon_index"] = result["polygon_index"].replace(
+            cudf.Series(self.lhs.index, index=cp.arange(len(self.lhs.index)))
         )
-        return point_result
+        # Using allpairs for all requests with more than 31 polygons.
+        if not self.allpairs:
+            if len(result) == 0:
+                return cudf.Series([False] * len(self.lhs))
+            final_result = cudf.Series([False] * len(point_indices))
```
harrism, line 236: Names like "result" and "final_result" make this more confusing than it needs to be. Name the intermediates clearly to describe what they contain.

isVoid (approved):

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -168,23 +185,71 @@ def _op(self, lhs, points):
...
 
-    def postprocess(self, point_indices, point_result):
-        """Postprocess the output GeoSeries to ensure that they are of the
-        correct type for the predicate."""
-        result = cudf.DataFrame({"idx": point_indices, "pip": point_result})
-        df_result = result
+    def _postprocess_brute_force_result(self, point_indices, point_result):
+        # If there are 31 or fewer polygons in the input, the result
+        # is a dataframe with one row per point and one column per
+        # polygon.
```
harrism, line 248: One option would just be to change this. For the small-polygon-count version, just convert the bitmask into the same format output as the quadtree (all pairs): one row per intersection, containing a polygon index and a point index.

Then contains_properly only needs to convert the results if the user requests pairwise instead of all-pairs.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -154,11 +168,14 @@ def preprocess(self, lhs, rhs):
         point_indices = geom.point_indices()
         from cuspatial.core.geoseries import GeoSeries
 
-        final_rhs = GeoSeries(
-            GeoColumn._from_points_xy(xy_points._column)
-        ).points
+        final_rhs = GeoSeries(GeoColumn._from_points_xy(xy_points._column))
         return (lhs, final_rhs, point_indices)
 
+    def _should_use_quadtree(self):
```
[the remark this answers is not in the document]
thomcom: I thought the code for this function was self explanatory: if there are more than 31 polygons in the lhs, it has multipolygons, or the user specified allpairs.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: Great idea, this is easy to do and exposes what the function does in a way that helps majorly with comprehension. +1 for adding complexity this time for sure.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -168,23 +185,71 @@ def _op(self, lhs, points):
...
+
+        # Replace the part index with the polygon index
+        part_result = parts_df.merge(point_result, on="part_index")
+        # Replace the polygon index with the row index
+        result = geom_df.merge(part_result, on="part_index")
+        result = result[["polygon_index", "point_index"]]
+        result = result.drop_duplicates()
+        # Replace the polygon index with the original index
+        result["polygon_index"] = result["polygon_index"].replace(
+            cudf.Series(self.lhs.index, index=cp.arange(len(self.lhs.index)))
```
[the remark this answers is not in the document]
thomcom: This suggestion combined with Mark's suggestion above to place this code into a function with well defined inputs and outputs greatly increases the comprehensibility of this area, thanks.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -168,23 +185,71 @@ def _op(self, lhs, points):
...
         )
-        return point_result
+        # Using allpairs for all requests with more than 31 polygons.
```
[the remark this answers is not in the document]
thomcom: Fixed.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: Agreed, I was being too terse and have expanded the variable names to meaningful language.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -193,15 +258,45 @@ def postprocess(self, point_indices, point_result):
...
-        point_result.name = None
-        return point_result
+            result = df_result
+
+        final_result = cudf.Series([False] * len(self.lhs))
+
+        if len(result.columns) > 1:
+            final_result[result.index] = cp.diag(result.values)
+        else:
+            final_result[result.index] = result[result.columns[0]]
```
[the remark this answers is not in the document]
thomcom: Done.

python/cuspatial/cuspatial/utils/join_utils.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
thomcom: Ah whoops I had removed it in another branch...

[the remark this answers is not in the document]
thomcom: Done.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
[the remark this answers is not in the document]
isVoid: I think you should mention this is because the byte limited API supports only 31 polygons, and list reasons why you don't use pairwise_PiP instead. Otherwise this would look like arbitrary decisions to external contributors.

python/cuspatial/cuspatial/core/spatial/bounding.py
```diff
@@ -51,9 +51,7 @@ def polygon_bounding_boxes(polygons: GeoSeries):
     # by combining the geometry offset and parts offset of the multipolygon
     # array.
 
-    poly_offsets = polygons.polygons.part_offset.take(
-        polygons.polygons.geometry_offset
-    )
+    poly_offsets = polygons.polygons.part_offset
```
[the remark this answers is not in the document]
thomcom: This change specifically breaks multi-polygon point in polygon testing, because it discards the notion we talked about yesterday that quadtree pip has no notion of multipolygons, and we take advantage of that at the python layer.

I think this effectively collapses MultiPolygons into Polygons, combining all the rings of the MultiPolygon into a single polygon. I'm not sure that's correct behavior anyway because then the exterior ring of the second polygon in a MultiPolygon becomes an interior ring, right?

python/cuspatial/cuspatial/core/binpreds/binpreds.py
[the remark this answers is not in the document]
thomcom: Done.

cpp/src/utility/point_in_polygon.cuh
thomcom: Done and done.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: done

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: This is a valid option but I can't think of a way to do it with `cudf` operations and no iteration.

thomcom: Turns out it is really easy to change the output format with a `cudf.stack()` call, so I'm investigating this approach to reduce code-paths.

python/cuspatial/cuspatial/core/spatial/bounding.py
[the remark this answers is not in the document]
isVoid: Thanks, this is clear now. Can you add a line in the doc to mention that multipolygon bounding box is unsupported now?

python/cuspatial/cuspatial/core/binpreds/binpreds.py
thomcom: I've finished that refactor, it makes solving the discrete math problem at the end harder but I'm ok with it for the current iteration of the architecture.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -154,11 +168,14 @@ def preprocess(self, lhs, rhs):
...
 
-        final_rhs = GeoSeries(
-            GeoColumn._from_points_xy(xy_points._column)
-        ).points
+        final_rhs = GeoSeries(GeoColumn._from_points_xy(xy_points._column))
         return (lhs, final_rhs, point_indices)
 
+    def _should_use_quadtree(self):
+        return (
+            len(self.lhs) >= 32 or has_multipolygons(self.lhs) or self.allpairs
```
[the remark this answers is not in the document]
thomcom: So that I don't have to resolve the multipolygon results along two different code paths.

cpp/src/utility/point_in_polygon.cuh
```diff
@@ -32,6 +34,7 @@ inline __device__ bool is_point_in_polygon(T const x,
                                            cudf::column_device_view const& poly_points_y)
 {
   bool in_polygon     = false;
+  bool is_colinear    = false;
```
trxcllnt, line 37:
```suggestion
```

cpp/src/utility/point_in_polygon.cuh
```diff
@@ -54,10 +57,25 @@ inline __device__ bool is_point_in_polygon(T const x,
...
+
+      // If points are on the polygon edge, they are not contained in the polygon.
+      if (float_equal(run * rise_to_point, run_to_point * rise)) { return false; }
 
       if (y_in_bounds && x < (run / rise) * rise_to_point + x0) { in_polygon = not in_polygon; }
     }
+    if (is_colinear) {
+      in_polygon = false;
+      break;
+    }
```
trxcllnt, line 78: And now that nothing is setting `is_colinear`, we can remove L37 and L75-L78

```suggestion
```

trxcllnt (approved):

cpp/src/utility/point_in_polygon.cuh
thomcom: oh good idea :p

PR #912, opened 2023-02-09
Reduce error handling verbosity in CI tests scripts

AjayThorve:
This PR adds a less verbose [trap method](https://github.com/rapidsai/cugraph/blob/f2b081075704aabc789603e14ce552eac3fbe692/ci/test.sh#L19), for error handling to help ensure that we capture all potential error codes in our test scripts, and works as follows:

- setting an environment variable, EXITCODE, with a default value of 0
- setting a trap statement triggered by ERR signals which will set EXITCODE=1 when any commands return a non-zero exit code

cc @ajschmidt8


ci/test_cpp.sh
```diff
@@ -35,6 +34,8 @@ rapids-mamba-retry install \
 rapids-logger "Check GPU usage"
 nvidia-smi
 
+EXITCODE=0
+trap "EXITCODE=1" ERR
```
ajschmidt8, line 38: Can you remove the `TODO` a few lines down?

ajschmidt8 (changes requested):

ci/test_cpp.sh
AjayThorve: resolved, thanks!

ajschmidt8 (approved):
:fire: :fire:

ajschmidt8:
:fire:

Issue #885, opened 2023-02-02
[BUG]: Inconsistent values produced by directed_hausdorff_distance across different versions of cuspatial


abhilshitsoni-tomtom:
### Version

22.02

### On which installation method(s) does this occur?

Docker

### Describe the issue

I am having 2 parallel setups to test cuSpatial on my databricks runtime that uses docker container 

1. The first one uses docker container built using https://github.com/databricks/containers/tree/master/ubuntu/gpu/cuda-11/rapids (image officially published on dockerhub as databricksruntime/gpu-rapids:cuda11 ) this uses rapids 22.02 with cuda11.2 and is deployed on databricks with following settings 

Databricks Runtime:  9.1 LTS (includes Apache Spark 3.1.2, Scala 2.12)
Worker Type : Standard NC6_v3 (uses NVIDIA Tesla V100 GPU)
Driver Type : Standard NC6_v3  (uses NVIDIA Tesla V100 GPU)
Docker Image URL:  databricksruntime/gpu-rapids:cuda11

2. The second setup uses docker container built using https://github.com/databricks/containers/tree/master/ubuntu/gpu/cuda-10.1/rapids (image officially published on dockerhub as databricksruntime/gpu-rapids:cuda10.1 ) this uses rapids 0.13 with cuda 10.1 and is deployed on databricks with following settings 

Databricks Runtime:  7.3 LTS (includes Apache Spark 3.0.1, Scala 2.12)
Worker Type : Standard NC6_v3 (uses NVIDIA Tesla V100 GPU)
Driver Type : Standard NC6_v3  (uses NVIDIA Tesla V100 GPU)
Docker Image URL:  databricksruntime/gpu-rapids:cuda10.1

My usecase is similar to trajectory clustering where I need to calculate a pairwise directed_hausdorff_distance matrix to be provided to DBSCAN. However, both systems are giving inconsistent results of distance matrix on same data. On inspecting further I tried to run the directed_hausdorff_distance example code for both the versions taken from here 

https://github.com/rapidsai/cuspatial/blob/branch-22.02/python/cuspatial/demos/hausdorff_clustering_test_toy.py
Here is the code snippet followed by its output on both the setups

```
import numpy as np
from sklearn.cluster import DBSCAN, AgglomerativeClustering

from cudf import Series

import cuspatial

in_trajs = []
in_trajs.append(np.array([[1, 0], [2, 1], [3, 2], [5, 3], [7, 1]]))
[...]

PR #916, opened 2023-02-10
Improve Hausdorff Many Column Performance

isVoid:
## Description

This PR changes the C++ hausdorff distance interface to return the owning item that contains the interface and a table view into the result. This avoids a numba reshape that slows performance when there are many columns. 

Python API benchmark (10K input spaces)
branch-23.06:
```
--------------------------------------------------------- benchmark: 1 tests ---------------------------------------------------------
Name (time in s)                                     Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------
bench_directed_hausdorff_distance_many_spaces     1.4473  1.4938  1.4688  0.0168  1.4655  0.0175       2;0  0.6808       5           1
--------------------------------------------------------------------------------------------------------------------------------------
```

this pr:
```
-------------------------------------------------------------- benchmark: 1 tests --------------------------------------------------------------
Name (time in ms)                                      Min       Max      Mean   StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------
bench_directed_hausdorff_distance_many_spaces     221.5014  288.3597  237.1724  28.7577  223.4605  21.0320       1;1  4.2163       5           1
------------------------------------------------------------------------------------------------------------------------------------------------
```

6.5x speedup.

contributes to #1013 

The CPP benchmarks shows no significant performance regression:
```
Comparing branch2304.json to after.json
Benchmark                                                           Time             CPU      Time Old      Time New       CPU Old       CPU New
------------------------------------------------------------------------------------------------------------------------------------------------
HausdorffBenchmark/hausdorff/32/4/manual_time                    -0.0369         -0.0519             0             0             1             0
HausdorffBenchmark/hausdorff/64/4/manual_time                    -0.0245         -0.0480             0             0             1             1
HausdorffBenchmark/hausdorff/512/4/manual_time                   -0.0173         -0.0125             1             1             1             1
HausdorffBenchmark/hausdorff/4096/4/manual_time                  +0.0603         +0.1068             8             8             8             9
HausdorffBenchmark/hausdorff/8192/4/manual_time                  -0.0256         -0.0240            18            18            18            18
HausdorffBenchmark/hausdorff/32/8/manual_time                    -0.0068         +0.0780             0             0             1             1
HausdorffBenchmark/hausdorff/64/8/manual_time                    -0.0168         -0.0213             0             0             1             1
HausdorffBenchmark/hausdorff/512/8/manual_time                   -0.0216         -0.0242             1             1             2             2
HausdorffBenchmark/hausdorff/4096/8/manual_time                  -0.0296         -0.0092            12            12            13            13
HausdorffBenchmark/hausdorff/8192/8/manual_time                  +0.0314         +0.0328            28            29            28            29
HausdorffBenchmark/hausdorff/32/64/manual_time                   -0.0139         -0.0879             0             0             1             1
HausdorffBenchmark/hausdorff/64/64/manual_time                   -0.0001         +0.0343             0             0             1             1
HausdorffBenchmark/hausdorff/512/64/manual_time                  -0.0935         -0.0816             5             5             6             6
HausdorffBenchmark/hausdorff/4096/64/manual_time                 -0.0299         -0.0295           186           180           187           181
HausdorffBenchmark/hausdorff/8192/64/manual_time                 -0.1097         -0.1094           685           610           685           610
HausdorffBenchmark/hausdorff/32/128/manual_time                  +0.0065         +0.1026             0             0             1             1
HausdorffBenchmark/hausdorff/64/128/manual_time                  +0.0035         -0.0169             1             1             1             1
HausdorffBenchmark/hausdorff/512/128/manual_time                 -0.0312         -0.0299            10            10            11            10
HausdorffBenchmark/hausdorff/4096/128/manual_time                -0.0388         -0.0388           613           589           613           589
HausdorffBenchmark/hausdorff/8192/128/manual_time                -0.0080         -0.0080          2353          2335          2354          2335

```

## Checklist


isVoid:
Planning to address https://github.com/rapidsai/cuspatial/issues/885 in this PR too. Switching to draft.

isVoid:
This is now ready for review.

cpp/src/spatial/hausdorff.cu
```diff
@@ -80,18 +85,23 @@ struct hausdorff_functor {
                                            result_view.begin<T>(),
                                            stream);
 
-    return result;
+    thrust::host_vector<cudf::size_type> splits(num_spaces - 1);
+    thrust::sequence(thrust::host, splits.begin(), splits.end(), num_spaces, num_spaces);
+
+    return {std::move(result),
+            cudf::table_view(cudf::detail::split(result->view(), splits, stream))};
```
harrism, line 92: Shouldn't be calling into cudf::detail...

cpp/src/spatial/hausdorff.cu
```diff
@@ -80,18 +85,23 @@ struct hausdorff_functor {
                                            result_view.begin<T>(),
                                            stream);
 
-    return result;
+    thrust::host_vector<cudf::size_type> splits(num_spaces - 1);
+    thrust::sequence(thrust::host, splits.begin(), splits.end(), num_spaces, num_spaces);
```
harrism, line 89: Could this be on the device?

cpp/include/cuspatial/distance/hausdorff.hpp
```diff
@@ -87,7 +88,7 @@ namespace cuspatial {
  *
  * @note Hausdorff distances are asymmetrical
  */
-std::unique_ptr<cudf::column> directed_hausdorff_distance(
+std::pair<std::unique_ptr<cudf::column>, cudf::table_view> directed_hausdorff_distance(
```
harrism, line 91: Documentation has to be updated. Also, this is an API breaking change.

harrism (changes requested):
Need more explanation of this in the documentation.

cpp/src/spatial/hausdorff.cu
isVoid: Only detail API exposes the stream parameter though...

cpp/src/spatial/hausdorff.cu
isVoid: No, `cudf.split` takes `host_span` as its argument.

harrism: But do we really need to use cudf::split to do this operation? It's all just views, right?

python/cuspatial/cuspatial/core/spatial/distance.py
```diff
@@ -91,10 +91,8 @@ def directed_hausdorff_distance(xs, ys, space_offsets):
         ys,
         as_column(space_offsets, dtype="uint32"),
     )
-    with cudf.core.buffer.acquire_spill_lock():
-        result = result.data_array_view(mode="read")
-        result = result.reshape(num_spaces, num_spaces)
-        return DataFrame(result)
+
+    return DataFrame._from_columns(result, range(num_spaces)).T
```
thomcom, line 95: Do you know if transpose is still considered to be unadvised? It used to be a substantial performance hit, maybe the transpose could be performed via thrust instead of here in python?

isVoid: It's probably still a performance hit. The kernel is (unfortunately written) in transposed order. I'll have to modify the kernel to avoid that.

cpp/src/spatial/hausdorff.cu
isVoid: That's true, we could hand-roll our split operation here. cuspatial's column doesn't contain any null, so we don't need to do null counting on device.

cpp/src/spatial/hausdorff.cu
isVoid: The only place where cuDF uses stream is with the null counting on device. Here we don't need to count nulls since the result doesn't contain any.

python/cuspatial/cuspatial/core/spatial/distance.py
isVoid: Would you say I should change the c++ kernel to avoid breaking python interface, or to keep as is and break the python interface?

thomcom: I'm inclined to break the python interface, the transposition of the result doesn't matter at all iirc.

thomcom: It's a square matrix with i,j or j,i, right?

isVoid: Yes, it will be a change in the order of the output coordinate.

harrism: At this stage it's still OK to break either API, as long as we tag it appropriately.

cpp/src/spatial/hausdorff.cu
harrism: OK. Just trying to avoid calling cuDF detail API...

python/cuspatial/cuspatial/core/spatial/distance.py
isVoid: Transpose is removed.

harrism (approved):

thomcom (approved):

PR #932, opened 2023-02-14
Remove shapefile reader, conda GDAL dependency, move cmake gdal dependency to cpp tests only

### LINKED ISSUE #931: [FEA]: Remove shapefile reader and remove libcuspatial GDAL dependency

New Feature

Medium

#872 deprecated the shapefile reader in release 23.02. We should remove it completely in 23.04 and also remove the GDAL dependency from libcuspatial. The quadtree Point-in-polygon C++ tests rely on GDAL for comparison so we can just move the dependency to the tests CMakelists.txt, and then update that test in a followup PR to use a different comparison.

_No response_

### Additional context

_No response_

harrism:
## Description

Closes #931.

Removes the shapefile reader (C++ and Python), tests and benchmarks.

Moves the C++ GDAL dependency from libcuspatial's CMakeLists.txt to tests/CMakeLists.txt since one test depends on GDAL.

## Checklist


dependencies.yaml
```diff
@@ -42,7 +42,6 @@ dependencies:
           - &cmake_ver cmake>=3.23.1,!=3.25.0
           - c-compiler
           - cxx-compiler
-          - gdal>3.5.0,<3.6.0
```
bdice, line 45: I would keep this line since gdal is required for building tests.

ajschmidt8:
I'm seeing this failure hidden in the C++ tests log

- https://github.com/rapidsai/cuspatial/actions/runs/4180825096/jobs/7242371229#step:6:849

image

dependencies.yaml
harrism: We are trying to remove the gdal dependency for building libcuspatial, to ease creating wheels. So is this really necessary if it's only for building tests?

bdice: This list should include everything needed to build libcuspatial from source. That includes tests, if we treat it the same as other RAPIDS packages.

This is used to generate a conda environment and does not affect wheel builds.

isVoid (approved):

dependencies.yaml
ajschmidt8: Re-added this in 4d72fa2.

ajschmidt8 (approved):

PR #945, opened 2023-02-18
Fix OB bug in `linestring_intersection_intermediates.remove_if` Function

isVoid:
## Description

This PR fixes a bug mentioned by @thomcom at https://github.com/rapidsai/cuspatial/pull/862#pullrequestreview-1303680652. When computing the lookup id for each group to find the number of geometry removed in previous groups, the previous computation is flawed and resulted in OB access.

Besides, this PR also adds `CUDA_TRY(cudaGetLastError())` in several kernel calls in the intersection code path.

This PR also refined the documentation for `remove_if` and related helper function.

## Checklist


cpp/include/cuspatial/experimental/detail/find/find_and_combine_segment.cuh
```diff
@@ -72,6 +73,8 @@ void find_and_combine_segment(OffsetRange offsets,
   auto [threads_per_block, num_blocks] = grid_1d(num_spaces);
   simple_find_and_combine_segments_kernel<<<num_blocks, threads_per_block, 0, stream.value()>>>(
     offsets, segments, merged_flag);
+
+  CUSPATIAL_CUDA_TRY(cudaGetLastError());
```
harrism, line 77: This will not catch errors if the kernel has not thrown one yet. Synchronization is required. Therefore I think you should use this instead. In a release build, it's the same as what you have. But in a debug build, it synchronizes the stream.

```suggestion
  CUSPATIAL_CHECK_CUDA(stream.value());
```

cpp/include/cuspatial/experimental/detail/find/find_duplicate_points.cuh
```diff
@@ -68,6 +69,8 @@ void find_duplicate_points(MultiPointRange multipoints,
   auto [threads_per_block, num_blocks] = grid_1d(multipoints.size());
   find_duplicate_points_kernel_simple<<<num_blocks, threads_per_block, 0, stream.value()>>>(
     multipoints, duplicate_flags);
+
+  CUSPATIAL_CUDA_TRY(cudaGetLastError());
```
harrism, line 73:
```suggestion
  CUSPATIAL_CHECK_CUDA(stream.value());
```

cpp/include/cuspatial/experimental/detail/linestring_intersection_count.cuh
```diff
@@ -84,10 +84,14 @@ void pairwise_linestring_intersection_upper_bound_count(MultiLinestringRange1 mu
...
                                                         rmm::cuda_stream_view stream)
 {
+  if (multilinestrings1.size() == 0) return;
+
   auto [threads_per_block, num_blocks] = grid_1d(multilinestrings1.size());
   detail::
     count_intersection_and_overlaps_simple<<<num_blocks, threads_per_block, 0, stream.value()>>>(
       multilinestrings1, multilinestrings2, points_count_it, segments_count_it);
+
+  CUSPATIAL_CUDA_TRY(cudaGetLastError());
```
harrism, line 94:
```suggestion
  CUSPATIAL_CHECK_CUDA(stream.value());
```

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -439,6 +525,8 @@ pairwise_linestring_intersection_with_duplicates(MultiLinestringRange1 multiline
       points.geoms->begin(),
       segments.geoms->begin());
 
+  CUSPATIAL_CUDA_TRY(cudaGetLastError());
```
harrism, line 528:
```suggestion
  CUSPATIAL_CHECK_CUDA(stream.value());
```

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
```diff
@@ -45,11 +46,30 @@ namespace detail {
...
+ *  previous removed geometry from `reduced_values` by searching `reduced_keys`.
+ *
+ *  Example:
+ *  i:                             0  1  2  3  4  5  6  7
+ *  Geometry Collection Offset:    0  0  0  1  3  4  4  4
+ *  Reduced Keys:                  3  4  5
+ *  j:                            -1 -1 -1  0  1  2  2  2
+ *  Reduced Values:                0  2  2                  (Removed 2 geometries)
+ *  Should subtract:               /  /  /  0  2  2  2  2
+ *  Result:                        0  0  0  0  1  2  2  2
```
harrism, line 61: I get everything up to the `j:` line. But after that I'm trying to follow what these last 3 lines mean, and what "removed 2 geometries" means.  The code at this point is: 

```
if (j < 0) return offset;

return offset - reduced_values_begin[j];
```

But what is `offset` in the example above? And how does this code correspond with the last 3 lines of the example?

harrism (changes requested):
Use `CUSPATIAL_CHECK_CUDA` instead. Plus one question.

cpp/include/cuspatial/experimental/detail/linestring_intersection_with_duplicates.cuh
isVoid: `offset` is `GeometryCollectionOffset[i]` in the example. 

`GeometryCollectionOffset[i]` should subtract the total number of removed geometries in all preceding lists as a result of `remove_if`. This number is `reduced_value[j]`. `0 2 2` means:
```
0 geometries are removed for the first non empty list
2 geometries are removed for the first and the second nonempty list
2 geometries are removed for the first, second and third nonempty list
In total, 2 geometries are removed.
```

To me, the most difficult part to grasp is why `j` can be achieved by searching `reduced_key`.  Recall in the example:
```
Geometry Collection Offset:    0  0  0  1  3  4  4  4
reduced key: 3 4 5
```

The values in `reduced_key` represents the *end position of the indices* (index of index) for the 3 non-empty lists in the column. e.g. `GeometryCollectionOffset[3] == 1` means the first non empty list ends at position 1 of the column.
This makes `reduced_key` the boundary condition for partitioning the `GeometryCollectionOffset` array.

isVoid: Docs are updated.

harrism (approved):

PR #974, opened 2023-03-07
Remove GDAL dependency in quadtree spatial join tests.

### LINKED ISSUE #958: [FEA]: Remove dependency on GDAL in libcuspatial point-in-polygon tests

Improvement

Medium

We have removed GDAL as a dependency (#932) in all but one C++ unit test: https://github.com/rapidsai/cuspatial/blob/branch-23.04/cpp/tests/join/point_in_polygon_test_large.cpp

GDAL is used for its `contains` functionality to verify our quadtree spatial join. This test only has a few polygons, so we could use our all-pairs point-in-polygon API which is limited to 31 polygons for verification, and drop the GDAL dependency completely, which would help with building PIP wheel unit tests.

_No response_

### Additional context

_No response_

harrism:
## Description
Closes #958.  Replaces usage of GDAL `contains` in the quadtree point-in-polygon tests with a call to a different cuSpatial point-in-polygon function. This allows us to remove all current dependencies on GDAL in cuSpatial.

Depends on #973. This PR currently updates some expectations to make tests pass that are fixed in that PR. 

## Checklist


cpp/tests/join/point_in_polygon_test_large.cpp
```diff
@@ -235,57 +172,47 @@ TYPED_TEST(PIPRefineTestLarge, TestLarge)
...
-  auto point_a = fixed_width_column_wrapper<uint32_t>(expected_point_indices.begin(),
-                                                      expected_point_indices.end());
-  auto point_b =
-    fixed_width_column_wrapper<uint32_t>(actual_point_indices.begin(), actual_point_indices.end());
-  CUDF_TEST_EXPECT_COLUMNS_EQUAL(point_a, point_b, verbosity);
+    for (int point_index = 0; point_index < hits->size(); point_index++) {
+      std::uint32_t bits = hits_host[point_index];
+      while (bits != 0) {
+        std::uint32_t t          = bits & -bits;          // get only LSB
+        std::uint32_t poly_index = __builtin_ctzl(bits);  // get index of LSB
```
isVoid, line 196: Since it's `int` type here, shouldn't the most compatible function is `__builtin_ctz` instead of `__builtin_ctzl`?

isVoid (approved):
No more gdal dependency!

thomcom (approved):

cpp/tests/join/point_in_polygon_test_large.cpp
```diff
@@ -235,57 +172,48 @@ TYPED_TEST(PIPRefineTestLarge, TestLarge)
...
+    std::vector<uint32_t> expected_point_indices;
 
-  auto point_a = fixed_width_column_wrapper<uint32_t>(expected_point_indices.begin(),
-                                                      expected_point_indices.end());
-  auto point_b =
-    fixed_width_column_wrapper<uint32_t>(actual_point_indices.begin(), actual_point_indices.end());
-  CUDF_TEST_EXPECT_COLUMNS_EQUAL(point_a, point_b, verbosity);
+    for (int point_index = 0; point_index < hits->size(); point_index++) {
+      // iterate over set bits
+      std::uint32_t bits = hits_host[point_index];
```
trxcllnt, line 194: Small nit: `bits` is unsigned here, but `hits_host` is signed.

Maybe a larger question, but why does `cuspatial::point_in_polygon` return an INT32 column vs. a UINT32? Does the sign bit indicate something special? If not, unsigned conveys the sign bit isn't special and the column is a bitmask.

Alternatively, we could return a column with empty `data` and a full `null_mask`. This would let us test whether values were null without reading the bits ourselves?

harrism: OMG, I think it may be because cuDF didn't support unsigned types until here: https://github.com/rapidsai/cudf/pull/5292

jjacobelli (approved):

PR #976, opened 2023-03-07
Add `point_polygon_distance`, add non-owning polygon objects

### LINKED ISSUE #703: Introduce data structure to simplify multigeometry API
As demonstrated in https://github.com/rapidsai/cuspatial/pull/677, the geometry input of the header only API can quickly get out of control for complex geometry. We should try to simplify the APIs to improve developer experience.

## Device view over the physical memory of nested types
The data structure should be a view over the physical memory layouts, thus it should be cheap to construct the data structure on host, passing the structure to device and invoke the method on device.

Unlike `cudf::column_view` and `cudf::column_device_view`, the data structure is completely templated (not type erased), and always assumes the view are device views.

## Assumes GeoArrow memory layout (#649)

The structures holds a view to offset arrays and point arrays. Offset arrays are assumed to always conform to arrow's offset array layout, which is specified in (https://arrow.apache.org/docs/format/Columnar.html#variable-size-binary-layout).

## Accessors

Due to the complex nature of geometries, developers may want to support different patterns of access from the data structure.

### Element-wise accessors (top-down access)

The simplest of all is element-wise accessor. A kernel is launched on a per-geometry level, accessors like `.element(idx)` should return a geometry object from the array.

### Component-wise accessors (bottom-up access)

Other parallel patterns may require a thread to work on one component from the array, such as paralleled on a point of a `multilinestring_array`. In this case bottom-up traversal utilities should be supported, and should be easily implemented by a binary search in the offset array.

### LINKED ISSUE #756: [FEA] Pairwise Point Polygon Distance
Part of #231 , to support all geometry pairwise distance, point-polygon distance is needed.

We already have point-linestring distance, we need to check if the point is contained in the polygon. If not, compute shortest distance to all rings.

isVoid:
## Description

This PR adds `multipolygon_range`, `multipolygon_ref`, `polygon_ref` as non-owning objects for geoarrow compliant polygon types.

This PR also adds `pairwise_point_polygon_distance` to compute the shortest distances between two columns of multipoints and multilinestrings.

closes #703, closes #756

## Checklist


harrism:
> This PR also adds pairwise_point_polygon_distance to compute the shortest distances between two columns of multipoints and multilinestrings.

Do you mean multipolygons not multilinestrings?

isVoid:
> > This PR also adds pairwise_point_polygon_distance to compute the shortest distances between two columns of multipoints and multilinestrings.
> 
> Do you mean multipolygons not multilinestrings?

Yes. Should be multipolygons.

cpp/include/cuspatial/experimental/ranges/multipolygon_range.cuh
```diff
@@ -0,0 +1,157 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
```
isVoid, line 2:
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.
```

cpp/include/cuspatial/experimental/ranges/multipolygon_range.cuh
```diff
@@ -0,0 +1,157 @@
...
+  /// Returns the `multipolygon_idx`th multipolygon in the range.
+  template <typename IndexType>
+  CUSPATIAL_HOST_DEVICE auto operator[](IndexType multipolygon_idx);
+
+  /// Returns the `segment_idx`th segment in the multipolygon range.
+  template <typename IndexType>
+  CUSPATIAL_HOST_DEVICE auto get_segment(IndexType segment_idx);
+
+  /// Returns `true` if `point_idx`th point is the first point of its
+  /// multipolygon
```
isVoid, line 120:
```suggestion
  /// Returns `true` if `point_idx`th point is the first point of `geometry_idx`th
  /// multipolygon
```

cpp/include/cuspatial/experimental/detail/ranges/multipolygon_range.cuh
```diff
@@ -0,0 +1,282 @@
...
+  static_assert(is_vec_2d<iterator_value_type<VecIterator>>(),
+                "Coordinate range must be constructed with iterators to vec_2d.");
+}
+
+template <typename GeometryIterator,
+          typename PartIterator,
+          typename RingIterator,
+          typename VecIterator>
+CUSPATIAL_HOST_DEVICE auto
+multipolygon_range<GeometryIterator, PartIterator, RingIterator, VecIterator>::num_multipolygons()
```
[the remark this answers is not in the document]
thomcom: If this is the only file that it is used in, then no. If it needs to be repeated throughout, then maybe.

cpp/include/cuspatial/experimental/point_polygon_distance.cuh
```diff
@@ -0,0 +1,45 @@
...
+ * @return Output Iterator past the last distance computed
+ *
+ * [LinkLRAI]: https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator
+ * "LegacyRandomAccessIterator"
+ */
+template <class MultiPointRange, class MultiPolygonRange, class OutputIt>
+OutputIt pairwise_point_polygon_distance(MultiPointRange multipoints,
+                                         MultiPolygonRange multipoiygons,
+                                         OutputIt distances_first,
+                                         rmm::cuda_stream_view stream = rmm::cuda_stream_default);
```
thomcom, line 42: Params do not match the docs. Returning an `OutputIt` and also taking one as undocumented argument seems fishy.

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -50,6 +51,63 @@ auto make_device_uvector(std::initializer_list<T> inl,
...
+                     RingArray ring_offsets_array,
+                     CoordinateArray coordinate_offsets_array)
+    : _geometry_offsets_array(geometry_offsets_array),
+      _part_offsets_array(part_offsets_array),
+      _ring_offsets_array(ring_offsets_array),
+      _coordinate_offsets_array(coordinate_offsets_array)
+  {
+  }
+
+  /// Return the number of multilinestrings
```
thomcom, line 76:
```suggestion
  /// Return the number of multipolygons
```

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -50,6 +51,63 @@ auto make_device_uvector(std::initializer_list<T> inl,
...
+  auto size() { return _geometry_offsets_array.size() - 1; }
+
+  /// Return range object of the multilinestring array
```
thomcom, line 79:
```suggestion
  /// Return range object of the multipolygon array
```

cpp/tests/experimental/spatial/point_polygon_distance_test.cu
```diff
@@ -0,0 +1,427 @@
...
+
+  CUSPATIAL_RUN_TEST(
+    this->run_single,
+    {{P{0, 2}, P{0, 0}}, {P{1, 1}, P{-1, -1}}},
+    {0, 1, 2},
+    {0, 1, 2},
+    {0, 5, 9},
+    {P{-1, -1}, P{1, -1}, P{1, 1}, P{-1, 1}, P{-1, -1}, P{-1, 1}, P{1, 1}, P{0, -1}, P{-1, 1}},
+    {0.0, 0.0});
+}
```
thomcom, line 427: Do you have any hole tests in here? Are holes currently supported?

Consider a square polygon of radius 2 containing an interior ring of radius 1 and a point in the middle. Is the distance of the point from the polygon 0.5?

cpp/include/cuspatial/detail/utility/floating_point.cuh
```diff
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#pragma once
```
harrism, line 16: Update copyright.

cpp/include/cuspatial/detail/utility/offset_to_keys.cuh
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <thrust/binary_search.h>
+#include <thrust/distance.h>
+
+/** @brief Given list offset and row `i`, return a unique key that represent the list of `i`.
```
harrism, line 22: Why is this a "unique key"? It's just the index at which to insert `i` into the offset array to keep it sorted.  It's not unique, as your example shows: both 1 and 2 map to index 4.

harrism, line 22: The only difference between this functor and `upper_bound` is that it returns an index rather than an iterator. So just use the same style of docs.

```suggestion
/** @brief Returns the index of the first element in the range `[first, last)` such that `value < element` is true (i.e. strictly greater), or `distance(first, last)` if no such element is found.
```

And update the parameters names to match...

cpp/include/cuspatial/detail/utility/offset_to_keys.cuh
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *  The key is computed by performing a `upper_bound` search with `i` in the offset array.
+ *  Then subtracts the position with the start of offset array.
```
harrism, line 25: This implementation detail isn't needed in the example
.
```suggestion
```

cpp/include/cuspatial/detail/utility/offset_to_keys.cuh
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *  Example:
+ *  offset:  0 0 0 1 3 4 4 4
+ *  i:       0 1 2 3
+ *  key:     3 4 4 5
+ *
+ *  Note that the values of `key`, {offset[3], offset[4], offset[5]} denotes the ending
+ *  position of the first 3 non-empty list.
```
harrism, line 33: I don't understand this note.

cpp/include/cuspatial/detail/utility/offset_to_keys.cuh
```diff
@@ -0,0 +1,51 @@
...
+ */
+template <typename Iterator>
+struct offsets_to_keys_functor {
```
harrism, line 36: Seems a better name would be `upper_bound_index`, based on my explanation above?

cpp/include/cuspatial/experimental/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -30,44 +33,34 @@ namespace detail {
...
-    Cart2d b     = poly_points_first[ring_end - 1];
+    auto b       = last_segment.v2;
     bool y0_flag = b.y > test_point.y;
     bool y1_flag;
-    // for each line segment, including the segment between the last and first vertex
-    for (auto point_idx = ring_begin; point_idx < ring_end; point_idx++) {
-      Cart2d const a = poly_points_first[point_idx];
-      T run          = b.x - a.x;
-      T rise         = b.y - a.y;
+    auto ring_points = multipoint_ref{ring.point_begin(), ring.point_end()};
```
harrism, line 59: Cool.

cpp/include/cuspatial/experimental/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -100,5 +93,30 @@ __device__ inline bool is_point_in_polygon(Cart2d const& test_point,
 
   return point_is_within;
 }
+
+/**
+ * @brief Compatibility layer with non-OOP style input
```
harrism, line 98: Is this compatibility layer needed somewhere?

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
```diff
@@ -0,0 +1,170 @@
...
+  template <typename IndexType>
+  uint8_t __device__ operator()(IndexType pidx)
+  {
+    auto point        = thrust::raw_reference_cast(multipoints.point(pidx));
+    auto geometry_idx = multipoints.geometry_idx_from_point_idx(pidx);
+
+    bool intersects = false;
+    for (auto polygon : multipolygons[geometry_idx]) {
+      intersects = intersects || is_point_in_polygon(point, polygon);
+    }
```
harrism, line 68: Consider this option. `any_of` makes the intent clearer than a raw loop.

```suggestion
    auto const& polys = multipolygons[geometry_idx];
    bool intersects = thrust::any_of(thrust::seq, polys.begin(), polys.end(), [] (auto poly) { 
        return is_point_in_polygon(point, poly); 
    });
```

harrism: Not sure about performance, though.  any_of can early exit, at least. (You could add early exit to your loop, too.)

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
```diff
@@ -0,0 +1,170 @@
...
+          class IntersectionRange,
+          class OutputIterator>
+void __global__ pairwise_point_polygon_distance_kernel(MultiPointRange multipoints,
+                                                       MultiPolygonRange multipolygons,
+                                                       IntersectionRange intersects,
+                                                       OutputIterator distances)
+{
+  using T = typename MultiPointRange::element_t;
+
+  for (auto idx = threadIdx.x + blockIdx.x * blockDim.x; idx < multipolygons.num_points();
```
harrism, line 87: So iterating over segments in all multipolygons? Can this loop be made clearer with one of your refs?

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
```diff
@@ -0,0 +1,170 @@
...
+       idx += gridDim.x * blockDim.x) {
+    auto geometry_idx = multipolygons.geometry_idx_from_segment_idx(idx);
+    if (geometry_idx == MultiPolygonRange::INVALID_INDEX) continue;
+
+    if (intersects[geometry_idx]) {
+      // Leading thread of the pair writes to the output
+      if (multipolygons.is_first_point_of_multipolygon(idx, geometry_idx))
+        distances[geometry_idx] = T{0.0};
```
harrism, line 95: Does it matter if more than one thread writes zero to distances[geometry_idx]? Is the leader selection necessary?

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
```diff
@@ -0,0 +1,170 @@
...
+      continue;
+    }
+
+    T dist_squared = std::numeric_limits<T>::max();
+    auto [a, b]    = multipolygons.get_segment(idx);
+    for (vec_2d<T> point : multipoints[geometry_idx]) {
```
harrism, line 101: `thrust::reduce(thrust::seq, ...)`? Not sure about performance.

cpp/include/cuspatial/experimental/detail/ranges/multilinestring_range.cuh
```diff
@@ -29,6 +28,7 @@
 #include <cuspatial/vec_2d.hpp>
 
 #include <iterator>
+#include <optional>
```
harrism, line 31: Update copyright.

cpp/include/cuspatial/experimental/detail/ranges/multipoint_range.cuh
```diff
@@ -63,6 +63,8 @@ multipoint_range<GeometryIterator, VecIterator>::multipoint_range(GeometryIterat
     _points_begin(points_begin),
```
harrism, line 63: copyright.

cpp/include/cuspatial/experimental/detail/ranges/multipolygon_range.cuh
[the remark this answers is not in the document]
harrism: This is the detail implementation file. It's fine. Alternative is to put it inline in the declaration in the header, but I think this is better. Nobody will be typing this, and you would use find to jump to the function you need to look at in this file anyway.

cpp/include/cuspatial/experimental/geometry/linestring_ref.cuh
```diff
@@ -38,6 +38,11 @@ class linestring_ref {
   /// Return iterator to one past the last segment
```
harrism, line 38: copyright.

cpp/include/cuspatial/experimental/geometry/segment.cuh
```diff
@@ -19,6 +19,8 @@
 #include <cuspatial/cuda_utils.hpp>
```
harrism, line 19: copyright

cpp/include/cuspatial/experimental/point_polygon_distance.cuh
harrism: It's standard for most of our algorithms, just like STL. Returns an iterator to one past the last computed results, as the @return doc says.

I think the doc for `@tparam OutputIt` needs more detail to say that this must contain a floating point type.

cpp/include/cuspatial/experimental/point_polygon_distance.cuh
```diff
@@ -0,0 +1,45 @@
...
+#include <rmm/cuda_stream_view.hpp>
+
+namespace cuspatial {
+
+/**
+ * @ingroup distance
+ * @brief Computes pairwise multipoint to multipolygon distance
+ *
+ * @tparam MultiPointRange An instance of template type `MultiPointRange`
+ * @tparam MultiPolygonRange An instance of template type `MultiPolygonRange`
```
harrism, line 28: These docs are a bit self-referential!

cpp/include/cuspatial/experimental/point_polygon_distance.cuh
```diff
@@ -0,0 +1,45 @@
...
+ * @tparam OutputIt iterator type for output array. Must meet the requirements of [LRAI](LinkLRAI).
+ *
+ * @param multipoints Range of multipoints, one per computed distance pair.
+ * @param multipolygons Range of multilinestrings, one per computed distance pair.
+ * @return Output Iterator past the last distance computed
```
harrism, line 33:
```suggestion
 * @stream ...
 * @return Output Iterator past the last distance computed
```

cpp/include/cuspatial/experimental/ranges/multipoint_range.cuh
```diff
@@ -51,6 +51,8 @@ class multipoint_range {
   using point_t       = iterator_value_type<point_it_t>;
   using element_t     = iterator_vec_base_type<point_it_t>;
 
+  int32_t INVALID_IDX = -1;
```
harrism, line 54: const or constexpr? 

```suggestion
  constexpr int32_t INVALID_IDX = -1;
```

cpp/include/cuspatial/experimental/ranges/multipoint_range.cuh
```diff
@@ -51,6 +51,8 @@ class multipoint_range {
   using point_t       = iterator_value_type<point_it_t>;
```
harrism, line 51: copyright

cpp/include/cuspatial/experimental/ranges/multipolygon_range.cuh
```diff
@@ -0,0 +1,158 @@
...
+  CUSPATIAL_HOST_DEVICE auto multipolygon_end();
+
+  /// Return the iterator to the first multipolygon in the range.
+  CUSPATIAL_HOST_DEVICE auto begin() { return multipolygon_begin(); }
+
+  /// Return the iterator to the one past the last multipolygon in the range.
+  CUSPATIAL_HOST_DEVICE auto end() { return multipolygon_end(); }
+
+  /// Given the index of a segment, return the geometry (multipolygon) index where the
+  /// segment locates.
```
harrism, line 105:
```suggestion
  /// Given the index of a segment, return the index of the geometry (multipolygon) that contains the segment.
```

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -16,6 +16,7 @@
```
harrism, line 16: copyright.

harrism (changes requested):
Great work. I love the `refs` and `ranges`. They make the library much more maintainable and the code more readable.

cpp/include/cuspatial/experimental/point_polygon_distance.cuh
isVoid: > must contain a floating point type.

More precisely, "must be convertible from floating point type", since user can use output transform iterator to convert the result and transform into other types.

cpp/tests/experimental/spatial/point_polygon_distance_test.cu
isVoid: Holes are supported, see test: 
- `OnePairOnePolygonTwoRings`
- `OnePairOnePolygonTwoRings2`
- `OnePairOnePolygonTwoRings3`

Test case is a box with box hole inside. The three tests are 3 points in the hole, in the polygon area and outside the polygon.

I will add comments to the tests to make them easier to read.

cpp/include/cuspatial/detail/utility/offset_to_keys.cuh
isVoid: This functor was pulled from the `intersection_with_duplicates` file and contains notes that's out of context here. Removed.

cpp/include/cuspatial/experimental/detail/algorithm/is_point_in_polygon.cuh
isVoid: In all existing brute force point_in_polygon APIs.

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
isVoid: > Not sure about performance, though.
I'll write benchmark after this PR. I like readability of `any_of`.

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
isVoid: Probably no performance penalty, but may fail `compute-sanitizer --tools racecheck` though.

isVoid: Hmm, although it passes `compute-sanitizer --tool racecheck`, maybe there is no race condition?

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
isVoid: Needs benchmarking, but better redability.

cpp/include/cuspatial/experimental/point_polygon_distance.cuh
isVoid: This is meant to be `multipoint_range` and `multipolygon_range` (lower case), referring to the name of the templates.

cpp/include/cuspatial/experimental/ranges/multipoint_range.cuh
isVoid: `constexpr`

isVoid: Wait, this wasn't meant to be added since it's not used. There is `INVALID_INDEX` in `multipolygon_range`.

isVoid: Removed

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
harrism: If there is a race condition, it exists whether one thread or all threads write the location.

Also, shouldn't you be initializing the output to std::numeric_limits<T>::max() as in the min loop below?

Finally, why the `continue`? Doesn't the leading thread need to participate in the min reduction?

cpp/include/cuspatial/experimental/detail/point_polygon_distance.cuh
harrism: Agree. In this case, while a `reduce` signifies intent better, the loop is already pretty clear.

cpp/include/cuspatial/detail/utility/upper_bound_index.cuh
```diff
@@ -0,0 +1,58 @@
...
+ *
+ *  Example:
+ *  ```
+ *  offset:  0 0 0 1 3 4 4 4
+ *  i:       0 1 2 3
+ *  key:     3 4 4 5
+ *  ```
+ */
+template <typename Iterator>
+struct upper_bound_index_functor {
```
harrism, line 40: Hmmm, are you sure you can't use `make_geometry_id_iterator`? https://github.com/rapidsai/cuspatial/blob/main/cpp/include/cuspatial/experimental/iterator_factory.cuh#L369

harrism (approved):

thomcom (approved):

Issue #823, opened 2022-11-28
Avoid `vector<bool>`, use `thrust::host_vector<bool>` instead.


thomcom:
In tests where you compare against device data, just use a `thrust::device_vector` or `thrust::host_vector`.

_Originally posted by @harrism in https://github.com/rapidsai/cuspatial/pull/750#discussion_r1016004573_

PR #981, opened 2023-03-08
Use Quadtree, Pairwise, and Brute-Force PIP methods when appropriate with `.contains_properly` and other pip-only binary predicates.

thomcom:
## Description
This PR replaces #910.

#910 replaced the use of brute-force pip with Quadtree pip.

In contrast, this PR uses quadtree pip when appropriate (primarily via `mode="allpairs"` calls). It also uses pairwise pip when appropriate, and brute-force pip when appropriate. Most importantly, this PR also implements the elimination of the 31-polygon limit for brute-force pip. It is still limited by large lists of polygons since the resulting dataframe contains one row for each point and one column for each polygon, but builds the result from a single column. As a result the size of the input polygon parts * the number of points must be less than `2^32`.

## Checklist


harrism (changes requested):
Looks good, pending performance comparison of new and old brute-force pip for 31 polygons. I have some suggested improvements to the C++ code for generating the output table view.

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+#include <cudf/detail/null_mask.hpp>
+#include <cudf/table/table.hpp>
+#include <cudf/types.hpp>
```
isVoid, line 18: cudf includes should go below cuspatial includes.

cpp/src/spatial/point_in_polygon.cu
```diff
@@ -40,36 +43,38 @@ struct point_in_polygon_functor {
...
+    auto tid  = cudf::type_to_id<bool>();
     auto type = cudf::data_type{tid};
-    auto results =
-      cudf::make_fixed_width_column(type, size, cudf::mask_state::UNALLOCATED, stream, mr);
 
-    if (results->size() == 0) { return results; }
+    auto results = rmm::device_uvector<bool>(cudf::size_of(type) * size, stream, mr);
+    // cudf::make_fixed_width_column(type, size, cudf::mask_state::UNALLOCATED, stream, mr);
+
+    if (results.size() == 0) { return std::pair(make_empty_column(type), cudf::table_view()); }
```
isVoid, line 69: No need to initialize `results` if early return.

```suggestion
    if (size == 0) { return std::pair(make_empty_column(type), cudf::table_view()); }
    auto results = rmm::device_uvector<bool>(cudf::size_of(type) * size, stream, mr);
```

cpp/tests/experimental/spatial/point_in_polygon_test.cu
```diff
@@ -62,8 +62,8 @@ TYPED_TEST(PointInPolygonTest, OnePolygonOneRing)
   auto poly_point =
     this->make_device_points({{-1.0, -1.0}, {1.0, -1.0}, {1.0, 1.0}, {-1.0, 1.0}, {-1.0, -1.0}});
 
-  auto got      = rmm::device_vector<int32_t>(test_point.size());
-  auto expected = std::vector<int32_t>{false, false, false, false, true, true, true, true};
+  auto got      = rmm::device_vector<bool>(test_point.size() * poly_offsets.size());
+  auto expected = std::vector<bool>{false, false, false, false, true, true, true, true};
```
isVoid, line 66: `std::vector<bool>` *may be* a coalesced bit array (implementation defined). I'm not sure if `EXPECT_EQ` knows unpacking before comparing.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -168,40 +175,82 @@ def _op(self, lhs, points):
...
+            parts_map = cudf.Series(
+                cp.arange(len(part_sizes)), name="part_index"
+            ).repeat(part_sizes)
+            parts_df = parts_map.reset_index(drop=True).reset_index()
+            # Mapping of parts to polygons
+            parts_to_geoms = cp.array(self.lhs.polygons.geometry_offset)
+            geometry_sizes = parts_to_geoms[1:] - parts_to_geoms[:-1]
+            geometry_map = cudf.Series(
+                cp.arange(len(geometry_sizes)), name="polygon_index"
+            ).repeat(geometry_sizes)
```
isVoid, line 215: https://github.com/rapidsai/cuspatial/pull/910#discussion_r1113698735
```suggestion
            pcol = self.lhs.polygons
            offsets = pcol.ring_offset.take(pcol.part_offset).take(self.geometry_offset)
            sizes = offsets[1:] - offsets[:-1]
            geometry_map = self.Series(cp.arange(len(pcol.xy) // 2).repeat(sizes)
```

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -129,6 +134,10 @@ def __call__(self) -> cudf.Series:
 
 
 class ContainsProperlyBinpred(BinaryPredicate):
+    def __init__(self, lhs, rhs, align=True, mode="pairs"):
```
isVoid, line 137: Making sure @harrism 's question is not falling out of sync:

> Looking at geopandas APIs, I see a lot of named mode parameters rather than boolean. how is commonly used, e.g.
> 
> lanes_districts = gpd.sjoin(districts, bike_lane, how='inner', predicate='intersects')
> lanes_districts
> 
> So should we use how="allpairs" (and how="pairwise") instead of a Boolean parameter? Perhaps how might be needed also for other things, so maybe we need a new name for this, e.g. parallelism="allpairs" with options of "all_pairs", "pairwise", "one_to_all", etc.
> 
> Thoughts?

https://github.com/rapidsai/cuspatial/pull/910#discussion_r1113812489

python/cuspatial/cuspatial/core/binpreds/binpreds.py
```diff
@@ -211,15 +260,14 @@ def postprocess(self, point_indices, point_result):
...
-        point_result.name = None
-        return point_result
+        point_result["point_index"] = point_indices
+        hits = point_result.groupby("point_index").sum()
+        size = point_result.groupby("point_index").count()
+        x = hits != size
+        y = size > 0
+        z = hits > 0
+        group_result = x & y & z
+        return group_result
```
isVoid, line 270: Can you add comment to explain this? Difficult to understand without comments.

python/cuspatial/cuspatial/core/binpreds/contains.py
```diff
@@ -12,7 +15,69 @@
 from cuspatial.utils.column_utils import normalize_point_columns
 
 
-def contains_properly(
+def contains_properly_quadtree(
+    test_points,
+    polygons,
+):
```
isVoid, line 21: Suggest renaming to `quedtree_contains_properly` and `pairwise_conains_properly` for consistency.

```suggestion
def contains_properly_quadtree(
    test_points: GeoSeries,
    polygons: GeoSeries,
):
```

python/cuspatial/cuspatial/core/binpreds/contains.py
```diff
@@ -12,7 +15,69 @@
...
+    )
+    polygons_and_points["point_index"] = point_indices.iloc[
+        polygons_and_points["point_index"]
+    ].reset_index(drop=True)
+    polygons_and_points["part_index"] = polygons_and_points["polygon_index"]
+    polygons_and_points.drop("polygon_index", axis=1, inplace=True)
+    return polygons_and_points
+
+
+def contains_properly_pairwise(
```
isVoid, line 80: This wasn't covered in geoseries refactor... Do you want to convert it to geoseries input?

python/cuspatial/cuspatial/core/binpreds/contains.py
```diff
@@ -63,9 +127,10 @@ def contains_properly(
         as_column(poly_points_x),
         as_column(poly_points_y),
     )
+    poly_offsets = poly_offsets[:-1]
+    poly_ring_offsets = poly_ring_offsets[:-1]
```
isVoid, line 131: Is this still needed? with #973 merged the offsets are geoarrow compliant.

python/cuspatial/cuspatial/core/binpreds/contains.py
```diff
@@ -84,6 +149,10 @@ def contains_properly(
...
             poly_points_y,
         )
-
-    result = Series(pip_result, dtype="bool")
+    if isinstance(pip_result, NumericalColumn):
+        result = DataFrame(pip_result)
+    else:
+        result = DataFrame(
+            {k: v for k, v in enumerate(pip_result)}, dtype="bool"
+        )
```
isVoid, line 157: Can we just write separate returns in each of the `if` branches?

python/cuspatial/cuspatial/core/spatial/bounding.py
```diff
@@ -51,23 +51,21 @@ def polygon_bounding_boxes(polygons: GeoSeries):
     # by combining the geometry offset and parts offset of the multipolygon
     # array.
 
-    poly_offsets = polygons.polygons.part_offset.take(
-        polygons.polygons.geometry_offset
-    )
+    poly_offsets = polygons.polygons.part_offset
```
isVoid, line 54: I'm not sure if I understand why you need to revert this change. This should enable support for both polygon and multipolygon support regardless. Does the previous code break any existing code?

isVoid (changes requested):
I mainly have one question regarding a code revert with `polygon_bounding_boxes`. In GeoSeries refactor, I computed the new `part_offset` by gathering the `geometry_offset` and this treats the each multipolygon as simply a collection of points. In theory this should support polygon bounding boxes, as well as multipolygon bounding boxes. Did this break any usage of the API?

python/cuspatial/cuspatial/core/binpreds/contains.py
thomcom: Sure!

cpp/tests/experimental/spatial/point_in_polygon_test.cu
harrism: https://github.com/rapidsai/cuspatial/issues/823

harrism: Just change this to `thrust::host_vector<bool>` and it should work.

harrism:
> [Integrate](https://github.com/rapidsai/cuspatial/pull/981/commits/e6252289f40f304a7c806360f8edbc3e8ccf2707) @harrism['s validation improvements.](https://github.com/rapidsai/cuspatial/pull/981/commits/e6252289f40f304a7c806360f8edbc3e8ccf2707)

Not sure why you didn't just merge 23.04 into your branch instead of the above. The validation improvements are merged.

python/cuspatial/cuspatial/core/binpreds/binpreds.py
harrism: Thanks @isVoid

Issue #259, opened 2020-07-23
[FEA] geometry-generator utility for mocking up test data


cwharris:
The ability to create simple shapes such as rectangles, regular polygons, points, line segments, donuts, etc, would be useful when "developing" test data. Often times we know we need to generate polygons which overlap, or are separated by a minimum distance. Doing this by hand is non-optimal. Having a utility that generates vertices and indices representing various shapes with known attributes _at scale_ could prove beneficial. We could develop _simple and deterministic_ algorithms to generate this sort of data automatically at test time.

- Need a bunch of rectangles with an expected overlap "pattern"?
- What about some nested donuts with alternating rings?
- Points that are definitely inside/outside certain polygons?

Having utilities like these would make testing easier and more robust.

PR #998, opened 2023-03-17
Add `geometry_generator` factory for programmatic generation of geometry arrays

isVoid:
## Description

To allow easy benchmarking for future APIs, I created `geometry_generator` that programmatically generate a geoarrow compliant geometry arrays. The first feature landing is `generate_multipolygon_array`.

This PR also adds `BaseFixtureWithParam`, which allows user to parameterize the tests with value and ranges.

Contributes to #259 

## Checklist


cpp/include/cuspatial_test/base_fixture.hpp
```diff
@@ -22,15 +23,12 @@ namespace cuspatial {
 namespace test {
 
 /**
- * @brief Base test fixture class from which all libcuspatial tests should inherit.
- *
- * Example:
- * ```
- * class MyTestFixture : public cuspatial::test::BaseFixture {};
- * ```
+ * @brief Uses CRTP to supply rmm resources for test fixtures.
```
harrism, line 26: Why is the CRTP needed here? Where is the template parameter used?

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -0,0 +1,255 @@
...
+#include <rmm/device_uvector.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <thrust/sequence.h>
+#include <thrust/tabulate.h>
+
+namespace cuspatial {
+namespace test {
+
+constexpr double PI = 3.14159265358979323846;
```
harrism, line 34: Should not be necessary. Use `M_PI` from `<cmath>`. And in C++20 we get a new constant.

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -0,0 +1,255 @@
...
+ * @tparam T Type of the coordinates
+ */
+template <typename T>
+struct multipolygon_generator_parameter {
+  using element_t = T;
+
+  std::size_t num_multipolygons;
+  std::size_t num_polygons_per_multipolygon;
+  std::size_t num_holes_per_polygon;
+  std::size_t num_sides_per_ring;
```
harrism, line 48:
```suggestion
  std::size_t num_edges_per_ring;
```

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -0,0 +1,255 @@
...
+ * @tparam T Type of coordinate
+ * @param point_local_idx Local point of the point
+ * @param num_sides Number of sides of the polygon
+ * @param centroid Centroid of the ring
+ * @param radius Radius of the ring
+ * @return Coordinate of the point
+ */
+template <typename T>
+vec_2d<T> __device__ generate_ring_coordinate(std::size_t point_local_idx,
+                                              std::size_t num_sides,
```
harrism, line 78:
```suggestion
                                              std::size_t num_edges,
```

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -0,0 +1,255 @@
...
+};
+
+/**
+ * @brief Generate coordinates for the ring based on the local index of the point.
+ *
+ * The ring is generated by walking a point around a centroid with a fixed radius.
+ * Each step has equal angles.
+ *
+ * @tparam T Type of coordinate
+ * @param point_local_idx Local point of the point
```
harrism, line 70: Please clarify. What does "Local point of the point" mean?

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -0,0 +1,255 @@
...
+template <typename T>
+vec_2d<T> __device__ generate_ring_coordinate(std::size_t point_local_idx,
+                                              std::size_t num_sides,
+                                              vec_2d<T> centroid,
+                                              T radius)
+{
+  // Overrides last coordinate to make sure ring is closed.
+  if (point_local_idx == num_sides) return vec_2d<T>{centroid.x + radius, centroid.y};
+
+  T angle = 2.0 * PI * point_local_idx / num_sides;
```
harrism, line 85: Is there an integer division here? I guess not due to order of operations, but perhaps use grouping to make it explicit: 
```suggestion
  T angle = (2.0 * PI * point_local_idx) / num_sides;
```

cpp/tests/utility_test/test_float_equivalent.cu
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
```
harrism, line 2:
```suggestion
 * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```

harrism (approved):
Minor comments. Thanks for this utility!

cpp/include/cuspatial_test/base_fixture.hpp
isVoid: Functionally making this class a base class for both test fixtures would also work. Making it as a mixin class is because there may be other classes that can also use this mixin to provide RMM resources.

cpp/include/cuspatial_test/geometry_generator.cuh
isVoid: `local index of the point`

cpp/include/cuspatial_test/geometry_generator.cuh
isVoid: There shouldn't be. But parens are welcome.

cpp/include/cuspatial_test/base_fixture.hpp
harrism: I understand why it should be a mixin. I'm asking why is CRTP needed in order to make it a mixin? It doesn't use anything from the template parameter?

isVoid: Perhaps I misunderstood the use of CRTP in this case. If it doesn't use the template parameter, then there isn't a need for it.

PR #1002, opened 2023-03-20
Add `pairwise_point_polygon_distance` benchmark

isVoid:
## Description

This PR adds pairwise point polygon distance benchmark. Depends on #998

## Checklist


thomcom (approved):

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -322,8 +333,20 @@ auto make_multipoints_array(std::initializer_list<std::initializer_list<vec_2d<T
...
+  return multipoint_array<rmm::device_vector<std::size_t>, rmm::device_vector<vec_2d<T>>>{
+    rmm::device_vector<std::size_t>(offsets), rmm::device_vector<vec_2d<T>>(coordinates)};
+}
+
+/**
+ * @brief Factory method to construct multipoint array by moving the offsets and coordinates from
+ * `rmm::device_uvector`.
+ */
+template <typename IndexType, typename T>
+auto make_multipoints_array(rmm::device_uvector<IndexType> geometry_offsets,
```
harrism, line 345:
```suggestion
auto make_multipoint_array(rmm::device_uvector<IndexType> geometry_offsets,
```

harrism (approved):

isVoid:
This PR require adding nvtx dependency to conda environment.

cpp/include/cuspatial/detail/nvtx/ranges.hpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023, NVIDIA CORPORATION.
```
harrism, line 2: I don't think you need to change this.

cpp/CMakeLists.txt
```diff
@@ -198,6 +210,7 @@ target_compile_definitions(cuspatial PUBLIC "SPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_${
 
 # Specify the target module library dependencies
 target_link_libraries(cuspatial PUBLIC cudf::cudf)
+target_link_libraries(cuspatial PRIVATE nvtx3-cpp)
```
robertmaynard, line 213: This is a `PUBLIC` dependency not a `PRIVATE` one. The nvtx3 headers are included by the cuspatial public headers and therefore consumers of cuspatial will also need nvtx3.

This will resolve the test compilation failures.

robertmaynard (changes requested):

cpp/CMakeLists.txt
bdice: We expected to need PUBLIC here and tried it but it caused another issue. @isVoid Can you provide the logs?

robertmaynard: The nvtx3 project doesn't have have install / export rules for the targets they create it looks like. So you would get errors when trying to export the cuspatial targets as they now depend on targets that have no way to be exported.

What we should do is specify `DOWNLOAD_ONLY` when getting nvtx3 and directly including `nvtxImportedTargets.cmake` ( https://github.com/NVIDIA/NVTX/blob/v3.1.0-c-cpp/CMakeLists.txt#L33 ). 

We will need to also setup install rules for this file and the nvtx3 headers, plus a custom `FINAL_CODE_BLOCK` I expect for the `rapids_export`.

I expect that this is why previously cuspatial copied nvtx3 directly, as it dramatically simplified the build-system.

harrism: > Should we look into contributing to nvtx3 to provide install / export rules?

Yes but don't hold your breath for it to be put in a release anytime soon.  It took them about 2 years to release the last merged feature I was waiting on.

harrism:
@isVoid is this ready to go into 23.06?

bdice:
This PR successfully implemented NVTX from GitHub for cudf: https://github.com/rapidsai/cudf/pull/15178

We can probably copy from it and start a new attempt at this for cuSpatial if desired.

PR #1030, opened 2023-04-03
Column API for linestring-polygon distance

### LINKED ISSUE #1028: Column API for pairwise linestring-polygon distance

isVoid:
## Description

This PR adds linestring-polygon column API.

closes #1028

## Checklist


cpp/include/cuspatial/distance/linestring_polygon_distance.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * @ingroup distance
+ * @brief Compute pairwise (multi)linestring-to-(multi)polygon Cartesian distance
+ *
+ * @param multilinestrings Geometry column of multilinestrings
+ * @param multipolygons Geometry column of multipolygons
+ * @param mr Device memory resource used to allocate the returned column.
+ * @return Column of distances between each pair of input geometries, same type as input coordinate
+ * types.
+ *
+ * @throw cuspatial::logic_error if `multilinestrings` and `multipolygons` has different coordinate
```
harrism, line 37:
```suggestion
 * @throw cuspatial::logic_error if `multilinestrings` and `multipolygons` have different coordinate
```

cpp/tests/spatial/linestring_polygon_distance_test.cpp
```diff
@@ -0,0 +1,230 @@
...
+  CUSPATIAL_RUN_TEST(this->run_single, this->empty_multilinestring(), this->empty_polygon(), {});
+};
+
+TYPED_TEST(PairwiseLinestringPolygonDistanceTest, MultiToMultiEmpty)
+{
+  CUSPATIAL_RUN_TEST(
+    this->run_single, this->empty_multilinestring(), this->empty_multipolygon(), {});
+};
+
+TYPED_TEST(PairwiseLinestringPolygonDistanceTest, SingleToSingleOnePair)
```
harrism, line 122: Generally we test real data cases in the header-only API tests only. To avoid redundant tests, the column-based API tests just test error cases and empty input cases.

If there are data cases tested here that are not in the header-only API tests, they should probably move there and any duplicate data tests removed from here.

harrism (changes requested):
Just one comment on testing.

cpp/tests/spatial/linestring_polygon_distance_test.cpp
isVoid: These are duplicates. Probably remnants from my tests regarding some confounding bug I found in Python, attempting to trace all the way down.

isVoid: Removed.

harrism (approved):

PR #1044, opened 2023-04-07
cuSpatial: Build CUDA 12 packages

### LINKED ISSUE #925: cuSpatial: CUDA 12 Conda Packages
Pushing for CUDA 12 Conda package builds of libraries for RAPIDS 23.04. This issue is for tracking that work.

Requirements (please recheck):

```[tasklist]
### Developer tasks
- [x] I've confirmed the list below is correct (or updated if not)
```

```[tasklist]
### Toolkit dependencies from Conda
- [ ] https://github.com/conda-forge/staged-recipes/pull/21723
- [ ] https://github.com/conda-forge/staged-recipes/pull/21350
- [ ] https://github.com/conda-forge/staged-recipes/pull/22355
```

```[tasklist]
### RAPIDS Conda dependencies
- [ ] https://github.com/rapidsai/rmm/issues/1207
- [ ] https://github.com/rapidsai/cudf/issues/12771
```

```[tasklist]
### Other dependencies from Conda using CUDA
- [ ] https://github.com/conda-forge/numba-feedstock/issues/112
```

isVoid:
## Description
This PR builds `libcuspatial` and `cuspatial` conda packages using CUDA 12. Closes #925.

Based on https://github.com/rapidsai/cudf/pull/12922.

## Checklist


bdice (commented):
Good start. Let’s minimize the dependencies. I don’t think cuSpatial requires all the same things as libcudf. I will revisit the math libraries question in a bit.

ajschmidt8:
Please use GitHub's _Draft PR_ feature in the future. Draft PRs have the benefit of preventing notifications to `codeowners` until PRs are marked _Ready for Review_, which helps cut down on excessive notifications for PRs that are still being worked on. CI will still run on Draft PRs.

Some useful information about Draft PRs:

- https://github.blog/2019-02-14-introducing-draft-pull-requests/
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

ajschmidt8:
Marking PR as draft.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -33,13 +33,32 @@ build:
...
-    - {{ compiler('cuda') }} {{ cuda_version }}
     - {{ compiler('cxx') }}
+    {% if cuda_major == "11" %}
+    - {{ compiler('cuda11') }} {{ cuda_version }}
+    {% else %}
+    - {{ compiler('cuda') }} {{ cuda_version }}
+    - cuda-cudart ={{ cuda_version }}
+    - cuda-cudart-dev ={{ cuda_version }}
+    - cuda-cudart-static ={{ cuda_version }}
+    - cuda-cupti ={{ cuda_version }}
```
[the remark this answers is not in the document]
isVoid: cuspatial uses NVBench, I'll add libcupti.

bdice:
> I will revisit the math libraries question in a bit.

The math library dependency I was thinking of was cusparse, which was removed in this PR: https://github.com/rapidsai/cuspatial/pull/959. No math libraries are needed here, which simplifies things. 😄

conda/recipes/libcuspatial/meta.yaml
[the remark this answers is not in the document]
isVoid:
```
CMake Error at build/_deps/nvbench-src/nvbench/CMakeLists.txt:75 (target_link_libraries):
  Target "nvbench" links to:

    CUDA::cuda_driver
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -56,13 +72,23 @@ outputs:
...
         - ninja
       run:
+        {% if cuda_major == "11" %}
         - cudatoolkit {{ cuda_spec }}
+        {% else %}
+        - cuda-cudart {{ cuda_spec }}
+        {% endif %}
+        - librmm ={{ minor_version }}*=cuda{{ cuda_major }}*
         - libcudf ={{ minor_version }}
         - librmm ={{ minor_version }}
```
bdice:
```suggestion
        - libcudf ={{ minor_version }}*=cuda{{ cuda_major }}*
        - librmm ={{ minor_version }}*=cuda{{ cuda_major }}*
```

conda/recipes/libcuspatial/meta.yaml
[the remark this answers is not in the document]
bdice: Yup. Add the `cuda-driver-dev` to fix that. I forgot where that was used.

conda/recipes/libcuspatial/meta.yaml
bdice: Resolved this myself in 01b5c6b since there wasn't a diff nearby that I could use a suggestion to fix. :)

dependencies.yaml
```diff
@@ -110,6 +110,16 @@ dependencies:
...
+          - matrix:
+              arch: x86_64
+              cuda: "12.0"
+            packages:
+              - cuda-nvcc_linux-64=12.0
+          - matrix:
+              arch: aarch64
+              cuda: "12.0"
+            packages:
+              - cuda-nvcc_linux-aarch64=12.0
```
isVoid, line 122: Can this be simplified to 
```
- matrix
    cuda: "12.0"
  packages:
    - cuda-nvcc
```
?

bdice: Yup! Refer to rmm/cudf for an example.

ci/build_python.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
bdice, line 2: This can be reverted.

ci/test_cpp.sh
```diff
@@ -18,6 +18,10 @@ set +u
 conda activate test
 set -u
 
+RAPIDS_CUDA_MAJOR="${RAPIDS_CUDA_VERSION%%.*}"
```
bdice, line 21: The changes in this file can be reverted.

conda/recipes/libcuspatial/build.sh
```diff
@@ -1,4 +1,6 @@
 # Copyright (c) 2018-2022, NVIDIA CORPORATION.
 
+export NVCC_PREPEND_FLAGS="${NVCC_PREPEND_FLAGS} -ccbin ${CXX}" # Needed for CUDA 12 nvidia channel compilers
```
bdice: The changes in this file can be reverted.

conda/recipes/libcuspatial/meta.yaml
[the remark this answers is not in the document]
bdice: I recommend looking at my recent cudf PR where I revised the benchmarks — we decided to disable CUPTI support in nvbench. Then you can drop some dependencies here, probably cuda-cupti and maybe the driver dev package too. Building with CUPTI means that the benchmark executables link to a CUDA 11.8 specific library, which breaks minor version compatibility.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,18 +34,35 @@ build:
...
   build:
     - {{ compiler('c') }}
-    - {{ compiler('cuda') }} {{ cuda_version }}
     - {{ compiler('cxx') }}
+    {% if cuda_major == "11" %}
+    - {{ compiler('cuda11') }} {{ cuda_version }}
+    {% else %}
+    - {{ compiler('cuda') }} {{ cuda_version }}
+    - cuda-cudart ={{ cuda_version }}
+    - cuda-cudart-dev ={{ cuda_version }}
```
bdice: Delete all the cuda-cudart* packages from build and add **only** cuda-cudart-dev to host.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,18 +34,35 @@ build:
...
+    - cuda-cudart-static ={{ cuda_version }}
+    - cuda-cupti ={{ cuda_version }}
+    - cuda-driver-dev {{ cuda_driver_version }}
+    {% endif %}
     - doxygen
     - gmock {{ gtest_version }}
     - gtest {{ gtest_version }}
-    - libcudf ={{ minor_version }}
-    - librmm ={{ minor_version }}
+    - libcudf ={{ minor_version }}*=cuda{{ cuda_major }}*
```
bdice, line 64: Revert these. We won’t pin on build strings, we are using the cuda-version dependency to constrain which packages work together.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -57,15 +74,24 @@ outputs:
       run_exports:
         - {{ pin_subpackage("libcuspatial", max_pin="x.x") }}
       ignore_run_exports_from:
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
```
bdice, line 78: You only need this line. The CUDA 12 compiler doesn’t have anything you need to ignore, and you can drop cupti by passing a CMake flag to disable nvbench cupti support as mentioned above.

dependencies.yaml
```diff
@@ -175,6 +185,15 @@ dependencies:
               cuda: "11.2"
             packages:
               - cudatoolkit=11.2
+          - matrix:
+              cuda: "12.0"
```
bdice, line 184: Put this list at the top so it is sorted from 12.0 to 11.2. Then drop all but the -dev packages. Don’t specify their versions, just pin cuda-version=12.0 in that list.

bdice (commented):
Thanks for working on this @isVoid! I have a few comments that might help.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,18 +34,35 @@ build:
...
+    - cuda-driver-dev
+    {% endif %}
+    - cuda-version ={{ cuda_version }}
     - doxygen
     - gmock {{ gtest_version }}
     - gtest {{ gtest_version }}
-    - libcudf ={{ minor_version }}
-    - librmm ={{ minor_version }}
+    - libcudf ={{ minor_version }}*=cuda{{ cuda_major }}*
+    - librmm ={{ minor_version }}*=cuda{{ cuda_major }}*
```
isVoid, line 65: Should not pin to specific cuda version.

conda/recipes/libcuspatial/meta.yaml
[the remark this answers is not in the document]
isVoid: Can you give me a link to that PR? Thanks :)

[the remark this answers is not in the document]
bdice: https://github.com/rapidsai/cudf/blob/0688872200d0a0c70d5e4d1c17e2e29ecd00af8f/conda/recipes/libcudf/build.sh#L9

https://github.com/rapidsai/cudf/pull/13528

isVoid:
This cpp build failure seems strange.. It's supposed to be fixed by a patch from cudf to thrust:

https://github.com/rapidsai/cudf/blob/c7e9405f939d9750c6a6321fbd66c14674951042/cpp/cmake/thirdparty/patches/thrust_transform_iter_with_reduce_by_key.diff#L22C23-L22C23

However it doesn't seem like it's applied for cuda 12.0 build (but applied for 11.x builds).

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,18 +34,29 @@ build:
...
```
bdice, line 59:
```suggestion
    - libcudf ={{ minor_version }}
    - librmm ={{ minor_version }}
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,13 +34,24 @@ build:
...
+    - cuda-version ={{ cuda_version }}
     - cmake {{ cmake_version }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
   host:
+    {% if cuda_major == "11" %}
     - cudatoolkit ={{ cuda_version }}
+    {% else %}
+    - cuda-cudart-dev
+    - cuda-driver-dev
```
bdice, line 52: I think we can drop this now that we don't use cupti in nvbench?

cpp/CMakeLists.txt
```diff
@@ -109,6 +109,10 @@ include(cmake/Modules/ConfigureCUDA.cmake)
 
 # add third party dependencies using CPM
 rapids_cpm_init()
+# find rapids-cmake configured thrust, not from system CCCL
+include(cmake/thirdparty/get_thrust.cmake)
+# find rapids-cmake configured libcudacxx, not from system CCCL
+include(cmake/thirdparty/get_libcudacxx.cmake)
```
bdice, line 115:
```suggestion
# find rapids-cmake configured libcudacxx, not from system CCCL
include(cmake/thirdparty/get_libcudacxx.cmake)
# find rapids-cmake configured thrust, not from system CCCL
include(cmake/thirdparty/get_thrust.cmake)
```

dependencies.yaml
```diff
@@ -159,6 +164,12 @@ dependencies:
...
         matrices:
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cuda-version=12.0
+              - cuda-cudart-dev
+              - cuda-nvrtc-dev
           - matrix:
               cuda: "11.8"
             packages:
```
bdice, line 175: The package lists for CUDA 11 versions (below) should look like:
```
- cuda-version=11.8
- cudatoolkit
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,13 +34,24 @@ build:
...
+    {% endif %}
```
bdice, line 53: Try this:
```suggestion
    {% endif %}
```

conda/recipes/libcuspatial/conda_build_config.yaml
```diff
@@ -15,3 +18,6 @@ gtest_version:
 
 sysroot_version:
   - "2.17"
+
+cuda_driver_version:
+  - "12.0.146"
```
bdice, line 23:
```suggestion
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -34,13 +34,21 @@ build:
...
+    {% else %}
+    - {{ compiler('cuda') }}
+    {% endif %}
+    - cuda-version ={{ cuda_version }}
     - cmake {{ cmake_version }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
   host:
+    {% if cuda_major == "11" %}
     - cudatoolkit ={{ cuda_version }}
```
bdice, line 49:
```suggestion
    - cudatoolkit
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -57,13 +65,20 @@ outputs:
...
       build:
         - cmake {{ cmake_version }}
         - ninja
       run:
+        {% if cuda_major == "11" %}
         - cudatoolkit {{ cuda_spec }}
+        {% else %}
+        - cuda-cudart
+        {% endif %}
+        - cuda-version ={{ cuda_version }}
```
bdice, line 81:
```suggestion
      host:
        - cuda-version ={{ cuda_version }}
      run:
        {% if cuda_major == "11" %}
        - cudatoolkit
        {% else %}
        - cuda-cudart
        {% endif %}
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -82,12 +97,19 @@ outputs:
...
       build:
         - cmake {{ cmake_version }}
       run:
         - {{ pin_subpackage('libcuspatial', exact=True) }}
+        {% if cuda_major == "11" %}
         - cudatoolkit {{ cuda_spec }}
+        {% else %}
+        - cuda-cudart
+        {% endif %}
+        - cuda-version ={{ cuda_version }}
```
bdice, line 113:
```suggestion
      host:
        - cuda-version ={{ cuda_version }}
      run:
        - {{ pin_subpackage('libcuspatial', exact=True) }}
        {% if cuda_major == "11" %}
        - cudatoolkit
        {% else %}
        - cuda-cudart
        {% endif %}
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

bdice:
CI is passing so I'm going to open this draft for review. I'll also start a review. 👍

bdice (commented):
I pushed a small change to leverage `run_exports` by putting `-dev` packages in the `host` section, so that they populate the `run` section implicitly. We adopted this approach in cudf because it plays better with the conda overlinking checker. Otherwise this looks good to me! I'd like a review from one other CUDA 12 expert before approving.

cpp/cmake/thirdparty/get_libcudacxx.cmake
```diff
@@ -0,0 +1,37 @@
...
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+# This function finds thrust and sets any additional necessary environment variables.
+function(find_and_configure_libcudacxx)
+  # Make sure we install libcudacxx beside our patched version of thrust
+  include(GNUInstallDirs)
+  set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/libcuspatial")
+  set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_INCLUDEDIR}/lib")
```
vyasr, line 20: Hmm I realize that this is copied directly from libcudf, but something seems off here. Why are we setting the libdir to be inside the include dir? If that's to have the side-by-side behavior, then I suspect what we really want is perhaps `"${CMAKE_INSTALL_INCLUDEDIR}/libcuspatial" too. @robertmaynard could you verify this?

I suspect we're getting away with an error here because libcudacxx is header-only so there aren't any libraries to install in practice. The only other possibility is that this lib dir setting is correct and necessary in order for the CMake config files to be installed in the normal search directories (lib/cmake), but that seems like a fragile approach if libcudacxx does ever start shipping libraries.

If my diagnosis is right we may need to port this fix back to libcudf too.

Also, if libcuspatial is doing the exact same patching as libcudf perhaps we could reuse libcudf's copy instead of generating yet another one? But I can see the possibility for conflict there so I'm OK with this approach too.

cpp/cmake/thirdparty/get_thrust.cmake
```diff
@@ -0,0 +1,45 @@
...
+  include(${rapids-cmake-dir}/cpm/thrust.cmake)
+  include(${rapids-cmake-dir}/cpm/package_override.cmake)
+
+  set(cuspatial_patch_dir "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/patches")
+  rapids_cpm_package_override("${cuspatial_patch_dir}/thrust_override.json")
+
+  # Make sure we install thrust into the `include/libcuspatial` subdirectory instead of the default
+  include(GNUInstallDirs)
+  set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/libcuspatial")
+  set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_INCLUDEDIR}/lib")
```
vyasr, line 26: Same question as for libcudacxx.

dependencies.yaml
vyasr, line 1: Should also update the notebooks, build-python, and test-python in this PR, even if they won't have any effect until some upstream work is completed.

vyasr (changes requested):
This is fairly close, but a couple of things need resolving and perhaps a bit of discussion.

cpp/cmake/thirdparty/get_libcudacxx.cmake
robertmaynard: > Why are we setting the libdir to be inside the include dir? I

Cudf does so since it is a patched version of the CCCL libraries and therefore shouldn't be picked up by any abitrary `find_package(Thrust)` since cudf disables some thrust features that other libraries could need.

Second, since we are constrained on the `CMAKE_INSTALL_INCLUDEDIR` and `CMAKE_INSTALL_LIBDIR` locations since the CCCL find packages search for headers and pre 2.2 had issues when the `include` and `lib` didn't share the same root ( `CMAKE_INSTALL_INCLUDEDIR` in this case ).

conda/recipes/cuspatial/meta.yaml
```diff
@@ -2,8 +2,9 @@
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') %}
 {% set minor_version = version.split('.')[0] + '.' + version.split('.')[1] %}
-{% set py_version = environ['CONDA_PY'] %}
 {% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
+{% set cuda_major = cuda_version.split('.')[0] %}
+{% set py_version = environ['CONDA_PY'] %}
```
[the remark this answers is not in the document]
bdice: It's used in the build string.
```
string: py{{ py_version }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```

cpp/cmake/thirdparty/get_libcudacxx.cmake
bdice: @isVoid We should double-check if we get libcudf's patched Thrust if we remove the changes for fetching/patching Thrust here, now that the cuda-nvcc include flag order has been fixed.

conda/recipes/cuspatial/meta.yaml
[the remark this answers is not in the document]
bdice: I fixed this in d2a20f927db9b4f646d485b189f7243a3e8a438b. Leaving the rest to @isVoid.

conda/recipes/cuspatial/meta.yaml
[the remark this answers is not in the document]
vyasr: Oh I didn't notice that this definition was just moved, I thought it was something new. Never mind then, carry on.

cpp/cmake/thirdparty/get_libcudacxx.cmake
vyasr: I get that we're trying to avoid normal `find_package` calls from finding the library, but I guess I would have expected us to want to place the libraries in some subdirectory of the lib dir (with a different name to subvert normal `find_package(Thrust)`) rather than putting libraries into the _include_ dir. Using a top-level `$INCLUDEDIR/lib` also seems fairly clobber-friendly, wouldn't it be better to place it inside a `rapids` or `cuspatial` directory?

isVoid: > now that the cuda-nvcc include flag order has been fixed.

You mean using -I in nvcc compiler instead of -isystem? Can you post the link to this fix for track record?

bdice: It moving the implicit include of -I$PREFIX/targets/… to the end rather than the beginning of the search path. No changes in -I vs. -isystem. https://github.com/conda-forge/cuda-nvcc-feedstock/pull/22

dependencies.yaml
isVoid: Can you hint what should be updated?

bdice: There are packages in those lists for pyproject/requirements that have -cu11 in the name, like cudf-cu11. You need a new matrix entry for CUDA 12 that uses cudf-cu12, etc.

conda/recipes/cuspatial/meta.yaml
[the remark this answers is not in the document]
isVoid: Looks like this is a one-off oversight. All other `ignore_run_exports_from` (in libcuspatial) have the cuda11 conditional.

dependencies.yaml
isVoid: For `test_python` there's no cuda11 dependency, should we just add a matrix with just the metapackage?

bdice: Yup. `build_python`, `notebooks`, and `run_python` are the ones that need updates here.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -86,12 +106,24 @@ outputs:
       number: {{ GIT_DESCRIBE_NUMBER }}
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
       ignore_run_exports_from:
-        - {{ compiler('cuda') }}
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
     requirements:
       build:
```
jakirkham, line 113: Should we have the compilers here?

```suggestion
      build:
        - {{ compiler('c') }}
        - {{ compiler('cxx') }}
        {% if cuda_major == "11" %}
        - {{ compiler('cuda11') }} ={{ cuda_version }}
        {% else %}
        - {{ compiler('cuda') }}
        {% endif %}
        - sysroot_{{ target_platform }} {{ sysroot_version }}
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -86,12 +106,24 @@ outputs:
...
         - cmake {{ cmake_version }}
+      host:
+        - cuda-version ={{ cuda_version }}
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
```
jakirkham, line 121: If we add the `compiler`s above, the `cudatoolkit` portion will be unneeded

```suggestion
        {% if cuda_major != "11" %}
        - cuda-cudart-dev
        {% endif %}
```

conda/recipes/cuspatial/meta.yaml
```diff
@@ -33,16 +34,27 @@ build:
...
+    {% else %}
+    - {{ compiler('cuda') }}
+    {% endif %}
+    - cuda-version ={{ cuda_version }}
     - ninja
     - sysroot_{{ target_platform }} {{ sysroot_version }}
   host:
+    {% if cuda_major == "11" %}
+    - cudatoolkit
+    {% endif %}
```
jakirkham, line 56: This is added by `{{ compiler('cuda11') }}` in the CUDA 11 case automatically. So no need to include it here

```suggestion
```

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -86,12 +106,24 @@ outputs:
       number: {{ GIT_DESCRIBE_NUMBER }}
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
       ignore_run_exports_from:
-        - {{ compiler('cuda') }}
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
     requirements:
       build:
         - cmake {{ cmake_version }}
```
jakirkham, line 114: Do we need `ninja` here too (as it is used elsewhere)?

```suggestion
        - cmake {{ cmake_version }}
        - ninja
```

jakirkham (commented):
Thanks Michael and everyone reviewing! 🙏

Had a couple suggestions below

cpp/cmake/thirdparty/get_libcudacxx.cmake
isVoid: > We should double-check if we get libcudf's patched Thrust if we remove the changes for fetching/patching Thrust here, now that the cuda-nvcc include flag order has been fixed.

It seems removing the patched thrust now no longer causes the issue.

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -59,13 +67,25 @@ outputs:
       run_exports:
         - {{ pin_subpackage("libcuspatial", max_pin="x.x") }}
       ignore_run_exports_from:
-        - {{ compiler('cuda') }}
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
```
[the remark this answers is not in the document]
isVoid: Probably unneeded. Removed and trying if CI passes.

conda/recipes/libcuspatial/meta.yaml
bdice: No, we do not need ninja here. Ninja should be in the top-level `build` but not the subpackages' `build` sections.

conda/recipes/libcuspatial/meta.yaml
bdice: No, we have already finished compilation by the time we're packaging libcuspatial and libcuspatial-tests. Compilers are needed in the top-level `build` section for use by `build.sh`, but the individual package outputs only install (copy) files to the conda environment and do not perform compilation.

The only argument I could see for adding compilers would be to influence the host/run environments via `run_exports` but I think that is necessary here (we haven't traditionally done this).

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -59,13 +67,25 @@ outputs:
...
     requirements:
       build:
```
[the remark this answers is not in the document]
bdice: Resolving, see conversation here: https://github.com/rapidsai/cuspatial/pull/1044#discussion_r1248296401

conda/recipes/libcuspatial/meta.yaml
```diff
@@ -59,13 +67,25 @@ outputs:
...
         - cmake {{ cmake_version }}
         - ninja
+      host:
+        - cuda-version ={{ cuda_version }}
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
```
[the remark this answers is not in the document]
bdice: Resolving, see https://github.com/rapidsai/cuspatial/pull/1044#discussion_r1248296401

conda/recipes/libcuspatial/meta.yaml
bdice: Resolving, see https://github.com/rapidsai/cuspatial/pull/1044#discussion_r1248296401

bdice (approved):
All looks good to me. Thanks @isVoid!

cpp/cmake/thirdparty/get_libcudacxx.cmake
vyasr: As long as this is working, given that it's copied from libcudf I'm fine approving here and following up with Robert separately. Addressing my concerns here is not on the critical path for CUDA 12.

vyasr (approved):
LGTM

raydouglass (approved):

jakirkham (approved):

PR #1048, opened 2023-04-07
Use thrust::host_vector instead of std::vector<bool> in tests

### LINKED ISSUE #823: Avoid `vector<bool>`, use `thrust::host_vector<bool>` instead.

cwharris:
## Description
Updates tests to use `thrust::host_vector<bool>` instead of `std::vector<bool>` and in one case `std::vector<int>` where int was logically bool.

closes #823

## Checklist


cwharris:
Another way we could do this is by creating a `host_vector_converter` class that has an implicit `initializer_list` constructor and an implicit `thrust::host_vector` conversion operator, eliminating the need to call a function.

isVoid (approved):

isVoid:
> Another way we could do this is by creating a host_vector_converter class that has an implicit initializer_list constructor and an implicit thrust::host_vector conversion operator, eliminating the need to call a function.

I'm curious how this is done. Wouldn't you still need to call `host_vector_converter` ctor somewhere to resolve to the conversion operator?

cwharris:
@isVoid Yes, you're right. We would have to modify the function signatures or explicitly call the converter constructor. I didn't think this through properly.

Issue #1036, opened 2023-04-05
Document all simple and single geometric special cases


thomcom:
Closed by #1064 

This task is for documenting and planning special cases that need to be tested among binary predicates, such as tests with polygons with holes and linestring and polygon crossings.

There are many special cases of polygons that will need to have tests written for them. I'm writing a suite of special case tests to handle all possible configurations of geometric relations between two features.

See https://github.com/thomcom/spatial-predicates

#1064 implements 33 unique test cases for the three basic types: point, linestring, and polygon. The tests are run with each of the binary predicates and are used for developing each feature. 

A future issue and PR will track the creation and validation of multi-geometry feature predicates.
Another future issue will track the execution of any special cases that don't fall into the original 33, if any are discovered.

_Originally posted by @thomcom in https://github.com/rapidsai/cuspatial/pull/1016#discussion_r1157761892_

PR #1050, opened 2023-04-07
Binary Predicate Test Dispatch

thomcom:
## Description
Currently `contains_properly`, `intersection`, and `equals` based predicates are all tested in their own files and lack explicit corner cases and boundary conditions for certain tests. They also do not test unimplemented cases at all.

This PR creates `binpred_test_dispatch.py` and `test_binpred_test_dispatch.py`. These files are used to exhaustively test the space of binary predicates and dispatched feature pairs. This will allow easy implementation of the remaining binary predicates in a test-driven fashion. Finally, failing tests are automatically skipped for now, until the entire suite has been finished. At that point the skip functionality will be removed and fails will be treated as such.

Each test of `feature1.predicate(feature2)` is also dispatched internally to a group of custom values for each feature type. These specified values will be used in the follow up PR for https://github.com/rapidsai/cuspatial/issues/1036, enabling truly exhaustive testing of all possible combinations of features and feature coordinate arrangements.

When this test dispatcher and types in #1036 are finished, the pre-existing tests should be able to be deleted.

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
+++ b/python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
@@ -59,11 +59,11 @@ def geotype_tuple(request):
 @pytest.fixture(
     params=[
         "contains_properly",
+        "geom_equals",
+        "intersects",
         "covers",
         "crosses",
         "disjoint",
-        "geom_equals",
-        "intersects",
         "overlaps",
         "touches",
         "within",
--- a/python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
+++ b/python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
@@ -1,0 +1,225 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+import pytest
+from shapely.geometry import LineString, MultiPoint, Point, Polygon
+
+import cuspatial
+
+"""Test Dispatch"""
+
+"""Testing all combinations of possible geometry types and
+binary predicates is a complex task. This file contains the
+basic fixtures for all geometry types that are required to
+cover different possible test outcomes. This file also contains
+the dispatching system that uses each fixture to generate
+a test for each possible combination of geometry types and
+binary predicates.
+"""
+
+"""The following fixtures are used to generate tests for
+each possible combination of geometry types and binary
+predicates. Each fixture returns a tuple of the form
+(geotype, geotype, predicate, expected_result). The
+geotype fixtures are used to generate the first two
+elements of the tuple. The predicate fixture is used to
+generate the third element of the tuple. The expected_result
+fixture is used to generate the fourth element of the tuple.
+"""
+
+"""The collection of all possible geometry types"""
+
+
+@pytest.fixture(
+    params=[
+        (Point, Point),
+        (Point, MultiPoint),
+        (Point, LineString),
+        (Point, Polygon),
+        (MultiPoint, Point),
+        (MultiPoint, MultiPoint),
+        (MultiPoint, LineString),
+        (MultiPoint, Polygon),
+        (LineString, Point),
+        (LineString, MultiPoint),
+        (LineString, LineString),
+        (LineString, Polygon),
+        (Polygon, Point),
+        (Polygon, MultiPoint),
+        (Polygon, LineString),
+        (Polygon, Polygon),
+    ]
+)
+def geotype_tuple(request):
+    return request.param
+
+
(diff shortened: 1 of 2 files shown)


python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -0,0 +1,244 @@
...
+        "triple_center_disjoint",
+        "border_overlap",
+        "interior_overlap",
+    ]
+)
+def test_type(request):
+    return request.param
+
+
+"""Named features for each feature dispatch."""
```
isVoid, line 98: Can you give a bit more detail on what basis did you choose these test examples from? I expect what you chose should be sufficient to cover most of  the corner cases for various geometry combination and binpreds.

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -0,0 +1,244 @@
...
+points_dispatch = {
+    "feature1": Point(0.0, 0.0),
+    "feature2": Point(1.0, 1.0),
+    "feature1_bo": Point(1.0, 1.0),
+    "feature2_bo": Point(0.0, 0.0),
```
isVoid, line 103: There are duplicate values in the dict, can you explain why?

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -0,0 +1,244 @@
...
+
+"""The following fixtures are used to generate tests for
+each possible combination of geometry types and binary
+predicates. The fixtures are combined in the test function
+in `test_binpreds_test_dispatch.py` to make the following
+tuple: (predicate, geotype, geotype, expected_result). The
+geotype fixtures are used to generate the first two
+elements of the tuple. The predicate fixture is used to
+generate the third element of the tuple. The expected_result
+fixture is used to generate the fourth element of the tuple.
```
isVoid, line 27: The actual function signature is a different, it has only 3 arguments.

```suggestion
tuple: (predicate, geotype_tuple, expected_result). The
geotype fixtures are used to generate the second 
elements of the tuple. The predicate fixture is used to
generate the first element of the tuple. The expected_result
fixture is used to generate the third element of the tuple.
```

python/cuspatial/cuspatial/tests/binpreds/test_binpred_test_dispatch.py
```diff
@@ -0,0 +1,59 @@
...
+by the test function. Will be removed when all tests are passing."""
+
+
+def skip_on_exception(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            pytest.skip(f"Skipping due to an exception: {e}")
```
isVoid, line 24: `xfail` is probably more suitable here. If they are eventually meant to pass.

python/cuspatial/cuspatial/tests/binpreds/test_binpred_test_dispatch.py
```diff
@@ -0,0 +1,59 @@
...
+        pd.testing.assert_series_equal(expected, got.to_pandas())
+    except AssertionError as e:
+        print("Binary Predicate Test failed")
+        print("----------------------------")
+        print(f"lhs: {lhs}")
+        print(f"rhs: {rhs}")
+        print(f"predicate: {predicate}")
+        print(f"test_type: {test_type}")
+        print(f"expected: {expected}")
+        print(f"got: {got}")
```
isVoid, line 56: I was concerned if these prints will be recorded in CI. But probably not since pytest by default doesn't capture stdouts.

python/cuspatial/cuspatial/tests/binpreds/test_binpred_test_dispatch.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+from functools import wraps
+
+import pandas as pd
+import pytest
+from binpred_test_dispatch import (  # noqa: F401
```
isVoid, line 7: Why noqa here?

python/cuspatial/cuspatial/tests/binpreds/test_binpred_test_dispatch.py
isVoid: Even still, I'm hoping perhaps we should disable them during CI. Since they are still run and brings burden.

PR #1072, opened 2023-04-13
Update clang-format to 15.0.7

### LINKED ISSUE #1078: [FEA]: Update clang-format to a recent version

New Feature

Medium

RAPIDS is in the process of updating to a more recent clang-format version.  This is already in progress in #1072 .

This will help users of the cuspatial devcontainer because it has a newer clang-format version than is used in CI, and there are often style mismatches as a result.

We may use an earlier version than 16.0.1 due to a bug introduced in that version that results in `operator*` being formatted as if it is a pointer (`a* T{1.0}` instead of `a * T{1.0}`). See https://github.com/rapidsai/cugraph-ops/pull/466#discussion_r1168977855

_No response_

### Additional context

_No response_

bdice:
This PR updates the clang-format version used by pre-commit.  Fixes #1078.

### DIFF
--- a/cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
+++ b/cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
@@ -122,8 +122,8 @@ OutputIt trajectory_distances_and_speeds(IndexT num_trajectories,
                           auto time_s = static_cast<T>(time_d.count()) *
                                         static_cast<T>(Period::num) / static_cast<T>(Period::den);
                           T dist_km   = thrust::get<1>(a) + thrust::get<1>(b);
-                          T dist_m    = dist_km * T{1000.0};  // km to m
-                          T speed_m_s = dist_m / time_s;      // m/ms to m/s
+                          T dist_m    = dist_km* T{1000.0};  // km to m
+                          T speed_m_s = dist_m / time_s;     // m/ms to m/s
                           return thrust::make_tuple(time_d.count(), dist_km, dist_m, speed_m_s);
                         });
 
--- a/cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
+++ b/cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
@@ -122,8 +122,8 @@ OutputIt trajectory_distances_and_speeds(IndexT num_trajectories,
                           auto time_s = static_cast<T>(time_d.count()) *
                                         static_cast<T>(Period::num) / static_cast<T>(Period::den);
                           T dist_km   = thrust::get<1>(a) + thrust::get<1>(b);
-                          T dist_m    = dist_km* T{1000.0};  // km to m
-                          T speed_m_s = dist_m / time_s;     // m/ms to m/s
+                          T dist_m    = dist_km * T{1000.0};  // km to m
+                          T speed_m_s = dist_m / time_s;      // m/ms to m/s
                           return thrust::make_tuple(time_d.count(), dist_km, dist_m, speed_m_s);
                         });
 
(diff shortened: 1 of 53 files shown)


bdice:
@harrism I think this just needs reviews, then should be good to merge. (Merging sooner means less likelihood of conflicts.)

cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
```diff
@@ -122,8 +122,8 @@ OutputIt trajectory_distances_and_speeds(IndexT num_trajectories,
                           auto time_s = static_cast<T>(time_d.count()) *
                                         static_cast<T>(Period::num) / static_cast<T>(Period::den);
                           T dist_km   = thrust::get<1>(a) + thrust::get<1>(b);
-                          T dist_m    = dist_km * T{1000.0};  // km to m
-                          T speed_m_s = dist_m / time_s;      // m/ms to m/s
+                          T dist_m    = dist_km* T{1000.0};  // km to m
```
isVoid, line 125: This seems like a wierd binary op formatting?

bdice: I agree this is weird. Perhaps we need a different clang-format rule. Can you look into this?

bdice: I briefly searched through https://clang.llvm.org/docs/ClangFormatStyleOptions.html but didn't immediately see the option that would control this. I think this would be better spelled as `static_cast<T>(1000.0)`. What do you think of that option?

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -102,7 +102,7 @@ vec_2d<T> __device__ polygon_centroid_displacement(vec_2d<T> centroid,
                                                    std::size_t part_local_idx,
                                                    T radius)
 {
-  return centroid + vec_2d<T>{part_local_idx * radius * T{3.0}, T{0.0}};
+  return centroid + vec_2d<T>{part_local_idx * radius* T{3.0}, T{0.0}};
```
harrism, line 105: Another operator* mistaken for a pointer.

cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
bdice: Cross-referencing this thread: https://github.com/rapidsai/cugraph-ops/pull/466#discussion_r1168977855

cpp/tests/experimental/trajectory/trajectory_test_utils.cuh
```diff
@@ -274,8 +274,8 @@ struct trajectory_test_data {
       auto time_s =
         static_cast<T>(time_d.count()) * static_cast<T>(Period::num) / static_cast<T>(Period::den);
       T dist_km   = thrust::get<1>(a) + thrust::get<1>(b);
-      T dist_m    = dist_km * T{1000.0};  // km to m
-      T speed_m_s = dist_m / time_s;      // m/ms to m/s
+      T dist_m    = dist_km* T{1000.0};  // km to m
```
harrism, line 277: Another.

cpp/include/cuspatial/experimental/detail/trajectory_distances_and_speeds.cuh
harrism: I've triaged this and found that the bug does not exist in 15.0.7 or 17.0.0 (dev branch), but does exist in 16.0.0, 16.0.1, and 16.0.2 (the latter could change since it is not yet released). For now, I'm going to update this PR to 15.0.7.

harrism:
Updated to use clang-format 15.0.7 which doesn't have the operator* / brace initialization issue discovered in 16.0.1. We should follow up after 16.0.x is released with a fix.

bdice (commented):
Approving changes made by @harrism (cannot approve my own PR though).

thomcom (approved):

harrism:
\merge

harrism (approved):
Approving changes made by @bdice. :)

PR #1104, opened 2023-04-27
Create `CommonGeometryColumnFixture` for Common Geometry Column Reuse

isVoid:
## Description

This PR adds a `CommonGeometryColumnFixture` manages the life cycle of some commonly used geometry columns that can be shared across different test suites. Currently, only empty geometry columns are added.

## Checklist


harrism:
So some of this was done by #1097 but it looks like this also refactors some test fixture code. So perhaps the title of this PR should change?

Also, I realized that in #1097 I missed moving this file. Could you fix it in this PR? https://github.com/rapidsai/cuspatial/blob/branch-23.06/cpp/include/cuspatial/distance/polygon_distance.hpp  The function in that file should move into cpp/include/cuspatial/distance.hpp

isVoid:
Let's hold on reviewing this PR, thinking on a better design for the geometry fixtures.

isVoid:
Supersedes by #1124

PR #1124, opened 2023-05-04
Refactor ST_Distance Column API and Cython

### LINKED ISSUE #1123: Refactor ST_Distance column API to ensure API homogeneity.
Current `ST_Distance` column API lacks homogeneity and is confusing for user to use. More recent APIs (point_polygon_distance) defines inputs that make uses `geometry_column_view` that greatly reduces the complexity of the inputs of these APIs. The older APIs should conform to the standards that the new APIs sets.

isVoid:
## Description

This PR makes all ST_Distance API conforms to a homogenous API format and documentation. This also greatly simplifies the implementation of each of the column APIs. Closes #1123 

This PR also introduces several `GeometryColumnFixtures` that manages the life time of a few commonly used geometry columns and use them across the tests of these APIs. Supersedes #1104 

This PR also fixes several bugs in the computation kernels when the input is empty.

## Checklist


cpp/include/cuspatial/distance.hpp
```diff
@@ -127,118 +127,64 @@ std::pair<std::unique_ptr<cudf::column>, cudf::table_view> directed_hausdorff_di
   rmm::mr::device_memory_resource* mr = rmm::mr::get_current_device_resource());
 
 /**
- * @brief Compute pairwise (multi)point-to-(multi)point Cartesian distance
+ * @brief Compute pairwise (multi)point-to-(multi)point Euclidean distance
```
harrism, line 130: Why Euclidean and not Cartesian? AFAIU Cartesian is Euclidean + a coordinate system.

cpp/include/cuspatial_test/geometry_fixtures.hpp
```diff
@@ -0,0 +1,199 @@
...
+  std::unique_ptr<cudf::column> empty_point_column;
+  std::unique_ptr<cudf::column> empty_linestring_column;
+  std::unique_ptr<cudf::column> empty_polygon_column;
+  std::unique_ptr<cudf::column> empty_multipoint_column;
+  std::unique_ptr<cudf::column> empty_multilinestring_column;
+  std::unique_ptr<cudf::column> empty_multipolygon_column;
+};
+
+/**
+ * @brief Test Fixture that initializes one geometry columns.
```
harrism, line 101:
```suggestion
 * @brief Test Fixture that initializes one geometry column.
```

harrism (commented):
Very nice cleanup. The new test fixtures really DRY up the code.

harrism (approved):

cpp/include/cuspatial/distance.hpp
isVoid: Thanks. I accept that "Cartesian distance" is short for Euclidean distance on Cartesian CRS. The reason why I hesitated is that there are other distance definition on a cartesian plane, such as Manhatten distance. But since Euclidean is the most commonly used distance (at least in GIS terms), it should be safe to assume that.

harrism: Got it. Good point.

PR #1134, opened 2023-05-11
Add `multilinestring_segment_manager` for segment related methods in multilinestring ranges

### LINKED ISSUE #1055: [BUG]: Segment Iterator is flawed, may fail if input contains 0-length geometry
### Version

23.04

_No response_

### Describe the issue

Currently, `multilinestring_range::segment_begin()` require that the input doesn't include empty geometries. Such as:

```
geometry offset: 0, 1, 2
part offset: 0, 0, 2
Coords: (0, 0) (1, 1)
```

A complete solution require counting the number of skipped geometries prior to current linestring, and it would require tracking the result in an owning object, which is not what *_range designed to do.

_No response_

### Relevant log output

_No response_

### Environment details

_No response_

### Other/Misc.

_No response_

isVoid:
## Description

closes #1055

This PR adds `multilinestring_segment_manager` owning class to track the lifetime of intermediate allocations that has to do with segment iterators. In addition `multilinestring_segment_range` is added as the non-owning class to the previous class to provide iterators to the segments. The syntax to use this object is like:

```c++
auto segment_manager = multilinestrings._segments(stream);
auto segment_range = segment_manager.range();
```
Then a user can easily access the segments in the multilinestrings with:
```c++
for (auto seg : segment_range) {
    auto length = sqrt(dot(seg.a, seg.b));
}
```

Secondly, this PR includes tests that has empty geometry collections in the input, linestring_polygon_distance now correctly computes nans for these input pairs.

In addition, `CUSPATIAL_EXPECTS_VALID_MULTI*_SIZES` is relaxed. Valid polygon and ring sizes are only implicitly required but not checked via the size of the arrays since they are insufficient checks. Accordingly, some tests are also relaxed.

## Checklist


cpp/include/cuspatial/detail/multilinestring_segment.cuh
```diff
@@ -0,0 +1,128 @@
...
+    _num_segments = _parent.num_points() - _non_empty_linestring_prefix_sum.element(
+                                             _non_empty_linestring_prefix_sum.size() - 1, stream);
+  }
+
+  /**
+   * @brief Return a non-owning `multilinestring_segment_range` object from this class
+   *
+   * @return multilinestring_segment_range
+   */
+  auto view()
```
harrism, line 112: I think this should be named differently? 

```suggestion
  auto segment_range()
```

cpp/include/cuspatial/detail/multilinestring_segment.cuh
```diff
@@ -0,0 +1,128 @@
...
+ * @note To use this class with a multipolygon range, cast the multipolygon range as a
+ * multilinestring range.
+ *
+ * TODO: Optimization: for range that does not contain any empty linestrings,
+ * `_non_empty_linestring_prefix_sum` can be substituted with `counting_iterator`.
+ *
+ * @tparam MultilinestringRange The multilinestring range to initialize this class with.
+ */
+template <typename MultilinestringRange>
+class multilinestring_segment {
```
harrism, line 71: This name sounds like it is a single segment. What this class really provides is just storage for a segment range. So perhaps we name it as such?

```suggestion
class multilinestring_segment_range_storage {
```

cpp/include/cuspatial/detail/range/multilinestring_segment_range.cuh
```diff
@@ -0,0 +1,270 @@
...
+  }
+};
+
+/// Deduction guide
+template <typename OffsetIterator, typename CountIterator, typename CoordinateIterator>
+to_valid_segment_functor(OffsetIterator, OffsetIterator, CountIterator, CoordinateIterator)
+  -> to_valid_segment_functor<OffsetIterator, CountIterator, CoordinateIterator>;
+
+/**
+ * @brief Non-owning object to `multilinestring_segment` class
```
harrism, line 122:
```suggestion
 * @brief A non-owning range of segments in a multlinestring.
```

cpp/include/cuspatial/detail/range/multilinestring_segment_range.cuh
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * A `multilinestring_segment_range` provide views into the segments of a multilinestring.
+ * The segments of a multilinestring has a near 1-1 mapping to the points of the multilinestring,
+ * except that the last point of a linestring and the first point of the next linestring does not
+ * form a valid segment. For example, the below multilinestring (points are denoted a letters):
```
harrism, line 127:
```suggestion
 * The segments of a multilinestring have a near 1:1 mapping to the points of the multilinestring,
 * except that the last point of a linestring and the first point of the next linestring do not
 * form a valid segment. For example, the below multilinestring (points are denoted a letters):
```

cpp/include/cuspatial/detail/range/multilinestring_segment_range.cuh
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * ```
```
harrism, line 129: Nice explanation. However, should this be in @internal docs?

cpp/include/cuspatial/detail/distance/linestring_polygon_distance.cuh
```diff
@@ -45,36 +44,44 @@
...
+    return lhs.is_empty() || rhs.is_empty();
+  }
+};
+
 /**
  * @brief Computes distances between the multilinestring and multipolygons
  *
- * @param multilinestrings Range to the multilinestring
- * @param multipolygons Range to the multipolygon
+ * This is a load balanced distance compute kernel. Each thread compute exactly 1 pair of segments
```
harrism, line 59: Have you measured performance benefits of the load balancing?

cpp/include/cuspatial/detail/distance/linestring_polygon_distance.cuh
```diff
@@ -119,18 +127,31 @@ OutputIt pairwise_linestring_polygon_distance(MultiLinestringRange multilinestri
...
+  auto size = multilinestrings.size();
+
+  if (size == 0) return distances_first;
 
   // Create a multipoint range from multilinestrings, computes intersection
   auto multipoints           = multilinestrings.as_multipoint_range();
   auto multipoint_intersects = point_polygon_intersects(multipoints, multipolygons, stream);
 
+  // Make views to the segments in the multilinestring
+  auto multilinestring_segments       = multilinestrings._segments(stream);
```
harrism, line 139: Why does `_segment()` have an underscore if its a public method?

cpp/include/cuspatial/range/multilinestring_range.cuh
```diff
@@ -154,23 +153,16 @@ class multilinestring_range {
...
 
-  /// Returns an iterator to the start of the segment
-  CUSPATIAL_HOST_DEVICE auto segment_begin();
-
-  /// Returns an iterator to the end of the segment
-  CUSPATIAL_HOST_DEVICE auto segment_end();
+  /// @internal
+  /// Returns the owning class that provides views into the segments of the multilinestring range
+  /// Can only be constructed on host
+  auto _segments(rmm::cuda_stream_view);
```
harrism, line 165: Why is this internal?

harrism (changes requested):
Nice algorithmic improvement. Some doc suggestions, some questions.

cpp/include/cuspatial/detail/distance/linestring_polygon_distance.cuh
isVoid: Will do some benchmarking to measure it. I can imagine when a kernel is extremely skewed to make the loop long there's performance benefit, but should probably also measure average cases.

cpp/include/cuspatial/range/multilinestring_range.cuh
isVoid: `multilinestring_segment_range` and `multilinestring_segment` are both in detail namespace.

cpp/include/cuspatial/detail/multilinestring_segment.cuh
harrism: My favorite of those is `multilinestring_segment_manager`.  Did you actually show the code for the class to ChatGPT and it wrote these?

isVoid: No. This is my prompt:
```
What are the common naming techniques to show that a C++ class is an owning object?
```

Then it showed a few techniques. I made these names based on its advice.

cpp/include/cuspatial/detail/multilinestring_segment.cuh
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * The owned memory in this struct is the vector `_non_empty_linestring_prefix_sum` of size equal to
+ * the number of linestrings in the multilinestring range plus 1. This vector holds the number
+ * of non empty linestrings that precedes the current linestring.
+ *
+ * This class is only meant for tracking the life time of the owned memory. To access the
+ * segment iterators, call `segment_range()` function to create a non-owning object of this class.
+ *
+ * For detailed explanation on the implementation of the segment iterators, see documentation
+ * of `multilinestring_segment_range`.
```
harrism, line 60: Does quoting documented symbol names prevent Doxygen from making them into clickable links?

harrism (approved):
Thanks for the renamings. Looks good.

Issue #869, opened 2023-01-23
[FEA]:  Use pip to install cuspatial


scalemailted:
### What is your question?

Given that Google Colaboratory now supports python3.8 and some of the other rapidai packages now support pip installation, is it possible to use pip to install cuspatial into a Google Colab notebook?

thomcom:
Thanks for your request @scalemailted ! The answer to your question is that we are working on it and it should be coming up soon.

scalemailted:
Thank you so much for both responses. I am so impressed by the responsiveness of this team! It appears that my question was elevated into a feature request and picked up into a sprint within just days of my asking.

harrism:
In this case your question happens to be aligned with existing requests / plans. The rest of RAPIDS is already moving to support PIP, so we want to do so for cuSpatial as well.

jarmak-nv:
Hi @scalemailted - just wanted to update you that due to some competing priorities we had to push this back to 23.06 - for 23.04 conda, docker, and building from source will remain the installation methods for cuSpatial.

PR #1148, opened 2023-05-22
[DRAFT] cuSpatial pip packages

trxcllnt:
## Description
Initial draft PR to build statically linked cuspatial wheels.

Closes https://github.com/rapidsai/cuspatial/issues/869.

## Checklist


harrism:
Please use Github's draft feature and avoid text in brackets in PR titles. By using Github's feature, we avoid assigning reviewers before they are needed.

vyasr (changes requested):
Before we go any further, could we split out the changes for updating to the breaking changes in libcudf's column/column_view constructors from this PR? And if there are devcontainers changes here that aren't strictly relevant for wheels (it seems like some unrelated refactoring made it in) could we split that out too?

trxcllnt:
@vyasr The devcontainer updates are necessary in order to build the pip package in the devcontainer environment, so I'd like to keep those in.

harrism:
```
WARNING:root:Failed to get options via gdal-config: [Errno 2] No such file or directory: 'gdal-config'
  CRITICAL:root:A GDAL API version must be specified. Provide a path to gdal-config using a GDAL_CONFIG environment variable or use a GDAL_VERSION environment variable.
  error: subprocess-exited-with-error
```

Hrmmm

ci/release/update-version.sh
```diff
@@ -67,10 +70,9 @@ done
...
-# Dependency versions in setup.py
-sed_runner "s/rmm==.*\",/rmm==${NEXT_SHORT_TAG_PEP440}.*\",/g" python/cuspatial/setup.py
-sed_runner "s/cudf==.*\",/cudf==${NEXT_SHORT_TAG_PEP440}.*\",/g" python/cuspatial/setup.py
+# Python pyproject.toml updates
+sed_runner "s/^version = .*/version = \"${NEXT_FULL_TAG}\"/g" python/cuspatial/pyproject.toml
 
 # Dependency versions in pyproject.toml
-sed_runner "s/rmm==.*\",/rmm==${NEXT_SHORT_TAG_PEP440}.*\",/g" python/cuspatial/pyproject.toml
 sed_runner "s/cudf==.*\",/cudf==${NEXT_SHORT_TAG_PEP440}.*\",/g" python/cuspatial/pyproject.toml
+sed_runner "s/rmm==.*\",/rmm==${NEXT_SHORT_TAG_PEP440}.*\",/g" python/cuspatial/pyproject.py
```
bdice, line 78:
```suggestion
sed_runner "s/rmm==.*\",/rmm==${NEXT_SHORT_TAG_PEP440}.*\",/g" python/cuspatial/pyproject.toml
```

conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -36,7 +36,7 @@ dependencies:
 - pytest-xdist
 - python>=3.9,<3.11
 - rmm==23.6.*
-- scikit-build>=0.13.1
+- scikit-build>=0.13.1,<0.17.2
```
bdice, line 39: I think this can be removed, see https://github.com/rapidsai/cudf/pull/13424

dependencies.yaml
```diff
@@ -41,6 +41,33 @@ files:
...
       - docs
       - py_version
+  py_build:
+    output: [requirements, pyproject]
+    pyproject_dir: python/cuspatial
+    requirements_dir: python/cuspatial
+    extras:
+      table: build-system
+    includes:
+      - build_cpp
```
bdice, line 51: The `build_cpp` lists only have `conda` output sections, which will have no effect on the `py_build` dependency file since it's outputting `requirements` and `pyproject`. I think we can safely remove:
```suggestion
```

dependencies.yaml
```diff
@@ -86,15 +113,22 @@ dependencies:
...
   build_python:
     common:
-      - output_types: [conda, requirements]
+      - output_types: [conda, requirements, pyproject]
         packages:
           - *cmake_ver
           - cython>=0.29,<0.30
-          - scikit-build>=0.13.1
+          - ninja
+          - scikit-build>=0.13.1,<0.17.2
```
bdice, line 121:
```suggestion
          - scikit-build>=0.13.1
```

dependencies.yaml
```diff
@@ -108,28 +142,22 @@ dependencies:
...
-              cuda: "11.4"
-            packages:
-              - "--extra-index-url=https://pypi.nvidia.com"
-              - cudf-cu11==23.6.*
-          - matrix:
-              cuda: "11.2"
-            packages:
-              - "--extra-index-url=https://pypi.nvidia.com"
-              - cudf-cu11==23.6.*
+          - matrix: {cuda: "11.8"}
```
bdice, line 147: Could we retain the "longer" form of YAML with more newlines/etc.? We don't use the shorter inline syntax like `{matrix: {cuda: "11.5}, ...}` in any other `dependencies.yaml` files afaik.

python/cuspatial/pyproject.toml
```diff
@@ -13,12 +13,98 @@
...
+[project]
+name = "cuspatial"
+version = "23.6.0"
+description = "cuSpatial: GPU-Accelerated Spatial and Trajectory Data Management and Analytics Library"
+readme = { file = "README.md", content-type = "text/markdown" }
+authors = [
+    { name = "NVIDIA Corporation" },
+]
+license = { text = "Apache 2.0" }
+requires-python = ">=3.8"
```
bdice, line 37: We are dropping Python 3.8 in RAPIDS 23.06.
```suggestion
requires-python = ">=3.9"
```

bdice (commented):
Nice work! Just a few small comments.

dependencies.yaml
trxcllnt: If we do, IMO this file gets much more difficult to parse (and will only get worse the over time). Much prefer the shorter syntax here.

thomcom (approved):

bdice (approved):

dependencies.yaml
harrism: Paul has explained to me that there are other (WIP) reasons to prefer this syntax.

vyasr: I tend to agree with Bradley here, I find the longer-form syntax more readable despite being more verbose. The lines below where the packages are also collapsed into the same line aren't my favorite, for instance. It's sort of like reading JSON where any pretty printer will split the fields into multiple lines.

I won't personally block on this change though since I don't read cuspatial code enough.

python/cuspatial/CMakeLists.txt
```diff
@@ -40,22 +42,54 @@ else()
...
-  install(TARGETS cuspatial DESTINATION cuspatial/_lib)
+  set(BUILD_TESTS OFF)
+  set(BUILD_BENCHMARKS OFF)
+  set(_exclude_from_all "")
+  if(CUSPATIAL_BUILD_WHEELS)
+
+    # We don't build C++ tests when building wheels, so we can also omit the test util and shrink
+    # the wheel by avoiding embedding GTest.
+    set(CUDF_BUILD_TESTUTIL OFF)
+    set(CUDF_BUILD_STREAMS_TEST_UTIL OFF)
```
vyasr, line 53: These are redundant because you're setting `BUILD_TESTS OFF` above and that is directly forwarded to the cudf build AFAICT.

vyasr (commented):
It looks like the flake8 configuration got lost.

There are a few things worth fixing here, but if you decide that this really has to get in ASAP I will leave these as comments without requesting changes so you can move forward and address later. I don't think anything is a strict blocker, the flake8 issue is the biggest one I see.

dependencies.yaml
ajschmidt8, line 1: The consequence of the `output:` changes in this file is that the following 4 files are created anytime `rapids-dependency-file-generator` is invoked without any arguments:

- `python/cuspatial/requirements_py_build.txt`
- `python/cuspatial/requirements_py_run.txt`
- `python/cuspatial/requirements_py_test.txt`
- `python/requirements_checks.txt`

Is this intended?

If so, we should commit those files to this PR.

If not, we should set the output back to `none` for the relevant entries.

ajschmidt8 (changes requested):

dependencies.yaml
vyasr: Great catch AJ. My understanding is that the new files are necessary for devcontainers, and if that is the case then yes we should commit those files to the repo.

python/cuspatial/CMakeLists.txt
harrism: Why is build tests off?

harrism (approved):
Readme instructions looks good!

dependencies.yaml
trxcllnt: I added these so they're compatible with the filtering in https://github.com/rapidsai/dependency-file-generator/pull/48, but I can remove them for now and add them later once that PR is merged.

python/cuspatial/CMakeLists.txt
vyasr: We don't want to build test into the wheel, and we certainly don't want to build libcudf tests into a cuspatial python wheel.

vyasr (approved):

README.md
```diff
@@ -133,7 +133,55 @@ We also provide nightly Conda packages built from the HEAD of our latest develop
 
 See the [RAPIDS release selector](https://rapids.ai/start.html#get-rapids) for more OS and version info.
 
-### Install from Source
+### Install with pip
+
+To install via pip:
+> **Note** cuSpatial is supported only on Linux or [through WSL](https://rapids.ai/wsl2.html), and with Python versions 3.9 and later
```
jarmak-nv, line 139:
```suggestion
> **Note** cuSpatial is supported only on Linux or [through WSL](https://rapids.ai/wsl2.html), and with Python versions 3.9 and 3.10
```

Being specific that we don't support 3.11/12

README.md
```diff
@@ -133,7 +133,55 @@ We also provide nightly Conda packages built from the HEAD of our latest develop
...
+The cuSpatial pip packages can be installed from NVIDIA's PyPI index:
+
+```shell
+# If using driver 525+, with support for CUDA Toolkit 12.0+
+pip install --extra-index-url=https://pypi.nvidia.com cuspatial-cu12
+
+# If using driver 450.80+, with support for CUDA Toolkit 11.2+
+pip install --extra-index-url=https://pypi.nvidia.com cuspatial-cu11
+
+# Or do this if you're unsure which CUDA Toolkit is supported by your driver:
```
jarmak-nv, line 150: This is just cool! Going to use it in the docs.rapids.ai/install page I think!

README.md
```diff
@@ -118,7 +118,7 @@ docker run --gpus all --rm -it \
     nvcr.io/nvidia/rapidsai/rapidsai-core:23.02-cuda11.8-runtime-ubuntu22.04-py3.10
 ```
 
-### Install from Conda
+### Install with Conda
 
 To install via conda:
 > **Note** cuSpatial is supported only on Linux or [through WSL](https://rapids.ai/wsl2.html), and with Python versions 3.9 and later
```
jarmak-nv, line 124:
```suggestion
> **Note** cuSpatial is supported only on Linux or [through WSL](https://rapids.ai/wsl2.html), and with Python versions 3.9 and 3.10
```

3.11/12 support fix.

jarmak-nv (commented):
Great stuff! Recommend making the suggestions so 3.11+ users don't try to install, but the readme updates are super thorough and clear. Love to see it!

README.md
trxcllnt: There might be a better way to get the CTK major version for the current driver, this is just something I whipped up on short notice.

vyasr: This looks eerily similar to something I just wrote yesterday: https://github.com/rapidsai/cudf/pull/13478/files#diff-76e052f6896a2d6de32795d5b94531dc822e5fc452495f2a1ba2e21ad2ee4970R14

trxcllnt: Yeah it is, but it uses `nvidia-smi` (which comes with the driver) instead of `nvcc`.

PR #1178, opened 2023-06-06
Use grid_stride_range in kernel loops

### LINKED ISSUE #650: [BUG] Should use grid-stride loop for several refactored APIs
**Describe the bug**
In `linestring_distance.cuh` and `hausdorff.cuh`, each thread should perform a grid-stride loop as shown in https://github.com/rapidsai/cuspatial/pull/573#discussion_r941915184. Previously this didn't fail be cause the size of input is limited by the size of cudf column, which is small enough to fit in a grid.

harrism:
Closes #650 

## Description
Adds a dependency on [Ranger](https://github.com/harrism/ranger) and uses `ranger::grid_stride_loop` with a range-based for loop in outer loops in every raw kernel in cuSpatial.  In the few cases where we had raw kernels without grid-stride loops, this adds one. Also adds guidance to the C++ developer guide.

## Checklist


isVoid (approved):
Thanks!

robertmaynard (changes requested):

robertmaynard (approved):

vyasr:
Do we want to move ranger into the rapidsai org or somewhere else shared before we introduce a dependency on it here?

harrism:
> Do we want to move ranger into the rapidsai org or somewhere else shared before we introduce a dependency on it here?

I am not opposed to that. That would make it easier for e.g. cuDF to adopt it.

harrism:
Decision is to merge this as-is for now. We can look at re-homing ranger if and when it is determined to be needed.

PR #1186, opened 2023-06-07
Improve Polygon.contains(LineString) predicate logic.

thomcom:
## Description
Depends on https://github.com/rapidsai/cuspatial/pull/1166

The previous logic for handling the case of a LineString that shares boundary points with a Polygon and has no interior points was wrong. This PR fixes that logic.

## Checklist


python/cuspatial/cuspatial/utils/binpred_utils.py
```diff
@@ -423,3 +400,62 @@ def _multipoints_is_degenerate(geoseries):
     ) & (y1.reset_index(drop=True) == y2.reset_index(drop=True))
     result[sizes_mask] = is_degenerate.reset_index(drop=True)
     return result
+
+
+def _pli_features_rebuild_offsets(pli, features):
```
isVoid, line 405: Documentation

python/cuspatial/cuspatial/utils/binpred_utils.py
```diff
@@ -423,3 +400,62 @@ def _multipoints_is_degenerate(geoseries):
...
+    )
+    offsets = cudf.Series(pli[0])
+    offset_sizes = offsets[1:].reset_index(drop=True) - offsets[
+        :-1
+    ].reset_index(drop=True)
+    in_indices = offset_sizes.index.repeat(offset_sizes)
+    if len(in_indices) == 0:
+        return _zero_series(len(pli[0]))
+    # Just replacing the indices from the feature series with
+    # the corresponding indices that are specified by the offsets buffer.
```
isVoid, line 417: Is this the "geometry index" (the part id that the current feature belongs to)?

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
```diff
@@ -49,7 +49,23 @@ def _preprocess(self, lhs, rhs):
         preprocessor_result = super()._preprocess_multipoint_rhs(lhs, rhs)
         return self._compute_predicate(lhs, rhs, preprocessor_result)
 
-    def _intersection_results_for_contains(self, lhs, rhs):
+    def _intersection_results_for_contains_linestring(self, lhs, rhs):
```
isVoid, line 52: The name sounds abstruse. It counts the number of intersecting points (and the segment vertices) that are equal to the rhs vertices. Perhaps `_count_equal_intersecting_points`?

isVoid: I also believe function names should reflect its implementation, not purpose.

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
```diff
@@ -87,47 +99,35 @@ def _compute_polygon_polygon_contains(self, lhs, rhs, preprocessor_result):
...
-        # according to the `intersection_with_no_containment` mask.
-        final_result[intersection_with_no_containment] = interior_tests
-        # LineStrings that do not are contained if the sum of intersecting
-        # and containing points is greater than or equal to the number of
-        # points that make up the linestring.
-        final_result[~intersection_with_no_containment] = (
-            contains + intersects >= rhs.sizes
+        rhs_sizes_less_line_intersection_size[
+            rhs_sizes_less_line_intersection_size <= 0
+        ] = 1
```
isVoid, line 126: 1. Why is it possible that there can be rows that has fewer vertices in the linestring to the number of intersection points?

2. Why do you need to explicitly set these rows to 1?

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
```diff
@@ -87,47 +99,35 @@ def _compute_polygon_polygon_contains(self, lhs, rhs, preprocessor_result):
...
+        final_result = contains + point_intersects_count == (
+            rhs_sizes_less_line_intersection_size
```
isVoid, line 128: This seems differ from your statement in the PR description, your description basically says:
```python
final_result = contains == rhs.sizes - linestring_points_on_boundary
```

To make this equal to the formula here, it seems like we need:
```
linestring_points_on_boundary == linestring_intersects_counts - point_intersects_count
```

Though I don't see the connection here. Can you show me some hints?

isVoid (changes requested):

isVoid:
> The sum of the points of the LineString that are contained in the Polygon is equal to the length of the LineString less the number of points in the LineString that are contained in the boundary of the Polygon.

I suggest we use stricter mathematical language here. 
- "sum of the points", there are infinitie points on a linestring, I believe you mean vertices.
- "length of the linestring", you probably mean the total number of vertices rather than the measurement of the linestring.

aka:
```
LineString is contained in the Polygon if:

The number of linestring vertices contained in the polygon equals the number of linestring vertices subtracts the number of vertices that falls on the polygon boundary.
```

aka:
```
LineString is contained in the Polygon if:

All vertices of the linestring either falls on the boundary of the polygon or is contained in the polygon.
```

Right?

harrism:
> LineString is contained in the Polygon if:
> 
> All vertices of the linestring either falls on the boundary of the polygon or is contained in the polygon.

I don't think that's sufficient. Because of concavities and holes, even if all vertices of a linestring fall on the boundary or are contained by the polygon, intermediate points in the edges could fall outside of the polygon where it is concave.

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -326,6 +326,78 @@ def predicate(request):
...
         point_polygon,
     ),
+    "linestring-polygon-cross-concave-edge": (
+        """
+    -----
+    x\\-/x
+    | x |
+    |   |
+    -----
+    """,
```
harrism, line 336: I can't tell what this (or most of the other) ascii art is showing.

python/cuspatial/cuspatial/utils/binpred_utils.py
```diff
@@ -423,3 +400,62 @@ def _multipoints_is_degenerate(geoseries):
...
+
+    # Recompute the offsets for the new series
+    grouped_sizes = in_sizes.groupby(level=0).sum().sort_index()
+    out_sizes = _zero_series(len(pli[0]) - 1)
+    out_sizes.iloc[grouped_sizes.index] = grouped_sizes
+    offsets = cudf.concat([cudf.Series([0]), out_sizes.cumsum()])
+    return offsets
+
+
+def _pli_points_to_multipoints(pli):
```
harrism, line 436: In all of these, should define what "pli" means.

harrism (changes requested):

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
```diff
@@ -87,47 +99,35 @@ def _compute_polygon_polygon_contains(self, lhs, rhs, preprocessor_result):
...
+        # Each linestring overlap counts as two point overlaps.
+        (
+            point_intersects_count,
+            linestring_intersects_count,
+        ) = self._intersection_results_for_contains_linestring(lhs, rhs)
+
+        # Subtract the length of the linestring intersections from the length
+        # of the rhs linestring, then test that the sum of contained points
+        # is equal to that adjusted rhs length.
+        rhs_sizes_less_line_intersection_size = (
```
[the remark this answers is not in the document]
thomcom: It is just python `black` formatting conventions. The line on a single line is 87 chars long, max is 79. Multi-line statements in python under `black` are made with parentheses such as this.

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
thomcom: Thanks for investigating this logic. This step is important to cover the tests of polygon.contains(linestring) where the LineString is contained only by the border off the polygon. In the above line, the number of points in the linestring is reduced by the number of points in the border-overlap region:

```
rhs_sizes_less_line_intersection_count = rhs.sizes - linestring_intersects_count
```
This is because LineString segments that overlap with the boundary of a polygon are not used in the counting of the vertices that may be contained in the polygon. Consider the following three examples:

image

Note that each shape has examples of the LineString overlapping the boundary of the polygon. In the first case, an edge overlaps from the corner, then goes inside to an interior point, then returns to the edge, then overlaps to the corner. The length of the LineString is 5. The first overlapping segment subtracts 2 from the length of the linestring, and the second overlapping segment subtracts another 2, leaving a linestring of length 1. 1 point is contained, and therefor the linestring is contained.

In the second example the same pattern occurs except the middle point goes out of the polygon instead of in. The same points are subtracted from the length of the linestring, but the number of contained points is 0, less than the (remaining) length of the linestring of 1, so it is not contained.

Finally, the LineString only overlaps with the edge of the polygon, on the left and on the bottom. This returns a LineString from pairwise-linestring-intersection of length 3, which is subtracted from the length of the LineString, also 3. `rhs_sizes_less_line_intersection_size` is now 0 for this example LineString, so `contains + point_intersects_count == 0`, which would return `contains = True`. Therefore I set the minimum LineString length to 1, so that some LineString point must be contained in order for `.contains` to evaluate to true.

Happy to chat this through with you in person if you can think of a more parsimonious approach.

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
```diff
@@ -87,47 +99,35 @@ def _compute_polygon_polygon_contains(self, lhs, rhs, preprocessor_result):
...
-        final_result[intersection_with_no_containment] = interior_tests
-        # LineStrings that do not are contained if the sum of intersecting
-        # and containing points is greater than or equal to the number of
-        # points that make up the linestring.
-        final_result[~intersection_with_no_containment] = (
-            contains + intersects >= rhs.sizes
+        rhs_sizes_less_line_intersection_size[
+            rhs_sizes_less_line_intersection_size <= 0
+        ] = 1
+        final_result = contains + point_intersects_count == (
```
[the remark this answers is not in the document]
thomcom: Python `black` formatting constraints for maximum line length of 79.

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
thomcom: In the description I believe I misspoke about the number of points that are contained in the polygon as demonstrated in the code:

```
final_result = contains + point_intersects_count == rhs.sizes - linestring_points_on_boundary
```

IIRC, you are concerned that `linestring_points_on_boundary` in your example doesn't take into account the number of point intersections, which are ignored in the description of this PR but not in `_compute_polygon_linestring_contains`.

`contains` equals the number of the LineString vertices that are contained by the polygon.

`point_intersects_count` equals the number of LineString vertices that intersect with the polygon: that is, the places where the LineString does not share an overlapping boundary with the polygon, but instead has a vertex that is colinear with the boundary of the polygon, but each segment that shares the vertex is not colinear. When a `point_intersects_count` intersection occurs, there must either: exist a point that is inside the polygon that will be counted by `contains` or a point outside of the polygon that will not be counted. 

When a vertex is colinear with the boundary of the polygon this is counted as "in" the polygon for a reason related to our above conversation, where the minimum LineString length must be 1. Consider the following example:

image

In the first case, `point_intersects_count == 2`, as the LineString vertices are colinear with the boundary of the polygon. `contains + point_intersects_count == rhs.sizes` and the LineString is contained.

In the second case, `point_intersects_count == 0`, as the LineString vertices are not colinear- it is a genuine intersection. `contains + point_intersects_count < rhs.sizes` and the LineString is not contained.

`linestring_intersects_count` is the number of LineString vertices that are from segments that are colinear with the polygon boundary. They are used only to reduce the number of points in the LineString that should be tested for .contains.

The final equation would be
```final_result = contains + point_intersects_count == rhs.sizes - linestring_intersects_count```

equal to

```final_result = contains == rhs.sizes - linestring_intersects_count - point_intersects_count``` as I think you are alluding to above, except that the LineString must maintain a size of at least 0 as you asked about above. Maybe I'm missing something here that would allow me to subtract `point_intersects_count` instead of using the line that sets the minimum `rhs_sizes_less_line_intersection_size` to 1. I'll be thinking of that.

Consider the final case in this response, 

image

In this case `point_intersects_count = 1`, and due to the overlapping segment, `rhs_sizes_less_line_intersection_count == 1`, and the LineString is contained. I could do `contains == rhs.sizes - linestring_intersects_count - point_intersects_count`, which in this case would be `(0 == 0)`, but that would break the third case in my earlier response, where the LineString is entirely colinear with the boundary of the polygon. In those cases the LineString is not contained, and as such it must be that `rhs_sizes_less_line_intersection_count > 0`

File: python/cuspatial/cuspatial/core/binpreds/feature_contains.py [hunk 10]
```diff
@@ -87,47 +99,35 @@ def _compute_polygon_polygon_contains(self, lhs, rhs, preprocessor_result):
...
+            rhs_sizes_less_line_intersection_size
         )
+
         return final_result
```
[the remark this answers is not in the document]
thomcom: The reason we see this pattern so frequently is so that I can debug the `final_result` in the context of its precursors and inspect it while designing the predicate, which is very convenient. I'd rather not remove all of these from the codebase and then readd them whenever I need to see the inner workings of a particular result, ok?

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
thomcom: It certainly isn't as informative as the graphs produced by `shapely` but I can't insert those here. Really they're just intended as a visual guide to make it easier to parse the specific geometry definitions that come immediately afterward, instead of having to do all mental or all on-paper arithmetic to reproduce what's being tested:
```
    "linestring-polygon-cross-concave-edge": (
        """
    -----
    x\\-/x
    | x |
    |   |
    -----
    """,
        LineString([(0.5, 0.0), (0.5, 1.0)]),
        Polygon([(0, 0), (0, 1), (0.3, 0.4), (1, 1), (1, 0)]),
    ),
```
This one in particular is just plain wrong though so I've fixed that:
```
    x  x  x
    |\ | /|
    | xx- |
    |  |  |
    ---x---
```
Which corresponds with this shapely image:
image

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
isVoid: > Note that each shape has examples of the LineString overlapping the boundary of the polygon. In the first case, an edge overlaps from the corner, then goes inside to an interior point, then returns to the edge, then overlaps to the corner. The length of the LineString is 5. The first overlapping segment subtracts 2 from the length of the linestring, and the second overlapping segment subtracts another 2, leaving a linestring of length 1. 1 point is contained, and therefor the linestring is contained.
> 
> In the second example the same pattern occurs except the middle point goes out of the polygon instead of in. The same points are subtracted from the length of the linestring, but the number of contained points is 0, less than the (remaining) length of the linestring of 1, so it is not contained.

Pardon me from not following the reasoning here. I took `linestring_intersects_count` meaning the number of vertices of the linestring that intersects with the ring of the polygon. Isn't for both cases, the value is 4? The subtracted result is 1 right?

> Finally, the LineString only overlaps with the edge of the polygon, on the left and on the bottom. This returns a LineString from pairwise-linestring-intersection of length 3, which is subtracted from the length of the LineString, also 3. rhs_sizes_less_line_intersection_size is now 0 for this example LineString, so contains + point_intersects_count == 0, which would return contains = True. Therefore I set the minimum LineString length to 1, so that some LineString point must be contained in order for .contains to evaluate to true.

This sounds a bit hacky to me. Also, at least the comparison should be `==` not `<=` right?

python/cuspatial/cuspatial/utils/binpred_utils.py
thomcom: I'm not sure what you are asking. See the updated docs perhaps? Consider the example now in `_pli_points_to_multipoints`:

```
    pli = (
        cudf.Series([0, 2, 4, 6]),
        cuspatial.GeoSeries([
            Point(0, 0),
            LineString([(1, 1), (2, 2)]),
            Point(3, 3),
            Point(4, 4),
            Point(5, 5),
            LineString([(6, 6), (7, 7)]),
        ])
    )
```
In this example, each pair of features in the GeoSeries (indices [0, 1], [2, 3], and [4, 5]) were produced by input feature pairs (lhs[0], rhs[0]), (lhs[1], rhs[1]), and (lhs[2], rhs[2]) respectively. The offsets buffer [0, 2, 4, 6] specifies this. However, I only want the points, so this function takes the points from the pli result (and their corresponding indices):
```
print(pli[1][pli[1].feature_type == Feature_Enum.POINT.value])
0    POINT (0.00000 0.00000)
2    POINT (3.00000 3.00000)
3    POINT (4.00000 4.00000)
4    POINT (5.00000 5.00000)
dtype: geometry
```
and creates a new offset buffer using only these values, specified by the old `offset` groupings:
```
[0, 2, 4, 6] -> [0, 1, 3, 4]
```

python/cuspatial/cuspatial/utils/binpred_utils.py
```diff
@@ -423,3 +400,62 @@ def _multipoints_is_degenerate(geoseries):
...
+
+def _pli_features_rebuild_offsets(pli, features):
+    in_sizes = (
+        features.sizes if len(features) > 0 else _zero_series(len(pli[0]) - 1)
+    )
+    offsets = cudf.Series(pli[0])
+    offset_sizes = offsets[1:].reset_index(drop=True) - offsets[
+        :-1
+    ].reset_index(drop=True)
+    in_indices = offset_sizes.index.repeat(offset_sizes)
```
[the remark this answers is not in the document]
thomcom: Yeah this gets used a lot, let me see what i can do.

python/cuspatial/cuspatial/core/binpreds/feature_contains.py
isVoid: I see you arrive at the same equality as I did above. 
```
final_result = contains == rhs.sizes - linestring_intersects_count - point_intersects_count
```

Great. I'm still confused with the discrepency between this and code. But perhaps focus on Mark's question below first.

python/cuspatial/cuspatial/utils/binpred_utils.py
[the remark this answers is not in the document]
thomcom: I don't see an immediate opportunity to improve the refactoring around this, let's save it for later?

python/cuspatial/cuspatial/utils/binpred_utils.py
isVoid: Thanks this is clear now.

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
harrism: Tip: use different characters for the vertices of different geometries. E.g. `x` for the polygon here and `o` for the linestring vertices. And possibly a third character for the "proper" intersections that occur (`*`?).

harrism (approved):

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -389,11 +389,11 @@ def predicate(request):
...
-    x----
-    |   |
-    |   |-x
-    | --/
-    x----
+    x------
+    |     |
+    |    ---x
+    | --- |
+    x------
```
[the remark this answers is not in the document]
harrism: Not ideal for ASCII...

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -328,11 +328,11 @@ def predicate(request):
...
-    -----
-    x\\-/x
-    | x |
-    |   |
-    -----
+    x  x  x
+    |\\ | /|
+    | xx- |
+    |  |  |
+    ---x---
```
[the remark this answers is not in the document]
thomcom: Double backslash is required for python inline comments like this, unfortunately. I like your diagram.

python/cuspatial/cuspatial/utils/binpred_utils.py
```diff
@@ -403,6 +403,24 @@ def _multipoints_is_degenerate(geoseries):
...
+    The original offsets buffer at `pli[0]` refers to the geometry pair
+    from which the feature intersections from `pli[1]` were computed. The new
+    offsets buffer specifies the same thing, but now accounts for the missing
+    values (either Points or LineStrings) that are present in `pli[1]` but
+    missing in the `features` argument.
+
+    See the docs for `_pli_points_to_multipoints` and
+    `_pli_lines_to_multipoints` for the rest of the explanation.
+    """
     in_sizes = (
```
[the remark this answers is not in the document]
thomcom: That's correct.

python/cuspatial/cuspatial/utils/binpred_utils.py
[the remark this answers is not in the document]
isVoid: Sure. We can punt till later.

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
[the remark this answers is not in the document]
isVoid: Right, this is similar to C style backslash escape.

isVoid (approved):

PR #1192, opened 2023-06-08
Simplify point-in-polygon header only APIs

### LINKED ISSUE #707: Simplify header-only point-in-polygon API
With introduction of `multi*_range` classes, point-in-polygon API can be simplified to just accepting a `multipoint_range` and a `multipolygon_range`.

Note that this refactor will "augment" the existing API to support multi* geometries. 

For `pairwise_point_in_polygon`, a simple way is to add additional loops on top of the current kernel. This can allow the API to support geometry collections, while maintaining performance for non-multi geometry inputs.

For `point_in_polygon`, the `all-pairs` version, it can be simplified to a `multipoint_ref` to a `multipolygon_ref`. In `multipolygon` we require that the input to be order-aware. Alternatively it can be a `multipoint_range` and `multipolygon_range`, while limiting the geometry offset iterators to be only counting iterators.

isVoid:
## Description

closes #707 

This PR simplifies `point_in_polygon` and `pairwise_point_in_polygon` API using `multipoint_range` and `multipolygon_range`. While these range methods supports a pair of multi geometry semantics, in this PR I'm not really targeting to create an API that works with multipoint-in-multipolygon semantics (such as "any point in any polygon", or "all points in any polygon", that sort of semantics). Therefore, this PR introduces `contains_only_single_geometry` method in `multipoint_range` and `multipolygon_range` that provides compile-time check for API that only want to work single-type geometry ranges. There are caveats to this introduction, see *caveats* section below.

This refactor results in a net decrease in LOC, and the compatibility layer for `is_point_in_polygon` is removed. The API dries up to a point where there's no raw kernel anymore.

In addition, previous `pairwise_point_in_polygon` API stores the row wise result in an int32_t per pair. Since it's only storing booleans, this PR uses `uint8_t` instead. This saves memory and increases throughputs.

### Caveats
`contains_only_single_geometry` method is an over-constrained method comparing to what it's really testing. A multipoint range can have a materialized column of integer sequences and still represent a single geometry column. The reason behind this refactor is that the original point-in-polygon test explicitly requires (by the number of arguments) that only single geometry columns accepted by the API. This means developers know at compile time that the geometry used in the API are constructed this way. `contains_only_single_geometry` is a `constexpr` method and verifies developers construction at compile time as well. This maintains the developer expectation while allowing `multi*_range` to be retrofit into the modern APIs.

## Checklist


cpp/include/cuspatial/detail/range/multipolygon_range.cuh
```diff
@@ -118,6 +118,19 @@ multipolygon_range<GeometryIterator, PartIterator, RingIterator, VecIterator>::m
...
+          typename RingIterator,
+          typename VecIterator>
+CUSPATIAL_HOST_DEVICE constexpr bool
+multipolygon_range<GeometryIterator, PartIterator, RingIterator, VecIterator>::
+  contains_only_single_geometry()
+{
+  using index_t =
+    multipolygon_range<GeometryIterator, PartIterator, RingIterator, VecIterator>::index_t;
+  return std::is_same_v<GeometryIterator, thrust::counting_iterator<index_t>>;
+}
```
isVoid, line 132: We can actually check this rather cheaply at runtime. Simply check `num_polygon == num_multipolygon`. Will change this.

harrism: Is there a need to do it at runtime rather than statically?

harrism (approved):
Nice improvement. Cool that it eliminates some raw kernels.

cpp/include/cuspatial/detail/range/multipolygon_range.cuh
isVoid: Yes, because checking if the iterator is a `counting_iterator` is too strict a check for single-polygon range. You can have a materialized array that stores a sequence, each increment by 1 and still represent a single-polygon range.

isVoid: The static check can be named something else, "is_lazy_evaluated_single_polygon_range"? If we find a use to it.

harrism: Just make it runtime and if we need a static version later we can add it.

isVoid:
To Ops: The changes in scripts are only for temporary test runs. Will revert once debugged.

cpp/include/cuspatial/detail/point_in_polygon.cuh
```diff
@@ -24,240 +24,109 @@
...
-                "OutputIt must point to 32 bit integer type.");
+  static_assert(std::is_same_v<iterator_value_type<OutputIt>, uint8_t>,
+                "OutputIt must be iterator to a uint8_t range.");
 
-  auto const num_test_points = std::distance(test_points_first, test_points_last);
-  auto const num_polys       = std::distance(polygon_offsets_first, polygon_offsets_last) - 1;
-  auto const num_rings       = std::distance(poly_ring_offsets_first, poly_ring_offsets_last) - 1;
-  auto const num_poly_points = std::distance(polygon_points_first, polygon_points_last);
+  CUSPATIAL_EXPECTS(points.num_multipoints() == points.num_points(),
+                    "Point in polygon API only support single point - single polygon tests. "
```
harrism, line 109:
```suggestion
                    "Point in polygon API only supports single point - single polygon tests. "
```

cpp/include/cuspatial/detail/point_in_polygon.cuh
```diff
@@ -24,240 +24,109 @@
...
+                    "Multipoint input is not accepted.");
 
-  CUSPATIAL_EXPECTS_VALID_POLYGON_SIZES(
-    num_poly_points,
-    std::distance(polygon_offsets_first, polygon_offsets_last),
-    std::distance(poly_ring_offsets_first, poly_ring_offsets_last));
+  CUSPATIAL_EXPECTS(polygons.num_multipolygons() == polygons.num_polygons(),
+                    "Point in polygon API only support single point - single polygon tests. "
```
harrism, line 113:
```suggestion
                    "Point in polygon API only supports single point - single polygon tests. "
```

harrism (approved):

PR #1195, opened 2023-06-08
Remove `osmnx` dependency and use a small local dataset

### LINKED ISSUE #1194: [BUG]: Remove external web service dependency in notebooks
### Version

23.08

Rapids-Compose

### Describe the issue

In branch-23.06 we introduced `osmnx` dependency to `cuspatial-api-examples`. While it's a very useful package to grep osm dataset in example, it depends on web services that are often unstable and may fail when called repeatedly from the same host. Since notebook are tested in CI, we should remove this dependency.

_No response_

### Relevant log output

_No response_

### Environment details

_No response_

### Other/Misc.

_No response_

isVoid:
## Description

close #1194
This PR removes `osmnx` dependency and instead use a small locally hosted dataset.

## Checklist


docs/source/user_guide/esb_3857.csv
```diff
@@ -0,0 +1,3 @@
+element_type,osmid,nodes,addr:city,addr:housenumber,addr:postcode,addr:state,addr:street,building,building:colour,building:levels,building:use,check_date,construction_date,ele,gnis:feature_id,height,name,name:en,name:es,name:he,name:hi,name:ko,name:ru,name:uk,name:zh,nycdoitt:bin,office,opening_hours,phone,start_date,tourism,website,wheelchair,wikidata,wikipedia,geometry
```
bdice, line 1: It looks like the relevant data here is the polygon shape, but this has names in 9 languages and other metadata. Can we strip this to only the relevant columns? Maybe geometry and place name only?

docs/source/user_guide/streets_3857.csv
```diff
@@ -0,0 +1,1865 @@
+,u,v,key,osmid,oneway,lanes,name,highway,maxspeed,reversed,length,geometry,access,bridge,ref,width,tunnel,junction
```
bdice, line 1: Similar to above -- trim this to only relevant columns.

bdice: GitHub agrees: 
image
https://docs.github.com/en/repositories/working-with-files/using-files/working-with-non-code-files#rendering-csv-and-tsv-data

thomcom (approved):

ajschmidt8 (approved):

PR #1197, opened 2023-06-12
Add Comprehensive Test for Multigeometry Range Classes

### LINKED ISSUE #991: [FEA]: Write tests for `multipolygon_{range/ref}`, `polygon_{range/ref}`, `linestring_{range/ref}`, etc.

Improvement

Medium

I love the new C++ object references you've designed to be used with thrust @isVoid, they seem like they'll really improve development of libcuspatial.

Because they are such fundamental elements of the libcuspatial API, I think that you should write unit tests for them that exercise the functionalities they are expected to have.

_No response_

### Additional context

_No response_

isVoid:
## Description

closes #991

This PR adds comprehensive tests for `multi*_range` class. The basic goal is to make sure every API in service is at least covered by 3 tests: a range with 0, 1 and 1000 elements. The tests are structured as below: for each range, a base fixture class is created. The base class defines virtual functions that its subclass should implement. These virtual functions include a `make_test_multi*` function that constructs the geometry array of that specific test case, as well as each sub test function that defines the expected value of that test case. The base class defines a `SetUp` function that calls the `make_test_multi*` to generate the array at test start up time. Then, a `run_test` function is defined to subsequently call every sub routine that test every API of that geometry range.

In addition, this PR fixes several small bug in `point_begin` accessor in `linestring_ref` and `multipolygon_ref` class. Also, a few unused functions in `multipolygon_range` have been removed.

## Checklist


robertmaynard (approved):
CMake changes look good

cpp/include/cuspatial/detail/geometry/polygon_ref.cuh
```diff
@@ -59,13 +59,13 @@ CUSPATIAL_HOST_DEVICE auto polygon_ref<RingIterator, VecIterator>::ring_end() co
 template <typename RingIterator, typename VecIterator>
 CUSPATIAL_HOST_DEVICE auto polygon_ref<RingIterator, VecIterator>::point_begin() const
 {
-  return _point_begin;
+  return thrust::next(_point_begin, *_ring_begin);
```
harrism, line 62: Can you explain these changes? Why can't point_begin be used as stored?

cpp/include/cuspatial/detail/geometry_collection/multilinestring_ref.cuh
```diff
@@ -74,13 +74,15 @@ CUSPATIAL_HOST_DEVICE auto multilinestring_ref<PartIterator, VecIterator>::part_
...
 {
-  return _point_begin;
+  return thrust::next(_point_begin, *_part_begin);
 }
 
 template <typename PartIterator, typename VecIterator>
 CUSPATIAL_HOST_DEVICE auto multilinestring_ref<PartIterator, VecIterator>::point_end() const
 {
-  return _point_end;
+  // _part_end is the one-past the last part index to the point of this multilinestring.
```
harrism, line 83:
```suggestion
  // _part_end refers to one past the last part index to the points of this multilinestring.
```
I don't quite understand this. A part index indexes points? Is it one past the last point of the part?

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -271,22 +282,34 @@ class multilinestring_array {
...
+                               rmm::device_uvector<vec_2d<T>>>(
+    std::move(geometry_inl), std::move(part_inl), std::move(coord_inl));
+}
 
-  return multilinestring_array<DeviceIndexVector, DeviceIndexVector, DeviceCoordVector>(
-    make_device_vector(std::move(geometry_inl)),
-    make_device_vector(std::move(part_inl)),
-    make_device_vector(std::move(coord_inl)));
+/**
+ * @brief Construct an owning object of a multilinestring array from ranges
```
harrism, line 297: From ranges? or device_vectors?

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -271,22 +282,34 @@ class multilinestring_array {
...
+ *
+ * @param geometry_inl Range of geometry offsets
+ * @param part_inl Range of part offsets
+ * @param coord_inl Ramge of coordinate
+ * @return multilinestring array object
+ */
+template <typename IndexType, typename T>
+auto make_multilinestring_array(rmm::device_vector<IndexType>&& geometry_inl,
```
harrism, line 305: Would be nice to use spans here (in newer libcu++), or template the container type?

harrism (approved):
Wow, this is a lot of tests. Impressed you were able to keep it all straight in a single PR... I confess I couldn't really review all the tests in detail but I like how they are all listed in the base class so it's easy to see the coverage.

cpp/include/cuspatial/detail/geometry/polygon_ref.cuh
isVoid: tl;dr. Because the offsets stored in a `*_ref` are *global* offsets to the lower level range.

Consider this example:
Multilinestring_range:
Geometry offsets: [0, 2, 4]
Part Offsets: [0, 2, 4, 6, 8]
Points: [P0, P1... P8]

A `multilinestring_ref` constructed for the second multilinestring is stored as below:
Part Offsets: [4, 6, 8]
Points: [P0, ... P8]

Notice that in part offset, the offsets are global offsets (4 means that the first point to the multilinestring locates at _point_begin + 4). You may ask why we don't store as below:

Part Offsets: [0, 2, 4]
Points: [P4 ... P8]

Because this requires a scan to the part offset and could be expensive.

cpp/include/cuspatial/detail/geometry_collection/multilinestring_ref.cuh
isVoid: See how multilinestring_ref is constructed: 

https://github.com/isVoid/cuspatial/blob/293e4c30b5b43ea57225fe4e52a55e18f4e99e3a/cpp/include/cuspatial/detail/range/multilinestring_range.cuh#L67

`*(_part_begin + _geometry_begin[i + 1])` is the part index of the last point of the current multilinestring. When constructing it, I increment it to maintain that the the end interator of a range should be one-past the last element.

cpp/include/cuspatial_test/vector_factories.cuh
isVoid: Agreed, the vector factories are fairly unorganized atm. They were created adhoc as handy functions when creating test cases. Definitely needs reorganizing.

Depending on APIs, they can be promoted into the main library if a new algorithm needs to return an owning data structure to store geometries.

cpp/include/cuspatial_test/vector_factories.cuh
isVoid: Fixed doc.

cpp/include/cuspatial_test/vector_factories.cuh
harrism: I missed that this is part of tests. So it's fine to just specialize it to device_vector.

thomcom (approved):
Radical tests.

cpp/include/cuspatial/detail/utility/zero_data.cuh
```diff
@@ -19,6 +19,8 @@
 
 #include <cuspatial/traits.hpp>
 
+#include <rmm/cuda_stream_view.hpp>
+
```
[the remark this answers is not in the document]
isVoid: zero_data_async does take a cuda stream view as input so I think it's fair to add this dependency. We need this because all previous usage has cuda_stream_view.hpp included prior to this file. But this PR cpp/tests/range/multilinestring_range_test.cu includes this file and previous usage doesn't include the stream view dependency.

Issue #1200, opened 2023-06-23
[BUG]: Bad results returned from `pairwise_linestring_intersection` on large self-equal LineStrings.


thomcom:
### Version

23.08

### On which installation method(s) does this occur?

Rapids-Compose

### Describe the issue

Hiya! I found a new issue with `pairwise_linestring_intersection` today working on the Watershed Demo Notebook for Release 23.06. 

When `pairwise_linestring_intersection` is called using the same linestring for the left and right hand sides, results are unpredictable and eventually fail as the linestring size increases. The API fails because the second element returned by `pairwise_linestring_intersection` eventually creates an invalid `GeoSeries` object with bad offsets.

For the reproducer, I created a LineString comprised of diagonal segments of length $sqrt(2)$ originating at the origin and rising by 1 in x and y. The LineString then connects the highest point along the diagonal with a point directly beneath it $(x, 0)$, and that returns to the origin making a closed LineString.

With this specific example, valid (though strange) results are returned up to a LineString of length 129. At 130 the API returns a broken GeoSeries for `geoms`. The results are strange because the first `n` of the segments get grouped into a single segment semi-randomly, say, `((0, 0), (32, 32))`, and then the subsequent segments all are returned separately: `((32, 32), (33, 33)), ((33, 33), (34, 34))`, etc.

Is this a valid error? There are two constraints that maybe make this invalid:
1. The LineStrings in the pairwise comparison are identical
2. The LineStrings are closed with the same start and end vertex.

I can see from the result returned here:
```
    geoms, look_back_ids = c_pairwise_linestring_intersection(
        linestrings1.lines.column(), linestrings2.lines.column()
    )

    (
        geometry_collection_offset,
        types_buffer,
        offset_buffer,
        points,
        segments,
    ) = geoms
```
That types_buffer when the length of the LineString is 130 all come back as points, that is, `type = 0`. When the length of the LineString is 129, they are returned correctly as `type = 2`. `look_back_ids` are also corrupt at length 130 and apparently valid at length 129 in this case.

### Minimum reproducible example

```python
    import cuspatial
    import cupy as cp
    from cuspatial.core.binops.intersection import pairwise_linestring_intersection

    size = 130.0
    xy = cp.concatenate([
        cp.arange(size - 2).repeat(2),
        cp.array([size - 3, 0, 0, 0])
    ])
    s1 = cuspatial.GeoSeries.from_linestrings_xy(
        xy,
        [0, len(xy)/2],
[...]

PR #1207, opened 2023-06-29
Sort the mergeable segments before computing merged segments

isVoid:
## Description

This PR is part-1 fix to \#1200. In `find_and_combine_segments` the N^2 algorithm depends on the fact that the API needs to be presorted with certain criteria. This Pr adds such sorting.

## Checklist


cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
```diff
@@ -19,18 +19,20 @@
...
+#include <thrust/sort.h>
 
 namespace cuspatial {
 namespace detail {
 
 /**
  * @internal
- * @brief Kernel to merge segments, naive n^2 algorithm.
+ * @brief Kernel to merge segments. Each thread works on a pair of segment spaces.
+ * naive n^2 algorithm.
```
isVoid, line 35:
```suggestion
 * @brief Kernel to merge segments. Each thread works on one segment space,
 * with a naive n^2 algorithm.
```

cpp/include/cuspatial/detail/intersection/linestring_intersection.cuh
```diff
@@ -237,13 +237,25 @@ linestring_intersection_result<T, index_t> pairwise_linestring_intersection(
   points.remove_if(range(point_flags.begin(), point_flags.end()), stream);
 
   if (segments.num_geoms() > 0) {
+    segments.debug_print();
```
isVoid, line 240:
```suggestion
```

cpp/include/cuspatial/detail/intersection/linestring_intersection.cuh
```diff
@@ -237,13 +237,25 @@ linestring_intersection_result<T, index_t> pairwise_linestring_intersection(
...
 
+    std::cout << "\n\n";
+    rmm::device_uvector<int> segment_flags_int(num_segments, stream);
+    thrust::copy(rmm::exec_policy(stream),
+                 segment_flags.begin(),
+                 segment_flags.end(),
+                 segment_flags_int.begin());
+    test::print_device_vector(segment_flags_int, "segment_flags");
+    std::cout << "\n\n";
+
```
isVoid, line 254:
```suggestion
```

cpp/include/cuspatial/detail/intersection/linestring_intersection.cuh
```diff
@@ -237,13 +237,25 @@ linestring_intersection_result<T, index_t> pairwise_linestring_intersection(
...
     segments.remove_if(range(segment_flags.begin(), segment_flags.end()), stream);
 
+    segments.debug_print();
+
```
isVoid, line 258:
```suggestion
```

cpp/include/cuspatial/detail/intersection/linestring_intersection_with_duplicates.cuh
```diff
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <cuspatial_test/test_util.cuh>
+
```
isVoid, line 20:
```suggestion
```

cpp/include/cuspatial/detail/intersection/linestring_intersection_with_duplicates.cuh
```diff
@@ -361,6 +363,16 @@ struct linestring_intersection_intermediates {
...
+  void debug_print()
+  {
+    test::print_device_vector(*offsets, "offsets: ");
+    test::print_device_vector(*geoms, "geoms: ");
+    test::print_device_vector(*lhs_linestring_ids, "lhs_linestring_ids: ");
+    test::print_device_vector(*lhs_segment_ids, "lhs_segment_ids: ");
+    test::print_device_vector(*rhs_linestring_ids, "rhs_linestring_ids: ");
+    test::print_device_vector(*rhs_segment_ids, "rhs_segment_ids: ");
+  }
```
isVoid, line 375:
```suggestion
```

cpp/tests/intersection/linestring_intersection_large_test.cu
```diff
@@ -2025,3 +2026,32 @@ TYPED_TEST(LinestringIntersectionLargeTest, LongInput)
...
+                   coordinate_functor<TypeParam>{});
+
+  coordinates.set_element(128, P{127.0, 0.0}, this->stream());
+  coordinates.set_element(129, P{0.0, 0.0}, this->stream());
+
+  auto rng = cuspatial::make_multilinestring_range(
+    1, geometry_offset.begin(), 1, part_offset.begin(), 130, coordinates.begin());
+
+  CUSPATIAL_RUN_TEST(this->template verify_legal_result, rng, rng);
+}
```
isVoid, line 2057: This is supposed to be fixed in #1208. Shouldn't be added as test here.

cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
```diff
@@ -44,6 +46,9 @@ void __global__ simple_find_and_combine_segments_kernel(OffsetRange offsets,
       merged_flag[i] = 0;
     }
 
+    // For each of the segment, loop over the rest of the segment in the space and see
+    // if it is mergeable with the current segment.
+    // Note that if the current segment is already merged. Skip checking.
```
harrism, line 51:
```suggestion
    // For each segment, loop over the rest of the segments in the space and see
    // if any are mergeable with the current segment.
    // Note that if the current segment is already merged, skip checking.
```

cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
```diff
@@ -57,20 +62,66 @@ void __global__ simple_find_and_combine_segments_kernel(OffsetRange offsets,
   }
 }
 
+/**
+ * @brief Comparator for sorting the segment range.
+ *
+ * This comparator makes sure that the segment range are sorted by the following keys:
```
harrism, line 68:
```suggestion
 * This comparator makes sure that the segment range is sorted by the following keys:
```

cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
```diff
@@ -57,20 +62,66 @@ void __global__ simple_find_and_combine_segments_kernel(OffsetRange offsets,
...
+  using index_t   = typename OffsetRange::value_type;
+  using T         = typename SegmentRange::value_type::value_type;
   auto num_spaces = offsets.size() - 1;
   if (num_spaces == 0) return;
 
+  // Construct a key iterator using the offsets of the segment and the segment itself.
+  auto space_id_iter         = make_geometry_id_iterator<index_t>(offsets.begin(), offsets.end());
+  auto space_id_segment_iter = thrust::make_zip_iterator(space_id_iter, segments.begin());
+
+  thrust::sort_by_key(rmm::exec_policy(stream),
```
harrism, line 119: What is the impact of the new sorting on performance of linestring intersection?

cpp/include/cuspatial/geometry/segment.cuh
```diff
@@ -55,6 +55,17 @@ class alignas(sizeof(Vertex)) segment {
   /// Return the length squared of segment.
   T CUSPATIAL_HOST_DEVICE length2() const { return dot(v2 - v1, v2 - v1); }
 
+  /// Return slope of segment.
+  T CUSPATIAL_HOST_DEVICE slope() { return (v2.y - v1.y) / (v2.x - v1.x); }
+
+  /// Return the lower left vertex of segment.
+  Vertex CUSPATIAL_HOST_DEVICE lower_left() { return v1 < v2 ? v1 : v2; }
+
+  bool CUSPATIAL_HOST_DEVICE collinear(segment<T> const& other)
```
harrism, line 64: Seems we use both `colinear` and `collinear` spelling in cuSpatial. Maybe we should choose one and be consistent (even if English is not).

cpp/tests/intersection/linestring_intersection_large_test.cu
```diff
@@ -27,6 +27,7 @@
 #include <cuspatial/geometry/vec_2d.hpp>
 #include <cuspatial/intersection.cuh>
 
+#include <rmm/exec_policy.hpp>
 #include <thrust/host_vector.h>
```
harrism, line 31:
```suggestion
#include <rmm/exec_policy.hpp>

#include <thrust/host_vector.h>
```

harrism (changes requested):
Looks good. Just a few suggestions and a question about perf impact.

cpp/include/cuspatial/geometry/segment.cuh
isVoid: Applied a find-and-replace for `colienar->collinear`.

cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
isVoid: Interestingly, because the sorting will make `find_and_merge_segment` kernel avoid computing many segments in the inner loop scan, in some cases the performance actually improved. In worst case scenario the performance decreases by 3.4%.

Ref Time: branch-23.08, Cmp Time: this branch.
```
|  FP type  |  num_pairs  |  num_linestring_per_multilinestring  |  num_segments_per_linestring  |   Ref Time |   Ref Noise |   Cmp Time |   Cmp Noise |          Diff |   %Diff |  Status  |
|-----------|-------------|--------------------------------------|-------------------------------|------------|-------------|------------|-------------|---------------|---------|----------|
|    F32    |    1000     |                  1                   |              100              |  11.537 ms |       2.38% |  11.930 ms |       2.33% |    393.774 us |   3.41% |   FAIL   |
|    F32    |    10000    |                  1                   |              100              |  12.108 ms |       2.83% |  12.042 ms |       5.02% |    -66.531 us |  -0.55% |   PASS   |
|    F32    |   100000    |                  1                   |              100              |  28.847 ms |       0.22% |  27.647 ms |       0.28% |  -1200.119 us |  -4.16% |   FAIL   |
|    F32    |   1000000   |                  1                   |              100              | 275.615 ms |       0.05% | 248.693 ms |       0.07% | -26922.072 us |  -9.77% |   FAIL   |
|    F64    |    1000     |                  1                   |              100              |  15.375 ms |       1.36% |  15.559 ms |       1.34% |    183.851 us |   1.20% |   PASS   |
|    F64    |    10000    |                  1                   |              100              |  17.503 ms |       2.20% |  17.775 ms |       1.83% |    272.187 us |   1.56% |   PASS   |
|    F64    |   100000    |                  1                   |              100              |  55.969 ms |       2.20% |  56.857 ms |       1.82% |    887.096 us |   1.58% |   PASS   |
|    F64    |   1000000   |                  1                   |              100              | 431.700 ms |       0.30% | 430.470 ms |       0.29% |  -1230.114 us |  -0.28% |   PASS   |
```

cpp/include/cuspatial/geometry/segment.cuh
```diff
@@ -55,6 +55,17 @@ class alignas(sizeof(Vertex)) segment {
   /// Return the length squared of segment.
   T CUSPATIAL_HOST_DEVICE length2() const { return dot(v2 - v1, v2 - v1); }
 
+  /// Return slope of segment.
+  T CUSPATIAL_HOST_DEVICE slope() { return (v2.y - v1.y) / (v2.x - v1.x); }
```
isVoid, line 59: Degenerate segments?

cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
```diff
@@ -19,20 +19,23 @@
...
 #include <thrust/uninitialized_fill.h>
 
 namespace cuspatial {
 namespace detail {
 
 /**
  * @internal
- * @brief Kernel to merge segments, naive n^2 algorithm.
+ * @brief Kernel to merge segments. Each thread works on one segment space,
+ * with a naive n^2 algorithm.
```
harrism, line 38: Perhaps naive is the wrong word. "brute force"?  Is lower complexity a possibility?

harrism (approved):
Looks good, thanks!

thomcom (approved):

cpp/include/cuspatial/detail/find/find_and_combine_segment.cuh
isVoid: Great question. What's the lower bound of the algorithm complexity here? Given a set of segments, to merge all mergeable segments, one at least needs to look at all segments once, so the lower bound is `O(N)` (N being the number of segments in the set). To achieve that, I assume you need to design some sort of hashing function. When you look at one segment, the hashing function will compute the key and look into the map to find if there is existing segment that can be merged with it. This key would be similar to what we designed in the sorting comparator here - a combination of group id, slope and lower left of the segment.

So that said, the current algorithm isn't naive at all - while there is a nested loop here, the sorting actually already did part of the hashing work like above. The sorting makes sure that the outer loop only run against all other segments once in every mergeable group, all subsequent segments are marked `merged_flag == 1` after the initial run. So the nested loop here is actually on the order of O(N) here. The aggregated complexity is dominated by the sorting, which is O(NlogN).

PR #1212, opened 2023-07-07
Adding Spherical Surface Processing Capability to Point-in-Polygon Queries

### LINKED ISSUE #1202: [FEA]: Adding Spherical Surface Processing Capability to  Point-in-Polygon Queries

Improvement

Low (would be nice)

I would like to work on solving point-in-polygon queries using three-dimensional cartesian coordinates on spherical surfaces. Users may prefer to use that approach for more accurate computations.

#### Coordinates
I have been working on this problem on spherical surfaces. Therefore, I plan to consider that all points are in three-dimensional cartesian coordinates. However, I think proposed algorithm should work with coordinates on geodetic datum, and/or ellipsoidal surfaces. For instance, Google Earth and Oracle Spatial database uses spherical coordinates.

#### Algorithm
I plan to use a slightly updated version of the ray-casting algorithm. Since we have to solve the problem on a spherical surface instead of lines we should use arcs. First, for a test point we presume that there is an arc/curve between the test point and the first point of the query polygon. Then, similar to the ray casting algorithm we are going to count number of intersections between arcs/curves of the query polygon. Below I am going to try to provide high-level pseudo-codes.

```
let PointInPolygon(QueryPolygon Q, PointList P)
    let out = B[|P|] // array of bool for output
    for each p in P parallel do
        let arc = DefineArc(p, Q[0][0])
        let n = 0 // n defines number of intersections
        for each q in Q do // Assume q is an arc
            n += Intersects(arc, q)
        let isLeft = LeftCheck(Q[0], p)
        out[p] = ( (n%2) == isLeft)
    return out
```

In this algorithm, considering the right hand rule, to decide if a point is on the left side of an arc/curve we can use a procedure as follows:

```
let LeftCheck(Arc a, Point p)
    let o = {0, 0, 0} // Origin
    let [p1, p2] = a // begin and end points of the arc
    let p1o = o - p1
    let p2p1 = p2 - p1
    let pp1 = p3 - p1
   
    // Compute orthogonal vector
    let p1o_p2p1 = p1o (x) p2p1 // (x) -> cross product
    // Check the angle between vectors to decide 
    let w = p1o_p2p1 (.) pp1 // (.) -> dot product
    return w > 0
```

Then intersection can be computed as follows:

```
let Intersect(Arc a1, Arc a2){
    let [p1, p2] = a1
    let [p2, p3] = a2
    let p1left = LeftCheck(a2,  p1)
    let p2left = LeftCheck(a2,  p2)
    let p3left = LeftCheck(a1,  p3)
    let p4left = LeftCheck(a1,  p4)
    let alpha = p1 (.) p3
    return (p1left != p2left) && (p3left != p4left) && alpha > 0;
}

```
#### Required changes
I think this update can be implemented with minimal changes

- Add a three-dimensional vector classs
- Resolve assertion problems.
- Add spherical point-in-polygon kernels under existing point-in-polygon kernels.
- Add tests

I have looked at GeoPandas, but seems they do not have this capability on GPUs.

### Additional context

_No response_

ayasar70:
## Description
This PR aims to solve point-in-polygon queries using three-dimensional cartesian coordinates on spherical surfaces. Users may prefer to use that approach for more accurate computations.
This PR "closes #1202 "

## Checklist


cpp/include/cuspatial/geometry/vec_3d.hpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/**
+ * @brief Compute the midpoint of `first` and `second`.
+ */
+template <typename T>
+vec_3d<T> CUSPATIAL_HOST_DEVICE midpoint(vec_3d<T> const& first, vec_3d<T> const& second)
+{
+  return (first + second) * T{0.5};
+}
```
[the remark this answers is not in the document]
ayasar70: Cleaned

cpp/include/cuspatial/detail/geometry/polygon_ref.cuh
```diff
@@ -21,6 +21,7 @@
 #include <cuspatial/iterator_factory.cuh>
 #include <cuspatial/traits.hpp>
 
+#include <cuspatial/geometry/vec_2d.hpp>
```
isVoid, line 24:
```suggestion
#include <cuspatial/geometry/vec_2d.hpp>
#include <cuspatial/geometry/vec_3d.hpp>
```

cpp/tests/point_in_polygon/point_in_polygon_test.cu
```diff
@@ -90,6 +122,24 @@ TYPED_TEST(PointInPolygonTest, OnePolygonOneRing)
                      {false, false, false, false, true, true, true, true});
 }
 
+TYPED_TEST(PointInPolygonTest, OnePolygonOneRingSpherical)
+{
+  CUSPATIAL_RUN_TEST(this->run_spherical_test,
+                     {{-2503.357, -4660.203, 3551.245},
```
isVoid, line 128: What CRS is this coordinate?

isVoid (approved):
Thanks for making this contribution!

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,88 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+ */
+template <typename T>
+__device__ bool is_intersecting(vec_3d<T> const& p1,
+                                vec_3d<T> const& p2,
+                                vec_3d<T> const& p3,
+                                vec_3d<T> const& p4)
+{
+  auto p1left = is_left(p3, p4, p1);
+  auto p2left = is_left(p3, p4, p2);
+  auto p3left = is_left(p1, p2, p3);
```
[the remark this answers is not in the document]
isVoid: Can be punted to future PRs.

cpp/include/cuspatial/geometry/vec_3d.hpp
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * @brief Compute 3D determinant of a 2x2 matrix with column vectors @p a and @p b.
+ */
+template <typename T>
+vec_3d<T> CUSPATIAL_HOST_DEVICE cross_product(vec_3d<T> const& a, vec_3d<T> const& b)
+{
+  return {a.y * b.z - a.z * b.y, a.z * b.x - a.x * b.z, a.x * b.y - a.y * b.x};
+}
+
+}  // namespace cuspatial
```
harrism, line 184: You must be failing clang-format checks if your files are not ending correctly. Make sure you either run `pre-commit run` before committing, or just enable "format on save" for C++ files in your editor.

```suggestion
}  // namespace cuspatial

```

cpp/include/cuspatial/geometry/vec_3d.hpp
```diff
@@ -0,0 +1,184 @@
...
+ * @brief Compute dot product of two 3D vectors.
+ */
+template <typename T>
+T CUSPATIAL_HOST_DEVICE dot(vec_3d<T> const& a, vec_3d<T> const& b)
+{
+  return a.x * b.x + a.y * b.y + a.z * b.z;
+}
+
+/**
+ * @brief Compute 3D determinant of a 2x2 matrix with column vectors @p a and @p b.
```
harrism, line 176:
```suggestion
 * @brief Compute cross product of two 3D vectors. 
 *
 * Equivalent to 3D determinant of a 2x2 matrix with column vectors @p a and @p b.
```

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+ *
+ * Implements another "Crossings test" algorithm by picking the first point
+ * of the query polygon as the reference point
+ *
+ * @tparam T type of coordinate
+ * @tparam PolygonRef polygon_ref type
+ * @param test_point point to test for point in polygon
+ * @param polygon polygon to test for point in polygon
+ * @return boolean to indicate if point is inside the polygon.
+ * `false` if point is on the edge of the polygon.
```
harrism, line 160: Document the behavior, not the return type. Doxygen does that for us.

```suggestion
 * @return True if point is inside the polygon, false if point is outside or on the edge.
```

harrism: Are you absolutely sure this test consistently returns false if the point is on the polygon boundary? This is important to know for DE-9IM. We have had trouble with this with 2D PiP.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+__device__ bool is_left(vec_3d<T> const p1, vec_3d<T> const& p2, vec_3d<T> const& p3)
+{
+  vec_3d<T> po  = {0, 0, 0};
+  auto p1o      = po - p1;
+  auto p1p2     = p2 - p1;
+  auto p1p3     = p3 - p1;
+  auto p1oxp1p2 = cross_product(p1o, p1p2);
+  auto w        = dot(p1oxp1p2, p1p3);
+
+  return w > 0;
```
harrism, line 121: Hmmm, I would have done `dot(cross(p1p2, p1p3), p3) > 0`. Seems more intuitive -- basically pointing right finger down vector p1p2, curling fingers toward p3, and then dot resulting vector with the vector from center of earth to p3 to see if it is positive magnitude.  But in any case, your code can be simplified (no need to subtract from zero).

```suggestion
  auto p1p2     = p2 - p1;
  auto p1p3     = p3 - p1;
  auto p1o_x_p1p2 = cross_product(-p1, p1p2);
  auto w        = dot(p1o_x_p1p2, p1p3);
  return w > 0;
```

harrism: My preferred approach would be a clear one liner like this -- this matches my computer graphics experience much better.
```suggestion
  return dot(cross(p2-p1, p3-p1), p3) > 0;
```

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+ * @param polygon polygon to test for point in polygon
+ * @return boolean to indicate if point is inside the polygon.
+ * `false` if point is on the edge of the polygon.
+ */
+template <typename T, class PolygonRef>
+__device__ inline bool is_point_in_polygon_spherical(vec_3d<T> const& test_point,
+                                                     PolygonRef const& polygon)
+{
+  bool check = false, left_check = false, point_is_within = false;
+  const T radius = sqrt(dot(test_point, test_point));
```
harrism, line 167: What if the polygon points have different radius from the test point? That would mean they are on different concentric spheres...

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+  auto p1p2     = p2 - p1;
+  auto p1p3     = p3 - p1;
+  auto p1oxp1p2 = cross_product(p1o, p1p2);
+  auto w        = dot(p1oxp1p2, p1p3);
+
+  return w > 0;
+}
+
+/** @brief check if two arcs, defined with (p1, p2) and (p3, p4)
+ * intersects (true or not)
```
harrism, line 125:
```suggestion
/** 
@internal
@brief Check if the geodesic arc with endpoints (p1, p2) intersects the arc with endpoints (p3, p4).
```

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+  auto alpha = dot(p1, p3);
+
+  return (p1left != p2left) && (p3left != p4left) && alpha > 0;
+}
+
+/**
+ * @brief Test if a point is inside a polygon on a spherical geometry.
+ *
+ * Implements another "Crossings test" algorithm by picking the first point
+ * of the query polygon as the reference point
```
harrism, line 153:
```suggestion
 * Implements a "crossings test" algorithm by picking the first point of the query 
 * polygon as the reference point
```

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+__device__ inline bool is_point_in_polygon_spherical(vec_3d<T> const& test_point,
+                                                     PolygonRef const& polygon)
+{
+  bool check = false, left_check = false, point_is_within = false;
+  const T radius = sqrt(dot(test_point, test_point));
+  vec_3d<T> check_point;
+  for (auto ring : polygon) {
+    auto ring_points  = multipoint_ref{ring.point_begin(), ring.point_end()};
+    auto num_segments = ring.num_segments();
+    bool closed_ring  = ring_points[0] == ring_points[num_segments];
```
harrism, line 172:
```suggestion
    int closed_ring  = static_cast<int>(ring_points[0] == ring_points[num_segments]);
```

harrism (changes requested):
Thanks for the contribution @ayasar70 !  This is my initial review.  Various doc suggestions and a couple of algorithmic questions/suggestions.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
ayasar70: Updated as single liner.

cpp/include/cuspatial/geometry/vec_3d.hpp
```diff
@@ -0,0 +1,184 @@
...
+ * @addtogroup types
+ * @{
+ */
+
+/**
+ * @brief A generic 3D vector type.
+ *
+ * This is the base type used in cuspatial for both Longitude/Latitude (LonLat) coordinate pairs and
+ * Cartesian (X/Y) coordinate pairs. For LonLat pairs, the `x` member represents Longitude, and `y`
+ * represents Latitude.
```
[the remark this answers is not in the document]
ayasar70: Updated.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,96 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+    vec_3d<T> b       = ring_points[num_segments - closed_ring];
+    size_t s          = 0;
+    for (vec_3d<T> a : ring_points) {
+      if (!check) {
+        left_check   = is_left(b, a, test_point);
+        const auto c = a + b;
+        const auto w = sqrtf(dot(c, c));
```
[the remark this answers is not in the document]
ayasar70: I was testing performance of those two. Seems, I forgot to update that line.

cpp/tests/point_in_polygon/point_in_polygon_test.cu
ayasar70: Those are projections that consider the world as a sphere.

isVoid: Ah I see, I remember you said this is the (X, Y, Z) coordinate using the center of earth as (0, 0, 0). For this algorithm it shouldn't matter how you choose the orthonormal basis as long as they are consistent. It would be nice to mention this in the docstring.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
isVoid: I think the overall assumption of this algorithm is that it's a PiP test on a spherical surface.

harrism: OK. Please document assumptions.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,88 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+                                vec_3d<T> const& p2,
+                                vec_3d<T> const& p3,
+                                vec_3d<T> const& p4)
+{
+  auto p1left = is_left(p3, p4, p1);
+  auto p2left = is_left(p3, p4, p2);
+  auto p3left = is_left(p1, p2, p3);
+  auto p4left = is_left(p1, p2, p4);
+
+  auto alpha = dot(p1, p3);
```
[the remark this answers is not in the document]
harrism: Please cite the origin of this algorithm if you can @ayasar70

harrism:
Just realized CI hasn't been running for this PR!

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
```diff
@@ -99,5 +100,88 @@ __device__ inline bool is_point_in_polygon(vec_2d<T> const& test_point, PolygonR
...
+/** @brief check if p3 is on the left side of the arc that is
+ * defined but p1 and p2
+ * @note: T can be float or double and T4 can be float4 or double4
+ * @param p1: first point of an arc
+ * @param p2: second point af an arc
+ * @param p3: a point to check
+ * @return bool
+ */
+template <typename T>
+__device__ bool is_left(vec_3d<T> const p1, vec_3d<T> const& p2, vec_3d<T> const& p3)
```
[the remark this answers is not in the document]
harrism: > I guess what I didn't understand before is the definition of "left" in a 3 dimension. But this algorithm seems to me pretty clear to me that you are testing if p3 is to one side or the other side of the O-P1-P2 plane.

It's "left" by the "right hand rule".  https://en.m.wikipedia.org/wiki/Right-hand_rule

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
harrism: @ayasar70 can you respond to this one?

harrism (approved):
I had one remaining unanswered question (about points on the boundary of the polygon. Please answer it before merging.

I will go ahead and approve. Nice work, and thank you. I note that this PR is just the beginning of support for this feature -- spherical PiP is only implemented in the C++ header-only API. To be a full feature it needs implementation in the column-based API and exposure in the Python API.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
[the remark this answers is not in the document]
ayasar70: It was an old paper, and I was not able to find it yet. If I cannot find it, I am going to work on the verification.

cpp/include/cuspatial/detail/algorithm/is_point_in_polygon.cuh
ayasar70: Yes. Whenever a point is on an arc of the query polygon, ``is_left`` function should return false since the angle between point and orthogonal vector is going to be 90 degrees. Since we only use multiplication and addition, we should be ok in terms of precision. I am going to double check to see if precision can cause a problem.

ayasar70:
> I had one remaining unanswered question (about points on the boundary of the polygon. Please answer it before merging.
> 
> I will go ahead and approve. Nice work, and thank you. I note that this PR is just the beginning of support for this feature -- spherical PiP is only implemented in the C++ header-only API. To be a full feature it needs implementation in the column-based API and exposure in the Python API.

Thanks! Sorry, I missed the question before. I have replied. Just talked with @isVoid, after this PR I am going to start working on column-based API and Python API.

PR #1214, opened 2023-07-11
Update to `rapidsai/devcontainers:23.08`

trxcllnt:
Updates the devcontainers to v23.08. The latest images now include Python and automatically generate the VSCode workspace on startup, so I was able to simplify the cuSpatial devcontainers a bit.


.devcontainer/conda/devcontainer.json
harrism, line 1: There's a lot of repetition between the several devcontainer.json versions (unified, single, conda, pip, etc.).  Is it possible to factor this into a single file with parameters/modules?

harrism (approved):

thomcom (approved):

ajschmidt8 (approved):

Issue #1208, opened 2023-06-29
Add argument to allow configuring whether to combine result segments in `pairwise_linestring_intersection`


isVoid:
According to @thomcom, `contains` fails with the following use case:
```python
from shapely.geometry import LineString
l1 = LineString([(0, 0), (0.25, 0), (0.75, 0), (1, 0)])
l2 = LineString([(0, 0), (1, 0)])
print(l1.contains(l2))
print(l2.contains(l1))
```

The number of vertices differs currently for both inputs, because `pairwise_linestring_intersection` merges
the result for both cases and in the code path, there's no way distinguishing them.

Issue #1063, opened 2023-04-10
[FEA]: Test and implement all multi (complex) feature types


thomcom:
### Is this a new feature, an improvement, or a change to existing functionality?

New Feature

### How would you describe the priority of this feature request

Critical (currently preventing usage)

### Please provide a clear description of problem you would like to solve.

This follow up issue to #1062 tracks progress toward implementing all feature combinations for complex features: MultiPoint, MultiLineString, and MultiPolygon.

PR #1220, opened 2023-07-17
MultiGeometry tests for MultiPoint and MultiLineString

### LINKED ISSUE #1210: [FEA]: Wrangling MultiLineString.crosses

New Feature

Critical (currently preventing usage)

This issue is that a MultiLineString made from two identical linestrings changes its `crosses` result as opposed to using the linestrings individually.

```
from shapely.geometry import LineString, MultiLineString
l1 = LineString([(0, 0), (1, 1)])
l2 = LineString([(0, 0), (1, 1)])
l = LineString([(0, 1), (0, 0), (1, 0)])
m = MultiLineString([l1, l2])
print(l1.crosses(l))
print(l2.crosses(l))
print(m.crosses(l))
```
```
False
False
True
```

Because I'm writing test generators that build a variety of tests from more basic inputs, I'm seeing this occur a few times in cases where a MultLineString `a` is created via duplicating another LineString `b` twice: `a = MultiLineString([b, b])`.

I'm confused because the problem doesn't occur if the LineStrings that comprise `a` have no intersection:

```
a = LineString([(0, 1), (0, 0), (1, 0)])
b = LineString([(0, 0.75), (1, 0.75)])
c = LineString([(0, 0.25), (1, 0.25)])
display(GeometryCollection([a, b, c]))
print(a.crosses(b))
print(a.crosses(c))
print(a.crosses(MultiLineString([b, c])))
print(a.crosses(MultiLineString([b, b])))
print(a.crosses(MultiLineString([c, c])))
```
image

However, if the LineStrings that make up `a` share a vertex, and that vertex intersects with `l`, then it seems like the LineStrings are combined into a single LineString, then crosses is computed. Anyway I'm just taking some notes here and reporting on my progress.

As you can see, the `crosses` result of a MultiLineString is not taking the `and` or `or` result of crosses on all the individual LineStrings.

_No response_

### Additional context

_No response_

thomcom:
## Description
Partial completion of #1063 
Closes #1210 

This PR has implementations of Multi geometry binary predicates for:
Point-MultiPoint
MultiPoint-Point
MultiPoint-MultiPoint
Point-MultiLinestring
MultiLineString-Point
LineString-MultiLinestring
MultiLinestring-LineString
MultiLineString-MultiLineString

Remaining to be tested and implemented are the MultiPolygon configurations.

I spent a week or two trying to resolve issues that are present in this PR with MultiLineString-LineString predicates, which are discussed here.

(Multi)LineString-(Multi)LineString tests have issues in 4 predicates:

`crosses`

Crosses compares the vertices of intersection with boundary vertices of the lhs and rhs, and is true only when the intersection vertices are not equal to boundary vertices of the lhs or rhs.

There are four fundamental issues in the crosses predicate that I'm still befuddled on:
1. LineString "compaction": GEOS appears to connect sequential endpoints lines in a MultiLineString into a single LineString before computing intersections.
2. LineString equality bug: GEOS appears to produce incorrect results when a MultiLineString is composed of two or more identical LineStrings. This might be a occuring due to LineString "compaction" above, or due to an intersection vertex counting issue.
3. `pairwise_linestring_intersection` segment combination inconsistencies: Sometimes `pairwise_linestring_intersection` compacts a pair of `(0, 0), (0, 0)` intersections into a single `(0, 0)` and sometimes not. Should be fixed by #1207 or #1208

`touches`

The touches predicate appears to have the same fundamental issues as the crosses predicate and fails on the same cases. This is because touches compares vertices of intersection with boundary vertices of the lhs and rhs, and is true when the intersection vertices only equal boundary vertices.

`overlaps` - 2 tests
`within` - 2 tests

Overlaps and within both fail on the same test values based on `.contains`, which has a bug in GEOS that is reported in https://github.com/libgeos/geos/issues/933 and resolved in https://github.com/libgeos/geos/issues/937

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
+++ b/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
@@ -169,10 +169,21 @@ class LineStringPointContains(BinPred):
 
 class LineStringLineStringContainsPredicate(BinPred):
     def _preprocess(self, lhs, rhs):
+        # Typical preprocess:
+        # Flatten all multilinestrings to linestrings
+        # Duplicate the lhs according to geometry_offset sizes
+        # Process the contains function
+        rhs_self_intersection = _basic_intersects_pli(rhs, rhs)
+        rhs_no_segments = _points_and_lines_to_multipoints(
+            rhs_self_intersection[1], rhs_self_intersection[0]
+        )
         pli = _basic_intersects_pli(lhs, rhs)
         points = _points_and_lines_to_multipoints(pli[1], pli[0])
         # Every point in B must be in the intersection
-        equals = _basic_equals_count(rhs, points) == rhs.sizes
+        equals = (
+            _basic_equals_count(points, rhs_no_segments)
+            == rhs_no_segments.sizes
+        )
         return equals
 
 
--- a/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
+++ b/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
@@ -180,10 +180,10 @@ class LineStringLineStringContainsPredicate(BinPred):
     left and right hand side types. """
 DispatchDict = {
     (Point, Point): PointPointContains,
-    (Point, MultiPoint): ImpossiblePredicate,
+    (Point, MultiPoint): MultiPointMultiPointContains,
     (Point, LineString): ImpossiblePredicate,
     (Point, Polygon): ImpossiblePredicate,
-    (MultiPoint, Point): NotImplementedPredicate,
+    (MultiPoint, Point): MultiPointMultiPointContains,
     (MultiPoint, MultiPoint): MultiPointMultiPointContains,
     (MultiPoint, LineString): NotImplementedPredicate,
     (MultiPoint, Polygon): NotImplementedPredicate,
--- a/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
+++ b/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
@@ -28,6 +28,7 @@ from cuspatial.utils.binpred_utils import (
     _false_series,
     _linestrings_to_center_point,
     _open_polygon_rings,
+    _pli_lines_to_multipoints,
     _points_and_lines_to_multipoints,
     _zero_series,
 )
@@ -178,10 +179,10 @@ class LineStringLineStringContainsPredicate(BinPred):
             rhs_self_intersection[1], rhs_self_intersection[0]
         )
         pli = _basic_intersects_pli(lhs, rhs)
-        points = _points_and_lines_to_multipoints(pli[1], pli[0])
+        lines = _pli_lines_to_multipoints(pli)
         # Every point in B must be in the intersection
         equals = (
-            _basic_equals_count(points, rhs_no_segments)
+            _basic_equals_count(lines, rhs_no_segments)
             == rhs_no_segments.sizes
         )
         return equals
--- a/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
+++ b/python/cuspatial/cuspatial/core/binpreds/feature_contains.py
@@ -163,7 +163,7 @@ class MultiPointMultiPointContains(BinPred):
 class LineStringPointContains(BinPred):
     def _preprocess(self, lhs, rhs):
         intersects = _basic_intersects(lhs, rhs)
-        equals = _basic_equals_any(lhs, rhs)
+        equals = _basic_equals_count(lhs, rhs) == rhs.sizes
         return intersects & ~equals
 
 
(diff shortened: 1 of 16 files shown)


python/cuspatial/cuspatial/core/binpreds/feature_contains.py
```diff
@@ -154,31 +156,56 @@ def _preprocess(self, lhs, rhs):
...
+        # compact colinear segments into a single segment, so
+        # vertices in rhs will be lost and not countable.
+        # Consider the following example:
+        # lhs = [(0, 0), (1, 1)]
+        # rhs = [(0, 0), (0.5, 0.5), (1, 1)]
+        # The intersection of lhs and rhs is [(0, 0), (1, 1)].
+        # rhs is contained by lhs if all of the vertices in the
+        # intersection are in rhs. However, the intersection
+        # does not contain the vertex (0.5, 0.5) in the original rhs.
+        rhs_self_intersection = _basic_intersects_pli(rhs, rhs)
```
isVoid, line 185: It feels like you need a separate API that handles linestring merging. Something like:

```c++
void merge_points_on_multilinestring_range(MultiLinestringRange range, OutputIt output);
```

Because calling the self intersection just to remove the additional points are too expensive.

python/cuspatial/cuspatial/core/binpreds/feature_crosses.py
```diff
@@ -43,13 +45,23 @@ def _compute_predicate(self, lhs, rhs, preprocessor_result):
...
-        equals_lhs_count = _basic_equals_count(intersections, lhs)
-        equals_rhs_count = _basic_equals_count(intersections, rhs)
-        equals_lhs = equals_lhs_count != intersections.sizes
-        equals_rhs = equals_rhs_count != intersections.sizes
-        equals = equals_lhs & equals_rhs
-        return equals
+        points = _pli_points_to_multipoints(pli)
+        lines = _pli_lines_to_multipoints(pli)
+        # Optimization: only compute the subsequent boundaries and equalities
+        # of indexes that contain point intersections and do not contain line
```
isVoid, line 51:
```suggestion
        # of indices that contain point intersections and do not contain line
```

python/cuspatial/cuspatial/core/binpreds/feature_crosses.py
```diff
@@ -43,13 +45,23 @@ def _compute_predicate(self, lhs, rhs, preprocessor_result):
...
+        # intersections.
+        lhs_boundary = _lines_to_boundary_multipoints(lhs)
+        rhs_boundary = _lines_to_boundary_multipoints(rhs)
+        lhs_boundary_matches = _basic_equals_count(points, lhs_boundary)
+        rhs_boundary_matches = _basic_equals_count(points, rhs_boundary)
+        lhs_crosses = lhs_boundary_matches != points.sizes
+        rhs_crosses = rhs_boundary_matches != points.sizes
```
isVoid, line 58: It says "none of the points of intersection is in the boundary of the geometry". AKA, there is 0 points in the intersection results that are in the boundary. Shouldn't this be
```
lhs_boundary_matches == 0
```
?

python/cuspatial/cuspatial/core/binpreds/feature_overlaps.py
```diff
@@ -39,6 +42,15 @@ class OverlapsPredicateBase(EqualsPredicateBase):
     pass
 
 
+class LineStringLineStringOverlaps(BinPred):
+    def _preprocess(self, lhs, rhs):
```
isVoid, line 46: A small docstring helps:

```suggestion
    def _preprocess(self, lhs, rhs):
        """Linestring overlaps with another linestring if the overlapping part of 
        two linestring is not the entire section of the linestring.
        """
```
Sounds correct?

python/cuspatial/cuspatial/tests/binpreds/binpred_test_dispatch.py
```diff
@@ -183,6 +183,39 @@ def predicate(request):
         LineString([(0.0, 0.0), (1.0, 1.0)]),
         LineString([(0.5, 0.5), (1.0, 0.1), (-1.0, 0.1)]),
     ),
+    "linestring-linestring-touch-corner": (
+        """
+        x   x
+        |  /
+        | //x
```
isVoid, line 190: I thought only backward slash needs escaping, does forward slash too?

python/cuspatial/cuspatial/tests/binpreds/test_binpred_multigeometry_test_dispatch.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # Forward
+    pred_fn = getattr(lhs, pred)
+    got = pred_fn(rhs)
+    gpd_pred_fn = getattr(gpdlhs, pred)
+    expected = gpd_pred_fn(gpdrhs)
+    assert (got.values_host == expected.values).all()
+
+
+def test_point_multipoint(predicate):  # noqa: F811
```
isVoid, line 106: Why is `noqa` needed? There's no import here.

PR #1233, opened 2023-07-27
cuSpatial Polygonize Notebook

jarmak-nv:
## Description
This PR adds a simple raster -> vector polygonization notebook.

- Each pixel is converted into a square polygon
- A GeoDataFrame is made from the polygons
- Data is read from the Raster and converted to cuDF columns

This notebook serves to be a test-showcase for cuProj. It uses Sentinel-2 imagery and each tile generates about 120M polygons.

Regardless of status on the reviews, [the ](https://github.com/rapidsai/cuspatial/labels/5%20-%20DO%20NOT%20MERGE) has been applied until legal approval. It may be better for us to put this in notebook-contrib as well to prevent `rasterio` from being a notebook dependency.

## Checklist

### DIFF
--- a/notebooks/cuSpatial-Polygonize.ipynb
+++ b/notebooks/cuSpatial-Polygonize.ipynb
@@ -1,0 +1,192 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import rasterio\n",
+    "import cupy as cp\n",
+    "import cuspatial\n",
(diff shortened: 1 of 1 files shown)


harrism:
@jarmak-nv any reason not to get this merged?

jarmak-nv:
> @jarmak-nv any reason not to get this merged?

My hesitation largely stems from testing/dependencies - if we exclude it from notebook tests and don't add the dependencies to our containers then I'm fine with merging, but in doing so we may end up with it not functioning at some point and not realizing it due to the lack of testing.

PR #1284, opened 2023-10-11
Use branch-23.12 workflows.

bdice:
This PR switches back to using `branch-23.12` for CI workflows because the CUDA 12 ARM conda migration is complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ An example command from the Release Selector:
 docker run --gpus all --pull always --rm -it \
     --shm-size=1g --ulimit memlock=-1 --ulimit stack=67108864 \
     -p 8888:8888 -p 8787:8787 -p 8786:8786 \
-    nvcr.io/nvidia/rapidsai/notebooks:23.120-cuda11.8-py3.10
+    nvcr.io/nvidia/rapidsai/notebooks:23.12-cuda11.8-py3.10
 ```
 
 ### Install with Conda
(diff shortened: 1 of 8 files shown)


AyodeAwe (approved):

harrism:
@bdice looks like there is an `rmm-cu12` defined somewhere as `23.10`...

bdice:
@harrism I fixed the remaining 23.10 references on the 23.12 branch (not related to this PR as far as I could see). We'll need a Python codeowner review to merge.

README.md
```diff
@@ -113,7 +113,7 @@ An example command from the Release Selector:
 docker run --gpus all --pull always --rm -it \
     --shm-size=1g --ulimit memlock=-1 --ulimit stack=67108864 \
     -p 8888:8888 -p 8787:8787 -p 8786:8786 \
-    nvcr.io/nvidia/rapidsai/notebooks:23.120-cuda11.8-py3.10
+    nvcr.io/nvidia/rapidsai/notebooks:23.12-cuda11.8-py3.10
```
bdice, line 116: The `update-version.sh` regex for the README had an error, which I fixed.

python/cuproj/pyproject.toml
```diff
@@ -26,7 +26,7 @@ requires = [
 
 [project]
 name = "cuproj"
-version = "23.12.0"
+version = "23.12.00"
```
bdice, line 29: Aligning with other RAPIDS projects -- these lines are supposed to be "full versions" and not "PEP 440 versions." I'm not sure how this typo occurred, because the script is set to do this correctly as far as I can tell.

harrism (approved):

PR #1294, opened 2023-11-09
[TEST] Test cuSpatial with artifacts from rapidsai/rmm#1095

harrism:
Test cuSpatial with changes from rapidsai/rmm#1095

## Checklist


python/cuspatial/cuspatial/tests/spatial/distance/test_hausdorff_distance.py
```diff
@@ -14,7 +14,7 @@ def _test_hausdorff_from_list_of_spaces(spaces):
 def test_empty():
     actual = _test_hausdorff_from_list_of_spaces([])
 
-    expected = cudf.DataFrame([])
+    expected = cudf.DataFrame([], columns=cudf.Index([]))
```
isVoid, line 17: Likely an upstream cuDF change made the empty Dataframe uses a `RangeIndex` as the column names, instead of `cudf.Index`. Since semantically both represent empty column names, I simply change the gold result here to let tests pass.

harrism: OK thanks. Sounds like once everything is passing we will need to update this PR to remove the CI changes and get this fix merged in?

isVoid: I think that's right, otherwise we might have to handle this at release time.

PR #1295, opened 2023-11-15
Fix broken hausdorff test

isVoid:
This PR revises the gold result of hausdorff empty input test to address
an upstream change in cudf.

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/spatial/distance.py
+++ b/python/cuspatial/cuspatial/core/spatial/distance.py
@@ -80,7 +80,7 @@ def directed_hausdorff_distance(multipoints: GeoSeries):
 
     num_spaces = len(multipoints)
     if num_spaces == 0:
-        return DataFrame()
+        return DataFrame([])
 
     if not contains_only_multipoints(multipoints):
         raise ValueError("Input must be a series of multipoints.")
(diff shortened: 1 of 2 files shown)


bdice (approved):

vyasr (approved):
Fix seems fine to me, although I wonder if maybe it should be changed the other way in the long term i.e. have cuspatial return an empty RangeIndex rather than force cudf to produce an empty Index.

harrism:
Indeed that's what chatGPT thinks. https://chat.openai.com/share/43a1b380-e356-4b5e-8c87-08b07003435b

How hard would it be to fix cuSpatial to match cuDF here?

isVoid:
Maybe not that hard - probably need one more argument for `_from_columns` call.

isVoid:
Fundamentally these two calls are different:
```python
In [4]: print(cudf.DataFrame([]).columns)
RangeIndex(start=0, stop=0, step=1)

In [5]: print(cudf.DataFrame._from_columns([], range(0)).columns)
Index([], dtype='object')
```

One is the public API that cuspatial generate the gold result from, and the other is the internal factory function that cuspatial depends upon. I think the proper way forward is that cuspatial shouldn't rely on using internal function from cuDF and attempt to reconstruct the result using public API.

isVoid:
Ooops, I looked at the wrong place - empty input has a shortcut that directly return an empty dataframe. So the actual cause of error is:
```python
In [3]: cudf.DataFrame([]).columns
Out[3]: RangeIndex(start=0, stop=0, step=1)

In [4]: cudf.DataFrame().columns
Out[4]: Index([], dtype='object')
```

isVoid:
Ok I think I prefer this fix - not having to fix the golden result seems a better way forward.

bdice (approved):

harrism (approved):

thomcom (approved):

vyasr:
I agree I prefer this fix. Thanks for indulging my question!

PR #1304, opened 2023-12-05
Remove CUSPATIAL_BUILD_WHEELS and standardize Python builds

vyasr:
## Description

Some minor simplification in advance of the scikit-build-core migration to better align wheel and non-wheel Python builds. Additionally, this PR simplifies handling of the nvcomp targets and their installation.

## Checklist

### DIFF
--- a/python/cuspatial/CMakeLists.txt
+++ b/python/cuspatial/CMakeLists.txt
@@ -52,7 +52,7 @@ if(NOT cuspatial_FOUND)
   include(cmake/Modules/WheelHelpers.cmake)
   # TODO: This install is currently overzealous. We should only install the libraries that are
   # downloaded by CPM during the build, not libraries that were found on the system.  However, in
-  # practice right this would only be a problem is if libcudf was not found but some of the
+  # practice this would only be a problem if libcudf was not found but some of the
   # dependencies were, and we have no real use cases where that happens.
   install_aliased_imported_targets(
     TARGETS cuspatial arrow_shared nvcomp::nvcomp nvcomp::nvcomp_gdeflate nvcomp::nvcomp_bitcomp
--- a/python/cuspatial/CMakeLists.txt
+++ b/python/cuspatial/CMakeLists.txt
@@ -32,13 +32,6 @@ project(
 option(FIND_CUSPATIAL_CPP "Search for existing cuspatial C++ installations before defaulting to local files"
        OFF)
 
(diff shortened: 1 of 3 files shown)


python/cuspatial/CMakeLists.txt
```diff
@@ -32,52 +32,32 @@ project(
...
-  endif()
-
-  # Since there are multiple subpackages of cuspatial._lib that require access to libcuspatial, we place the
-  # library in the cuspatial directory as a single source of truth and modify the other rpaths
-  # appropriately.
-  install(TARGETS cuspatial DESTINATION ${cython_lib_dir})
+  include(cmake/Modules/WheelHelpers.cmake)
+  # TODO: This install is currently overzealous. We should only install the libraries that are
+  # downloaded by CPM during the build, not libraries that were found on the system.  However, in
+  # practice right this would only be a problem is if libcudf was not found but some of the
```
harrism, line 55:
```suggestion
  # practice this would only be a problem if libcudf was not found but some of the
```

harrism (approved):

raydouglass (approved):

python/cuspatial/CMakeLists.txt
```diff
@@ -44,40 +42,22 @@ endif()
...
-  set(_exclude_from_all "")
-  if(CUSPATIAL_BUILD_WHEELS)
-
-    # Statically link cudart if building wheels
-    set(CUDA_STATIC_RUNTIME ON)
-    set(CUSPATIAL_USE_CUDF_STATIC ON)
-    set(CUSPATIAL_EXCLUDE_CUDF_FROM_ALL ON)
-
-    # Always build wheels against the pyarrow libarrow.
-    set(USE_LIBARROW_FROM_PYARROW ON)
```
trxcllnt, line 56: @vyasr This causes cuDF's CMake to [build and install a second libarrow](https://github.com/rapidsai/cuspatial/actions/runs/7118401836/job/19381328724#step:9:719). Don't we need to link to pyarrow's `libarrow.so` due to the internal globals?

vyasr: Yes, this was an error on my part. I went through a couple of iterations of updating the cudf CMake and at one point I had made this detection automatic within cudf. However the other unifications here are desirable, so we just want to specify that variable. Fixed in #1334.

PR #1314, opened 2023-12-12
Use cuda::proclaim_return_type on device lambdas.

bdice:
## Description
This PR makes cuspatial compatible with Thrust 2 by adding `cuda::proclaim_return_type`, but does not upgrade the version of Thrust/CCCL just yet. Currently we use libcudacxx 2.1.0, which makes `cuda::proclaim_return_type` available, but we still use Thrust 1.17.2 which doesn't require device lambdas to have proclaimed return types. This diff is separated out from #1096. I will create a separate PR which contains only packaging changes / version updates for the CCCL 2 migration. This PR is **nonbreaking**, while the second PR will be **breaking**.

## Checklist

### DIFF
--- a/cpp/include/cuspatial/column/geometry_column_view.hpp
+++ b/cpp/include/cuspatial/column/geometry_column_view.hpp
@@ -16,7 +16,6 @@
 
 #pragma once
 
-#include <cuspatial/range/range.cuh>
 #include <cuspatial/types.hpp>
 
 #include <cudf/lists/lists_column_view.hpp>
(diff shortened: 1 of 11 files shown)


harrism (approved):
Thanks @bdice !

isVoid (approved):
Thank you!

harrism:
Looks like some device_vector ctor problems.

bdice:
> Looks like some device_vector ctor problems.

The changes I pulled in were originally targeting CCCL 2 -- this branch can only use the `cuda::proclaim_return_type` pieces because we're still building with Thrust 1.17.2 until I can move all of RAPIDS over. I'll make sure the PR that actually does the upgrade will incorporate those changes to use initializer_list constructors.

PR #1348, opened 2024-02-26
Add support for Python 3.11, require NumPy 1.23+

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/3

This PR adds support for Python 3.11.

It also proposes:

* declaring `numpy` as a runtime dependency, as `cuspatial` does import it directly
* a floor on `NumPy` of `numpy>=1.23` in tests, see https://github.com/rapidsai/build-planning/issues/3#issuecomment-1967952280

## Notes for Reviewers

This is part of ongoing work to add Python 3.11 support across RAPIDS.

The Python 3.11 CI workflows introduced in https://github.com/rapidsai/shared-workflows/pull/176 are *optional*... they are not yet required to run successfully for PRs to be merged.

This PR can be merged once all jobs are running successfully (including the non-required jobs for Python 3.11). The CI logs should be verified that the jobs are building and testing with Python 3.11.

See https://github.com/rapidsai/shared-workflows/pull/176 for more details.


AyodeAwe (approved):

jameslamb:
Looking into adding `numpy>=1.23` as we're pursuing in the other Python 3.11 PRs, I noticed that this project has tests that use `numpy` but doesn't list `numpy` explicitly as a test dependency. Pushed https://github.com/rapidsai/cuspatial/pull/1348/commits/6738d01e0d8accd606cfcd17f1d1341512955cb7 to fix that.

jakirkham:
Thanks James! 🙏

There are definitely required `import`s of NumPy in the library. So agree it should be a dependency

https://github.com/rapidsai/cuspatial/blob/603d76423223613af8ec1300d1ec8216946dc55b/python/cuspatial/cuspatial/utils/column_utils.py#L5

https://github.com/rapidsai/cuspatial/blob/603d76423223613af8ec1300d1ec8216946dc55b/python/cuspatial/cuspatial/core/geoseries.py#L9

https://github.com/rapidsai/cuspatial/blob/603d76423223613af8ec1300d1ec8216946dc55b/python/cuspatial/cuspatial/core/trajectory.py#L3

...

jameslamb:
Oh wow! @jakirkham I hadn't noticed those `numpy` imports outside of the *tests*. Given that, my proposal on the diff here to add `numpy` as a testing dependency isn't enough and it should become a runtime dependency. I can update this PR with that change.

jakirkham:
Yeah AFAICT cuProj only uses NumPy in tests. So this seems right

harrism (approved):

KyleFromNVIDIA (approved):

jakirkham:
Thanks all! 🙏

PR #1351, opened 2024-02-29
Fix Python 3.11 test failures

KyleFromNVIDIA:
## Description
This marks a memory test as `xfail`, which was failing on Python 3.11.

xref: https://github.com/rapidsai/cuspatial/issues/1352
xref: https://github.com/rapidsai/build-planning/issues/3

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/tests/test_geodataframe.py
+++ b/python/cuspatial/cuspatial/tests/test_geodataframe.py
@@ -320,7 +320,10 @@ def test_boolmask(gpdf, df_boolmask):
     assert_eq_geo_df(gi[df_boolmask], cugpdf_back[df_boolmask])
 
 
-@pytest.mark.xfail(reason="Flaky memory test")
+@pytest.mark.xfail(
+    reason="Size discrepancies between Python versions. See "
+    "https://github.com/rapidsai/cuspatial/issues/1352"
+)
 def test_memory_usage(gs):
     assert gs.memory_usage() == 224
     host_dataframe = gpd.read_file(
--- a/python/cuspatial/cuspatial/tests/test_geodataframe.py
+++ b/python/cuspatial/cuspatial/tests/test_geodataframe.py
@@ -320,8 +320,9 @@ def test_boolmask(gpdf, df_boolmask):
     assert_eq_geo_df(gi[df_boolmask], cugpdf_back[df_boolmask])
 
 
+@pytest.mark.xfail(reason="Flaky memory test")
 def test_memory_usage(gs):
-    assert gs.memory_usage() == 228
+    assert gs.memory_usage() == 224
     host_dataframe = gpd.read_file(
         gpd.datasets.get_path("naturalearth_lowres")
     )
--- a/python/cuspatial/cuspatial/tests/test_geodataframe.py
+++ b/python/cuspatial/cuspatial/tests/test_geodataframe.py
@@ -321,7 +321,7 @@ def test_boolmask(gpdf, df_boolmask):
 
 
 def test_memory_usage(gs):
-    assert gs.memory_usage() == 224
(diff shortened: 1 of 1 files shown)


python/cuspatial/cuspatial/tests/test_geodataframe.py
```diff
@@ -321,7 +321,7 @@ def test_boolmask(gpdf, df_boolmask):
 
 
 def test_memory_usage(gs):
```
bdice, line 323: We struggle with tests asserting things about memory usage a lot in RAPIDS. This type of test is very flaky and changes with dependencies, platform, etc.

Does this test serve a critical purpose? Can we remove or skip it? cc: @harrism

bdice: For context, we were getting `224` in Python 3.9 and 3.10, but we get `228` in Python 3.11.

jakirkham: It looks like @thomcom added the test in PR: https://github.com/rapidsai/cuspatial/pull/604

Maybe he knows more about this test and how best to address this issue?

KyleFromNVIDIA: He is currently on leave.

@bdice How have we handled these issues before? Disabling the tests or fixing them?

jakirkham: Let's mark this test as `xfail` (or skip if that doesn't work)

jakirkham:
```suggestion
@pytest.mark.xfail(reason="Size discrepancies between Python versions. See https://github.com/rapidsai/cuspatial/issues/1352")
def test_memory_usage(gs):
```

python/cuspatial/cuspatial/tests/test_geodataframe.py
```diff
@@ -321,7 +321,7 @@ def test_boolmask(gpdf, df_boolmask):
...
-    assert gs.memory_usage() == 224
+    assert gs.memory_usage() == 228
```
jakirkham, line 324: Since this would be `xfail`ed we can revert back to the original code

```suggestion
    assert gs.memory_usage() == 224
```

python/cuspatial/cuspatial/tests/test_geodataframe.py
```diff
@@ -320,6 +320,7 @@ def test_boolmask(gpdf, df_boolmask):
     assert_eq_geo_df(gi[df_boolmask], cugpdf_back[df_boolmask])
 
 
+@pytest.mark.xfail(reason="Flaky memory test")
```
jakirkham: Let's link the issue in the description. That way if someone wonders about this, they know where to go for more info

```suggestion
@pytest.mark.xfail(reason="Flaky memory test. See https://github.com/rapidsai/cuspatial/issues/1352")
```

python/cuspatial/cuspatial/tests/test_geodataframe.py
bdice: `xfail` is a good choice here.

For awareness, in cudf we enable `xfail_strict`. This causes tests to fail if an `xfail`'d test _passes_. (Because it was expected to fail, and did not.) That isn't enabled here, so it's safe to `xfail`. For some versions it will report its status as `xpass` (unexpected pass).

https://github.com/rapidsai/cudf/blob/3b228e2c6d3ec39fcba553c63d53a56760dc1ca6/python/cudf/cudf/tests/pytest.ini#L6

bdice (approved):

jakirkham (approved):

vyasr (approved):

PR #1358, opened 2024-03-07
Enable pytest failures on warnings from cudf (Ensure values being set are casted to the object's type)

mroeschke:
## Description
In 24.04, cudf issues a `FutureWarning` when setting a value to a column would change the data type of the original column. The fix involves casting the original column to the value's type first or vice versa. Could use a second eye on the correct direction to cast if a data type was used intentionally.

Additionally the test suite will now fail when a warning comes from cudf as apart of https://github.com/rapidsai/build-planning/issues/26

## Checklist


isVoid (commented):
Most of the type casts happens in binary predicates. Can @thomcom take a look?

harrism (approved):

harrism:
I am not an expert on Python casts so please get another review before merging.  Also, the title of the issue should probably include something about the casting, since the change to enable warnings as failures is a small part of the PR.

thomcom:
Thanks for your quick updates @isVoid looks good to me.

thomcom (approved):

PR #1360, opened 2024-03-13
Enable pytest failures on warnings on FutureWarnings (Replace deprecated `geopandas.dataset`  module)

mroeschke:
## Description
Continuing the effort in https://github.com/rapidsai/build-planning/issues/26 by enabling `FutureWarnings` and `DeprecationWarnings` to raise errors in the CI.

The primary change here is to replace `geopandas.dataset` usage with the files used (the data in https://github.com/geopandas/geopandas/tree/main/geopandas/tests/data)

## Checklist


python/cuspatial/cuspatial/tests/test_geodataframe.py
```diff
@@ -327,14 +326,12 @@ def test_boolmask(gpdf, df_boolmask):
...
+def test_memory_usage(gs, data_dir):
     assert gs.memory_usage() == 224
-    host_dataframe = gpd.read_file(
-        gpd.datasets.get_path("naturalearth_lowres")
-    )
+    host_dataframe = gpd.read_file(data_dir / "naturalearth_lowres.shp")
     gpu_dataframe = cuspatial.from_geopandas(host_dataframe)
     # The df size is 8kb of cudf rows and 217kb of the geometry column
-    assert gpu_dataframe.memory_usage().sum() == 224945
+    assert gpu_dataframe.memory_usage().sum() == 216793
```
mroeschke, line 334: This test is now essentially the same as `test_memory_usage_large` below. Is it OK to deduplicate?

harrism: It's the same *size* because there are no  geoseries in the geodataframe other than "geometry", right? The logic is different.

harrism (approved):
Thanks for finally fixing the datasets warnings.

python/cuspatial/cuspatial/tests/test_geodataframe.py
mroeschke: Ah yes you're correct. `gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))` also has pop_est, continent, name, iso_a3, and gdp_md_est columns. Is it OK if this test no longer has those columns?

harrism: I don't know. @isVoid @thomcom do you?

AyodeAwe (approved):

bdice (approved):

python/cuspatial/cuspatial/tests/test_geodataframe.py
thomcom: This does change the logic, previously the test was also validating that the cudf series are being read with from_geopandas. I don't think another test verifies this.

mroeschke: Sorry coming back to this now. 

> cudf series

Did you mean geopandas series? It looks like that may be tested in `python/cuspatial/cuspatial/tests/test_from_geopandas.py` and I don't see `cuspatial.from_geopandas` supporting a `cudf.Series`

PR #1374, opened 2024-04-18
Enable import tests

AyodeAwe:
Removes the `--no-test` flag currently set in conda build command and enables testing imports.


harrism:
@AyodeAwe is this still needed?

PR #1346, opened 2024-02-21
Fix `JOIN_POINT_IN_POLYGON_LARGE_TEST_EXP` test

trxcllnt:
## Description
The `JOIN_POINT_IN_POLYGON_LARGE_TEST_EXP` test isn't returning any points because they aren't evenly distributed.

The first commit adds a check to ensure the test fails when no results are returned. The second commit fixes the `JOIN_POINT_IN_POLYGON_LARGE_TEST_EXP` test.

Fixes #1380.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1381, opened 2024-05-07
Fix quadtree spatial join OOMs on large numbers of input polygons

### LINKED ISSUE #890: [BUG]: Strange behavior with quadtree point in polygon
### Version

23.04

Rapids-Compose

### Describe the issue

quadtree pip seems to experience a silent OOM or other error in certain larger data cases. This particular example requires the availability of our demo datasets `taxi2015.csv` and `taxi_zones.zip`. I'm not sure of another way to reproduce it, so I'm including the full example here. It doesn't always appear so I'm just remembering the issue here.

As you can see from the Relevant log output, quadtree returns 3.7m samples when only the first 120 polygons in `zones` are used. If the entire `zones` polygons list (263 polygons) is used, _something_ happens and only 16314 rows are returned. This is unexpected behavior and needs investigation.

```shell
# https://github.com/rapidsai/cuspatial/issues/890
import cuspatial
import cudf
import geopandas
host_zones = geopandas.read_file('taxi_zones.zip')
host_lonlat = host_zones.to_crs(epsg=4326)
zones = cuspatial.from_geopandas(host_lonlat)
taxi2015 = cudf.read_csv('taxi2015.csv')
def quadtree(polygons, points):
    poly_points_x = polygons.polygons.x
    poly_points_y = polygons.polygons.y
    poly_offsets = polygons.polygons.part_offset
    poly_ring_offsets = polygons.polygons.ring_offset
    test_points_x = points.points.x
    test_points_y = points.points.y
    scale = 50
    max_depth = 7
    min_size = 125
    x_max = poly_points_x.max()
    x_min = poly_points_x.min()
    y_max = poly_points_y.max()
    y_min = poly_points_y.min()
    point_indices, quadtree = cuspatial.quadtree_on_points(
        test_points_x,
        test_points_y,
        x_min,
        x_max,
        y_min,
        y_max,
        scale,
        max_depth,
        min_size,
    )
    poly_bboxes = cuspatial.polygon_bounding_boxes(
        poly_offsets, poly_ring_offsets, poly_points_x, poly_points_y
    )
    intersections = cuspatial.join_quadtree_and_bounding_boxes(
        quadtree, poly_bboxes, x_min, x_max, y_min, y_max, scale, max_depth
    )
    polygons_and_points = cuspatial.quadtree_point_in_polygon(
        intersections,
        quadtree,
        point_indices,
        test_points_x,
        test_points_y,
        poly_offsets,
        poly_ring_offsets,
        poly_points_x,
        poly_points_y,
    )
    return polygons_and_points
def make_geoseries_from_lon_lat(lon, lat):
    # Scatter the two columns into one column
    assert len(lon) == len(lat)
    xy = cudf.Series(cp.zeros(len(lon) * 2))
    xy[::2] = lon
    xy[1::2] = lat

    return cuspatial.GeoSeries(cuspatial.core._column.geocolumn.GeoColumn._from_points_xy(xy._column))
dropoffs = make_geoseries_from_lon_lat(
    taxi2015['dropoff_longitude'],
    taxi2015['dropoff_latitude']
)
print(quadtree(zones['geometry'].iloc[0:120], dropoffs))
print(quadtree(zones['geometry'], dropoffs))
```

### Relevant log output

```shell
polygon_index  point_index
0                    0          116
1                    0          387
2                    0          685
3                    0         2607
4                    0         3141
...                ...          ...
3686177            167     12304018
3686178            167     12323531
3686179            167     12351800
3686180            167     12444884
3686181            167     12484251

[3686182 rows x 2 columns]
       polygon_index  point_index
0                  0          116
1                  0          387
2                  0          685
3                  0         2608
4                  0         3142
...              ...          ...
16309              0     12504645
16310              1      7456107
16311              1      7530752
16312              1      7704910
16313              1     11938181

[16314 rows x 2 columns]
```

### Environment details

```shell
rapids-compose 23.04
```

### Other/Misc.

_No response_

trxcllnt:
## Description

Followup to https://github.com/rapidsai/cuspatial/pull/1346.

* Fixes some typos/omissions in types and CMake.
* Adds a new test that OOMs when quadtree_point_in_polygon is passed too many input polygons.
* Fixes quadtree spatial join to handle overflow while counting and more conservatively allocate output buffers.

Fixes #890.

* [Failing test run](https://github.com/rapidsai/cuspatial/actions/runs/8979838628/job/24662981350#step:7:840)
* [Passing test run](https://github.com/rapidsai/cuspatial/actions/runs/8981106226/job/24666403165#step:7:840)

## Checklist

### DIFF
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -23,6 +23,17 @@ add_library(cuspatial_benchmark_common OBJECT
 
 target_compile_features(cuspatial_benchmark_common PUBLIC cxx_std_17 cuda_std_17)
 
+set_target_properties(cuspatial_benchmark_common
+    PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/benchmarks>"
+               INSTALL_RPATH "\$ORIGIN/../../../lib"
+               CXX_STANDARD                        17
+               CXX_STANDARD_REQUIRED               ON
+               CUDA_STANDARD                       17
+               CUDA_STANDARD_REQUIRED              ON
+               POSITION_INDEPENDENT_CODE           ON
+               INTERFACE_POSITION_INDEPENDENT_CODE ON
+)
+
 target_link_libraries(cuspatial_benchmark_common
     PUBLIC benchmark::benchmark
            cudf::cudftestutil
@@ -43,6 +54,10 @@ function(ConfigureBench CMAKE_BENCH_NAME)
     set_target_properties(${CMAKE_BENCH_NAME}
         PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/benchmarks>"
                    INSTALL_RPATH "\$ORIGIN/../../../lib"
+                   CXX_STANDARD 17
+                   CXX_STANDARD_REQUIRED ON
+                   CUDA_STANDARD 17
+                   CUDA_STANDARD_REQUIRED ON
         )
     target_link_libraries(${CMAKE_BENCH_NAME} PRIVATE benchmark::benchmark_main cuspatial_benchmark_common)
     install(
@@ -61,7 +76,11 @@ function(ConfigureNVBench CMAKE_BENCH_NAME)
     ${CMAKE_BENCH_NAME}
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/benchmarks>"
                INSTALL_RPATH "\$ORIGIN/../../../lib"
-  )
+               CXX_STANDARD 17
+               CXX_STANDARD_REQUIRED ON
+               CUDA_STANDARD 17
+               CUDA_STANDARD_REQUIRED ON
+)
   target_link_libraries(
     ${CMAKE_BENCH_NAME} PRIVATE cuspatial_benchmark_common nvbench::main
   )
--- a/cpp/benchmarks/CMakeLists.txt
+++ b/cpp/benchmarks/CMakeLists.txt
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2019-2021, NVIDIA CORPORATION.
+# Copyright (c) 2019-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
+++ b/cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
@@ -24,14 +24,14 @@
 #include <cuspatial/range/multipolygon_range.cuh>
 #include <cuspatial/traits.hpp>
 
+#include <rmm/cuda_device.hpp>
 #include <rmm/device_uvector.hpp>
 #include <rmm/exec_policy.hpp>
 #include <rmm/resource_ref.hpp>
 
 #include <thrust/iterator/permutation_iterator.h>
+#include <thrust/iterator/transform_iterator.h>
 #include <thrust/scan.h>
-
-#include <cstdint>
 
 namespace cuspatial {
 namespace detail {
@@ -57,7 +57,7 @@ struct compute_poly_and_point_indices {
   using IndexType = iterator_value_type<QuadOffsetsIterator>;
 
   inline thrust::tuple<IndexType, IndexType> __device__
-  operator()(IndexType const global_index) const
+  operator()(std::uint64_t const global_index) const
   {
     auto const [quad_poly_index, local_point_index] =
       get_quad_and_local_point_indices(global_index, point_offsets_begin, point_offsets_end);
@@ -118,16 +118,22 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
 
   auto num_poly_quad_pairs = std::distance(poly_indices_first, poly_indices_last);
 
-  auto quad_lengths_iter =
-    thrust::make_permutation_iterator(quadtree.length_begin(), quad_indices_first);
+  auto quad_lengths_iter = thrust::make_transform_iterator(
+    thrust::make_permutation_iterator(quadtree.length_begin(), quad_indices_first),
+    cuda::proclaim_return_type<std::uint64_t>([] __device__(IndexType const& i) -> std::uint64_t {
+      return static_cast<std::uint64_t>(i);
+    }));
 
   auto quad_offsets_iter =
     thrust::make_permutation_iterator(quadtree.offset_begin(), quad_indices_first);
 
+  // Enumerate the point X/Ys using the sorted `point_indices` (from quadtree construction)
+  auto point_xys_iter = thrust::make_permutation_iterator(points_first, point_indices_first);
+
   // Compute a "local" set of zero-based point offsets from number of points in each quadrant
   // Use `num_poly_quad_pairs + 1` as the length so that the last element produced by
   // `inclusive_scan` is the total number of points to be tested against any polygon.
-  rmm::device_uvector<IndexType> local_point_offsets(num_poly_quad_pairs + 1, stream);
+  rmm::device_uvector<std::uint64_t> local_point_offsets(num_poly_quad_pairs + 1, stream);
 
   // inclusive scan of quad_lengths_iter
   thrust::inclusive_scan(rmm::exec_policy(stream),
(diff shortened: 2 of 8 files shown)


cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -118,16 +118,22 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
 
   auto num_poly_quad_pairs = std::distance(poly_indices_first, poly_indices_last);
 
-  auto quad_lengths_iter =
-    thrust::make_permutation_iterator(quadtree.length_begin(), quad_indices_first);
+  auto quad_lengths_iter = thrust::make_transform_iterator(
+    thrust::make_permutation_iterator(quadtree.length_begin(), quad_indices_first),
+    cuda::proclaim_return_type<std::uint64_t>([] __device__(IndexType const& i) -> std::uint64_t {
+      return static_cast<std::uint64_t>(i);
+    }));
```
trxcllnt, line 125: The quadtree length is a uint32 column, but we have to transform them into uint64 values so the `thrust::inclusive_scan` (below) accumulates into uint64s. Changing the output iterator type to uint64 wasn't sufficient to achieve this behavior.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -118,16 +118,22 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
     thrust::make_permutation_iterator(quadtree.offset_begin(), quad_indices_first);
 
+  // Enumerate the point X/Ys using the sorted `point_indices` (from quadtree construction)
+  auto point_xys_iter = thrust::make_permutation_iterator(points_first, point_indices_first);
+
   // Compute a "local" set of zero-based point offsets from number of points in each quadrant
   // Use `num_poly_quad_pairs + 1` as the length so that the last element produced by
   // `inclusive_scan` is the total number of points to be tested against any polygon.
-  rmm::device_uvector<IndexType> local_point_offsets(num_poly_quad_pairs + 1, stream);
+  rmm::device_uvector<std::uint64_t> local_point_offsets(num_poly_quad_pairs + 1, stream);
```
trxcllnt, line 136: `local_point_offsets` is a prefix sum over the number of points in each of the input quadrants, representing the maximum number of possible matches per quadrant. When testing a large number of polygons against a large quadtree, these sums can easily exceed the size of uint32 and overflow.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -136,52 +142,67 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
-  // Allocate the output polygon and point index pair vectors
-  rmm::device_uvector<IndexType> poly_indices(num_total_points, stream);
-  rmm::device_uvector<IndexType> point_indices(num_total_points, stream);
-
-  auto poly_and_point_indices =
-    thrust::make_zip_iterator(poly_indices.begin(), point_indices.begin());
+  // Process in chunks of *at most* 25% of the free available device memory to avoid OOM'ing
+  // when allocating additional output space.
+  auto max_chunk_size =
+    static_cast<std::uint64_t>(rmm::percent_of_free_device_memory(25) / (sizeof(IndexType) * 2));
```
trxcllnt, line 154: I'm open to ideas for a better heuristic here.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -136,52 +142,67 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
-    thrust::copy_if(rmm::exec_policy(stream),
-                    global_to_poly_and_point_indices,
-                    global_to_poly_and_point_indices + num_total_points,
-                    poly_and_point_indices,
-                    detail::test_poly_point_intersection{point_xys_iter, polygons}));
+  // Allocate the output polygon and point index pair vectors
+  rmm::device_uvector<IndexType> poly_indices(0, stream);
+  rmm::device_uvector<IndexType> point_indices(0, stream);
+
+  std::uint64_t num_intersections = 0;
```
trxcllnt, line 162: `num_intersections` isn't numerically required to be a uint64 (since we use uint32 for the output index cols), but using this type avoids needing to cast in the `chunk_size` min calc on line 165.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -136,52 +142,67 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
+
+  for (std::uint64_t chunk_index = 0, chunk_offset = 0; chunk_index < num_chunks; ++chunk_index) {
```
trxcllnt, line 164: `chunk_index` isn't numerically required to be a uint64, but using uint64 avoids needing to cast in the `chunk_size` min calc on line 165.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
trxcllnt, line 162: `num_intersections` isn't numerically required to be a uint64, but using uint64 avoids needing to cast in the `resize()` calls on lines 167-168.

cpp/include/cuspatial_test/geometry_generator.cuh
```diff
@@ -399,8 +401,9 @@ auto generate_multipoint_array(multipoint_generator_parameter<T> params,
                    std::size_t{0},
                    params.num_points_per_multipoints);
 
-  auto engine_x = deterministic_engine(params.num_points());
-  auto engine_y = deterministic_engine(2 * params.num_points());
+  auto golden_ratio = (1 + std::sqrt(T{5})) / 2;
+  auto engine_x     = deterministic_engine(golden_ratio * params.num_points());
+  auto engine_y     = deterministic_engine((1 / golden_ratio) * params.num_points());
```
trxcllnt, line 406: During testing, I observed using `num_points()` and `num_points() * 2` as the seeds produced point positions distributed across two parallel diagonals. Maybe that was an artifact of my point count, but it seems rotating around the golden ratio produces a more even distribution.

cpp/tests/join/quadtree_point_in_polygon_test_oom.cu
```diff
@@ -0,0 +1,140 @@
...
+                                                               rmm::device_uvector<vec_2d<T>>>;
+
+  vec_2d<T> v_min{0.0, 0.0};
+  vec_2d<T> v_max{1'000.0, 1'000.0};
+  T const scale{1.0};
+  std::uint8_t const max_depth{15};
+  std::uint32_t const min_size{10'000};
+
+  std::size_t const num_points{1'000'000};
+  std::size_t const num_polys{24'000};
```
trxcllnt, line 77: 1M points x 24k polys = 24B potential PIP hits, which is enough to trigger the overflow/OOM bug on 24Gb cards like the V100. If we think it's necessary, I can make these values dependent on the amount of GPU memory available at runtime.

cpp/tests/join/quadtree_point_in_polygon_test_oom.cu
```diff
@@ -0,0 +1,140 @@
...
+      vec_2d<T>{(v_max - v_min).x / 2, (v_max - v_min).y / 2},
+      (v_max - v_min).x / 8},
+    this->stream());
+
+  auto multipolygons = multipoly_array.range();
+
+  auto expected_size = thrust::count_if(rmm::exec_policy(this->stream()),
+                                        points.begin(),
+                                        points.end(),
+                                        test_point_in_poly_functor<T>{multipolygons});
```
trxcllnt, line 101: Since we're using so much memory for the actual PIP test, we don't have enough memory left to verify the results of each point with `cuspatial::point_in_polygon`. Since we're already testing accuracy in other tests, I'm instead just verifying that `cuspatial::quadtree_point_in_polygon` returns the number of expected hits.

cpp/include/cuspatial/detail/range/multilinestring_range.cuh
```diff
@@ -74,6 +74,7 @@ template <typename GeometryIterator, typename PartIterator, typename VecIterator
 class multilinestring_range;
 
 template <typename GeometryIterator, typename PartIterator, typename VecIterator>
+CUSPATIAL_HOST_DEVICE
```
trxcllnt, line 77: clangd was complaining about an invalid host overload of a host device function, so I added this here to make it happy.

cpp/benchmarks/CMakeLists.txt
```diff
@@ -61,7 +76,11 @@ function(ConfigureNVBench CMAKE_BENCH_NAME)
     ${CMAKE_BENCH_NAME}
     PROPERTIES RUNTIME_OUTPUT_DIRECTORY "$<BUILD_INTERFACE:${CUSPATIAL_BINARY_DIR}/benchmarks>"
                INSTALL_RPATH "\$ORIGIN/../../../lib"
-  )
+               CXX_STANDARD 17
+               CXX_STANDARD_REQUIRED ON
+               CUDA_STANDARD 17
+               CUDA_STANDARD_REQUIRED ON
+)
```
trxcllnt, line 83: `target_compile_features(... cxx_std_17 cuda_std_17)` doesn't actually set the `-std=c++17` argument, we have to define the `{CXX,CUDA}_STANDARD` properties for that.

cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -363,6 +363,17 @@ class multipoint_array {
...
+   * @brief Copy the offset arrays to host.
+   */
+  auto to_host() const
+  {
+    auto geometry_offsets   = cuspatial::test::to_host<geometry_t>(_geometry_offsets);
+    auto coordinate_offsets = cuspatial::test::to_host<coord_t>(_coordinates);
+
+    return std::tuple{geometry_offsets, coordinate_offsets};
+  }
+
```
trxcllnt, line 376: Adding `multipoint_array::to_host()` to align with `multipolygon_array::to_host()`.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
harrism: Should this comment be in the code?

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
harrism: Should this comment be in the code?

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
harrism: Suggest at least defining a constant for the percent so its easy to find and add configuration for it later if needed.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -136,52 +142,67 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
   // The last element is the total number of points to test against any polygon.
   auto num_total_points = local_point_offsets.back_element(stream);
 
-  // Allocate the output polygon and point index pair vectors
-  rmm::device_uvector<IndexType> poly_indices(num_total_points, stream);
-  rmm::device_uvector<IndexType> point_indices(num_total_points, stream);
-
-  auto poly_and_point_indices =
-    thrust::make_zip_iterator(poly_indices.begin(), point_indices.begin());
+  // Process in chunks of *at most* 25% of the free available device memory to avoid OOM'ing
```
harrism, line 151: Perhaps a higher level comment explaining that the algorithm is chunked to avoid OOMing when the intermediate size is larger.

cpp/include/cuspatial_test/geometry_generator.cuh
harrism: Perhaps this comment should be in the code?

harrism (approved):
Nice improvement. I think adding some of the details from your review comments as actual code comments might be valuable. But I approve.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -148,64 +154,84 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
+    }
+
+    return std::pair{std::move(poly_indices), std::move(point_indices)};
+  };
+
+  try {
+    // First attempt to run the hit test assuming allocating space for all possible intersections
+    // fits into the available memory.
+    return run_quadtree_point_in_polygon(num_total_points);
+  } catch (std::exception const&) {
```
harrism, line 224: You should probably catch the more specific `rmm::out_of_memory` and then throw if any other exception is raised?

harrism (changes requested):
Thanks for fixing! One small request.

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -163,28 +181,57 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
```
trxcllnt, line 224:
```suggestion
  } catch (rmm::out_of_memory const&) {
```

harrism (approved):

trxcllnt:
Not sure what's going on with the `conda-notebook-tests` job. Can we admin merge this?

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
```diff
@@ -136,21 +148,27 @@ std::pair<rmm::device_uvector<IndexType>, rmm::device_uvector<IndexType>> quadtr
...
+  // The largest supported input size for thrust::count_if/copy_if is INT32_MAX.
+  // This functor iterates over the input space and processes up to INT32_MAX elements at a time.
+  std::uint64_t max_points_to_test = std::numeric_limits<std::int32_t>::max();
+  auto count_in_chunks             = [&](auto const& func) {
+    std::uint64_t memo{};
+    for (std::uint64_t offset{0}; offset < num_total_points; offset += max_points_to_test) {
+      memo += func(memo, offset, std::min(max_points_to_test, num_total_points - offset));
+    }
+    return memo;
+  };
```
isVoid, line 166: This looks like a count + reduce to me.. I'm ok with using a for loop in the small scope here. One small suggestion: renaming `func` to `count_func` could increase readability.

isVoid (approved):
One small code suggestion below. Otherwise lgtm!

cpp/include/cuspatial/detail/join/quadtree_point_in_polygon.cuh
trxcllnt: Yeah, this is a special form of `std::reduce`. The difference is mainly in how it passes `min(max, num_left_to_process)` as the `size` argument, which I thought would be more confusing to read if I implemented it as an iterator of `[max, max, .., leftover]`.

PR #1400, opened 2024-07-02
Change _GeoSeriesUtility._from_data(index=) default to None

mroeschke:
## Description
An upstream change in cudf uncovered a bug in `_GeoSeriesUtility._from_data` where `False` was being passed to `cudf.Series(index=)` which is an invalid value. The only valid `index` values are an actual `cudf.Index` or `None`, so setting to `None` as a more appropriate default value

Also updates the location of `assert_eq` which moved in https://github.com/rapidsai/cudf/pull/16063

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1402, opened 2024-07-07
Fix multipolygon geometry iterator.

bdice:
## Description
This PR fixes a bug in the multipolygon geometry iterator. The geometry iterator was returning part iterators by mistake, which masked an issue that we were patching out in rapids-cmake's CCCL. See https://github.com/rapidsai/rapids-cmake/pull/511.

With this fix, we can remove that patch from rapids-cmake's CCCL: https://github.com/rapidsai/rapids-cmake/pull/640

## Checklist

### DIFF
--- a/cpp/include/cuspatial/range/multipolygon_range.cuh
+++ b/cpp/include/cuspatial/range/multipolygon_range.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -115,10 +115,10 @@ class multipolygon_range {
   CUSPATIAL_HOST_DEVICE auto point_end();
 
   /// Return the iterator to the first geometry offset in the range.
-  CUSPATIAL_HOST_DEVICE auto geometry_offset_begin() { return _part_begin; }
+  CUSPATIAL_HOST_DEVICE auto geometry_offset_begin() { return _geometry_begin; }
 
   /// Return the iterator to the one past the last geometry offset in the range.
-  CUSPATIAL_HOST_DEVICE auto geometry_offset_end() { return _part_end; }
+  CUSPATIAL_HOST_DEVICE auto geometry_offset_end() { return _geometry_end; }
 
   /// Return the iterator to the first part offset in the range.
   CUSPATIAL_HOST_DEVICE auto part_offset_begin() { return _part_begin; }
--- a/cpp/include/cuspatial_test/vector_factories.cuh
+++ b/cpp/include/cuspatial_test/vector_factories.cuh
@@ -164,10 +164,26 @@ class multipolygon_array {
   {
     auto [geometry_offsets, part_offsets, ring_offsets, coordinates] = arr.to_host();
 
-    return os << "Geometry Offsets:\n\t{" << geometry_offsets << "}\n"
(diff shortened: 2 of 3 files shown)


cpp/include/cuspatial_test/vector_factories.cuh
```diff
@@ -164,10 +164,26 @@ class multipolygon_array {
   {
     auto [geometry_offsets, part_offsets, ring_offsets, coordinates] = arr.to_host();
 
-    return os << "Geometry Offsets:\n\t{" << geometry_offsets << "}\n"
-              << "Part Offsets:\n\t{" << part_offsets << "}\n"
-              << "Ring Offsets: \n\t{" << ring_offsets << "}\n"
-              << "Coordinates: \n\t{" << coordinates << "}\n";
+    auto print_vector = [&](auto const& vec) {
```
bdice, line 167: Without these changes, I was seeing errors that no `operator<<` overload existed for `thrust::host_vector`.

cpp/include/cuspatial/range/multipolygon_range.cuh
```diff
@@ -115,10 +115,10 @@ class multipolygon_range {
   CUSPATIAL_HOST_DEVICE auto point_end();
 
   /// Return the iterator to the first geometry offset in the range.
-  CUSPATIAL_HOST_DEVICE auto geometry_offset_begin() { return _part_begin; }
+  CUSPATIAL_HOST_DEVICE auto geometry_offset_begin() { return _geometry_begin; }
 
   /// Return the iterator to the one past the last geometry offset in the range.
-  CUSPATIAL_HOST_DEVICE auto geometry_offset_end() { return _part_end; }
+  CUSPATIAL_HOST_DEVICE auto geometry_offset_end() { return _geometry_end; }
```
bdice, line 121: This is the core bug: the `thrust::copy` call using the geometry offsets was getting the wrong size because the wrong iterator was being returned.

This only became visible in the one test case where a multipolygon contained more than one polygon.

cpp/tests/range/multipolygon_range_test.cu
```diff
@@ -1069,7 +1069,7 @@ class MultipolygonRangeOneTest : public MultipolygonRangeTestBase<T> {
   void test_geometry_offsets_it()
   {
     rmm::device_uvector<std::size_t> d_offsets = this->copy_geometry_offsets();
-    auto expected                              = make_device_vector<std::size_t>({0, 1});
+    auto expected                              = make_device_vector<std::size_t>({0, 2});
```
bdice, line 1072: This was a typo that was masked by some kind of silent failure earlier in the tests, I think. The geometry offsets are `{0, 2}`. (I'm not entirely sure how this test ever passed.)

miscco (approved):

harrism (approved):
Thanks for fixing! @isVoid should review too.

bdice:
Thanks for the reviews. There are some failing tests but I think they are related to #1400 and not this PR. I pinged @mroeschke to look into that PR, and I'll merge it into this one once it passes. Then we can remove the CCCL patches from cuDF and rapids-cmake.

cpp/include/cuspatial_test/vector_factories.cuh
trxcllnt: Did thrust remove this overload from host_vector?

trxcllnt (approved):

cpp/include/cuspatial_test/vector_factories.cuh
bdice: I wasn't aware that such an overload ever existed. I don't see that implemented: https://github.com/search?q=repo%3ANVIDIA%2Fcccl%20operator%3C%3C&type=code

Issue #1426, opened 2024-08-01
[BUG]: ZipCodes_Stops_PiP_cuSpatial failing: "data type 'geometry' not understood"


jameslamb:
### Version

24.08

### On which installation method(s) does this occur?

Conda

### Describe the issue

The `ZipCodes_Stops_PiP_cuSpatial` notebook is failing on 24.08 and 24.10, with the most recent nightlies of `cuspatial` and all its dependencies for those versions.

```text
TypeError                                 Traceback (most recent call last)
File /tmp/tmp.q8TobhA9hO/ZipCodes_Stops_PiP_cuSpatial-test.py:302
    299 zipcode_quadtree.set_polygon(d_states, poly_column='geometry')
    301 # Join state and zip code boundaries
--> 302 zipcode_by_state = zipcode_quadtree.point_left_join_polygon(["WKT", "ZCTA5CE10"], ["STUSPS"])
...
File copying.pyx:154, in cudf._lib.copying.gather()
File column.pyx:463, in cudf._lib.column.Column.to_pylibcudf()
File column.pyx:493, in cudf._lib.column.Column.to_pylibcudf()
File types.pyx:284, in cudf._lib.types.dtype_to_pylibcudf_type()
TypeError: data type 'geometry' not understood
```

<details><summary>full stacktrace (click me)</summary>

```text
TypeError                                 Traceback (most recent call last)
File /tmp/tmp.xjCxaX16y8/ZipCodes_Stops_PiP_cuSpatial-test.py:302
    299 zipcode_quadtree.set_polygon(d_states, poly_column='geometry')
    301 # Join state and zip code boundaries
--> 302 zipcode_by_state = zipcode_quadtree.point_left_join_polygon(["WKT", "ZCTA5CE10"], ["STUSPS"])
    304 # Get Californian zipcodes
    305 CA_zipcode = zipcode_by_state[zipcode_by_state.STUSPS == 'CA']

File /tmp/tmp.xjCxaX16y8/ZipCodes_Stops_PiP_cuSpatial-test.py:281, in QuadTree.point_left_join_polygon(self, point_columns, polygon_columns)
    279 points = self.points(point_columns)
    280 polygons = self.polygons(polygon_columns)
--> 281 joined = points.merge(polygons, on=["polygon_index", "point_index"], how="left")
    282 joined = joined.drop(["polygon_index", "point_index"], axis=1)
    283 return cuspatial.GeoDataFrame(joined)

[...]

Issue #1427, opened 2024-08-01
[BUG]: reset_index tests failing: "insert() got an unexpected keyword argument 'name'"


jameslamb:
### Version

24.10

### Describe the issue

On 24.10, several unit tests are failing:

```text
FAILED tests/test_geodataframe.py::test_reset_index[-0-False-False-None] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[-0-False-False-0] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[-0-False-False-1] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[-1-False-False-None] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[-1-False-False-0] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[-1-False-False-1] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[some_lv-0-False-False-None] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[some_lv-0-False-False-0] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[some_lv-0-False-False-1] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[some_lv-1-False-False-None] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[some_lv-1-False-False-0] - TypeError: insert() got an unexpected keyword argument 'name'
FAILED tests/test_geodataframe.py::test_reset_index[some_lv-1-False-False-1] - TypeError: insert() got an unexpected keyword argument 'name'
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/10199784404/job/28218724189?pr=1424))

### Other/Misc.

Not observing this on 24.08, so I believe it's 24.10-specific.

This is happening for both wheels and conda packages, across all combinations of CPU architecture, Python version, and CUDA version tested.

PR #1424, opened 2024-08-01
ensure notebook-testing script fails when notebooks fail

jameslamb:
## Description

The notebook-testing CI job in this repo does not actually cause a loud CI failure if any errors are detected in notebooks. That's because of a `bash` mistake I made in #1407... that PR moved notebook-checking into a function, but didn't add a `set -E` to be sure errors from inside that function were appropriately trapped.

This PR fixes that:

* ensures that notebook failures actually cause CI failures
* fixes 2 typos in `nyc_taxi_years_correlation.ipynb` code
  - *(not caught in #1422 because of this CI script bug)*

Context: https://github.com/rapidsai/cuspatial/pull/1422#pullrequestreview-2212919011

### How I tested this

Originally did not skip any notebooks. Saw the failures in one of the notebooks cause an actual merge-blocking CI failure.

Build link: https://github.com/rapidsai/cuspatial/actions/runs/10199784404/job/28219162698?pr=1424

### Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1429, opened 2024-08-03
Pass columns instead of Series to `cudf.DataFrame` in split-combine workflow

mroeschke:
## Description
closes https://github.com/rapidsai/cuspatial/issues/1426

It appears there was a change in 24.08 that broke a notebook demonstrating a `merge` on two geometry columns. It seems like the merge result tries to reconstruct a `GeoDataFrame` from a `dict[Any, GeoSeries | Series]` but the `Series.index` alignment requires the types to be recognized cudf types (not `"geometry"`)

I don't think this alignment is entirely necessary though since it goes through the `_split_out_geometry_columns`/`_recombine_columns` methods which appears to be used on operations that maintain row ordering so index alignment isn't required.

This PR instead passes a `dict[Any, GeoColumn | Column]` to `cudf.DataFrame._from_data` given that this row ordering is preserved.

(This PR also includes the fix for https://github.com/rapidsai/cuspatial/issues/1427) 

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/geodataframe.py
+++ b/python/cuspatial/cuspatial/core/geodataframe.py
@@ -8,7 +8,6 @@ from geopandas import GeoDataFrame as gpGeoDataFrame
 from geopandas.geoseries import is_geometry_type as gp_is_geometry_type
 
 import cudf
-import cudf.core.column
 from cudf.core.column import as_column
 from cudf.core.copy_types import BooleanMask, GatherMap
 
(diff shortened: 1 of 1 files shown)


harrism:
Is this fix needed in 24.08?

harrism (approved):

mroeschke:
> Is this fix needed in 24.08?

I wouldn't necessarily call this a show-stopping bug, so IMO it's OK to target for 24.10 (especially since the Python unit tests didn't break due to this bug). But I'm happy to backport it if others think it's necessary

jameslamb:
Based on @harrism 's 👍🏻 on https://github.com/rapidsai/cuspatial/pull/1429#issuecomment-2271818491 and our offline conversation about this, seems to me this fix belongs in 24.10.

I'm going to merge this into 24.10, to help with #1424. I also am available to help if we do decide later to backport it to 24.08.

PR #1436, opened 2024-08-08
Internal linkage kernels

robertmaynard:
## Description
Mark all cuspatial CUDA kernels with internal linkage to avoid runtime symbol collisions.

Also updated the CI tests to also verify that we add no new CUDA kernel with external linkage.

## Checklist

### DIFF
--- a/cpp/include/cuspatial/detail/point_linestring_nearest_points.cuh
+++ b/cpp/include/cuspatial/detail/point_linestring_nearest_points.cuh
@@ -54,7 +54,7 @@ template <class Vec2dItA,
           class OffsetIteratorB,
           class OffsetIteratorC,
           class OutputIt>
-void __global__
+void CUSPATIAL_KERNEL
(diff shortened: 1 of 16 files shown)


cpp/include/cuspatial/detail/point_linestring_nearest_points.cuh
```diff
@@ -54,7 +54,7 @@ template <class Vec2dItA,
           class OffsetIteratorB,
           class OffsetIteratorC,
           class OutputIt>
-void __global__
+void CUSPATIAL_KERNEL
```
harrism, line 57: Let's swap these to match all other kernels?

harrism (approved):
One tiny suggestion, but looks good! Thanks.

cpp/include/cuspatial/detail/point_linestring_nearest_points.cuh
robertmaynard: Done

bdice (approved):
Approving CI changes.

PR #1438, opened 2024-08-13
Fix _recombine_columns when either geocolumn or non-geocolumn was empty

mroeschke:
## Description
closes https://github.com/rapidsai/cuspatial/issues/1437

When `GeoDataFrame._split_out_geometry_columns` is called, there may be no geometry columns to split out into 2 DataFrames; therefore, once `GeoDataFrame._recombine_columns`, it assumed the 2 DataFrames were always non-empty. Added a check to bypass this validation if either was empty.

I ran the `ZipCodes_Stops_PiP_cuSpatial` notebook locally and the (non-commented) cells no longer raise an exception.

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/core/geodataframe.py
+++ b/python/cuspatial/cuspatial/core/geodataframe.py
@@ -169,9 +169,9 @@ class GeoDataFrame(cudf.DataFrame):
         The output is meant for GeoDataFrame._from_data.
         """
         if not (
-            geo_columns.empty or
-            data_columns.empty or
-            geo_columns.index.equals(data_columns.index)
+            geo_columns.empty
+            or data_columns.empty
+            or geo_columns.index.equals(data_columns.index)
         ):
             raise ValueError("geo_columns.index must equal data_columns.index")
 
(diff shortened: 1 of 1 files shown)


python/cuspatial/cuspatial/core/geodataframe.py
```diff
@@ -168,7 +168,9 @@ def _recombine_columns(
 
         The output is meant for GeoDataFrame._from_data.
         """
-        if not geo_columns.index.equals(data_columns.index):
+        if not (
+            geo_columns.empty or data_columns.empty
+        ) and not geo_columns.index.equals(data_columns.index):
```
harrism, line 173: De Morgan's rules can clean this up.

```suggestion
        if not (
            geo_columns.empty or
            data_columns.empty or
            geo_columns.index.equals(data_columns.index)
        ):
```

harrism (changes requested):
Looks good, just a logic simplification.

python/cuspatial/cuspatial/core/geodataframe.py
mroeschke: Thanks! Will commit

harrism (approved):
Thanks!

jameslamb:
I'm interested in getting this merged to unblock CI here, so I pushed the `pre-commit` changes CI was complaining about (just those automatically applied by `pre-commit run --all-files`). Figured you wouldn't mind, @mroeschke .

PR #1443, opened 2024-08-22
Drop Python 3.9 support

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/88

Finishes the work of dropping Python 3.9 support.

This project stopped building / testing against Python 3.9 as of https://github.com/rapidsai/shared-workflows/pull/235.
This PR updates configuration and docs to reflect that.

### How I tested this

Checked that there were no remaining uses like this:

```shell
git grep -E '3\.9'
git grep '39'
git grep 'py39'
```

And similar for variations on Python 3.8 (to catch things that were missed the last time this was done).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,8 @@
+# Copyright (c) 2019-2024, NVIDIA CORPORATION.
+
 [tool.black]
 line-length = 79
-target-version = ["py39"]
+target-version = ["py310"]
 include = '\.py?$'
 force-exclude = '''
 /(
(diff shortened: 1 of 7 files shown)


harrism (approved):

pyproject.toml
```diff
@@ -1,6 +1,8 @@
+# Copyright (c) 2019-2024, NVIDIA CORPORATION.
+
```
jakirkham, line 2: Thanks for fixing this! 🙏

jakirkham (approved):
Looks good. Thanks James! 🙏

PR #1444, opened 2024-08-24
[NO MRG] Testing CI

jakirkham:
## Description

An empty commit to test CI

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1441, opened 2024-08-19
Remove NumPy <2 pin, remove proj pin

seberg:
This PR removes the NumPy<2 pin which is expected to work for
RAPIDS projects once CuPy 13.3.0 is released (CuPy 13.2.0 had
some issues preventing the use with NumPy 2).

Also contributes to https://github.com/rapidsai/build-planning/issues/56 by removing the pin on `proj` (which allows this project to work with newer `fmt` / `spdlog`, in sync with conda-forge).

PR #1447, opened 2024-08-28
wheels: dynamically load `libcudf.so` from `libcudf` wheel

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/33.

Proposes the following for `cuspatial` wheels:

* add build and runtime dependencies on `libcudf` wheels
* stop vendoring copies of `libcudf.so`, `libnvcomp.so`, `libnvcomp_bitcomp.so`, and `libnvcomp_gdeflate.so`
  - *(load `libcudf.so` dynamically at runtime instead)*

And other related changes for development/CI:

* combine all `pip install` calls into 1 in wheel-testing scripts
  - *like https://github.com/rapidsai/cudf/pull/16575*
  - *to improve the chance that packaging issues are discovered in CI*
* `dependencies.yaml` changes:
   - more use of YAML anchors = less duplication
   - use dedicated `depends_on_librmm` and `depends_on_libcudf` groups
* explicitly pass a package type to `gha-tools` wheel uploading/downloading scripts

### Benefits of these changes

Unblocks CI in this repo (ref: https://github.com/rapidsai/cuspatial/pull/1444#issuecomment-2308151120, https://github.com/rapidsai/cuspatial/pull/1441#issuecomment-2308149200).

Reduces wheel sizes for `cuspatial` wheels by about 125MB 😁 

| wheel          | size (before)  | size (this PR) |
|:-----------:|-------------:|---------------:|
| `cuspatial` |   146.0M        |   21M               |
| `cuproj `     |       0.9M       |   0.9M              |
|**TOTAL**   |  **146.9M** | **21.9M**        |

*NOTES: size = compressed, "before" = 2024-08-21 nightlies (https://github.com/rapidsai/cuspatial/commit/c60bd4dfc25e62d68cdf520789c6c8bfc577724d), CUDA = 12, Python = 3.11*

<details><summary>how I calculated those (click me)</summary>

```shell
# note: 2024-08-21 because that was the most recent date with
#           successfully-built cuspatial nightlies
#
docker run \
    --rm \
    -v $(pwd):/opt/work:ro \
    -w /opt/work \
    --network host \
    --env RAPIDS_NIGHTLY_DATE=2024-08-21 \
    --env RAPIDS_NIGHTLY_SHA=c60bd4d \
    --env RAPIDS_PR_NUMBER=1447 \
    --env RAPIDS_PY_CUDA_SUFFIX=cu12 \
    --env RAPIDS_REPOSITORY=rapidsai/cuspatial \
    --env WHEEL_DIR_BEFORE=/tmp/wheels-before \
    --env WHEEL_DIR_AFTER=/tmp/wheels-after \
    -it rapidsai/ci-wheel:cuda12.5.1-rockylinux8-py3.11 \
    bash

mkdir -p "${WHEEL_DIR_BEFORE}"
mkdir -p "${WHEEL_DIR_AFTER}"

py_projects=(
    cuspatial
    cuproj
)

for project in "${py_projects[@]}"; do
    # before
    RAPIDS_BUILD_TYPE=nightly \
    RAPIDS_PY_WHEEL_NAME="${project}_${RAPIDS_PY_CUDA_SUFFIX}" \
    RAPIDS_REF_NAME="branch-24.10" \
    RAPIDS_SHA=${RAPIDS_NIGHTLY_SHA} \
        rapids-download-wheels-from-s3 python "${WHEEL_DIR_BEFORE}"

    # after
    RAPIDS_BUILD_TYPE=pull-request \
    RAPIDS_PY_WHEEL_NAME="${project}_${RAPIDS_PY_CUDA_SUFFIX}" \
    RAPIDS_REF_NAME="pull-request/${RAPIDS_PR_NUMBER}" \
        rapids-download-wheels-from-s3 python "${WHEEL_DIR_AFTER}"
done

du -sh ${WHEEL_DIR_BEFORE}/*
du -sh ${WHEEL_DIR_BEFORE}
du -sh ${WHEEL_DIR_AFTER}/*
du -sh ${WHEEL_DIR_AFTER}
```

</details>

Reduces the amount of additional work required to start shipping `libcuspatial` wheels.

### Background

This is part of ongoing work towards packaging `libcuspatial` as a wheel.

relevant prior work:

* packaging `libcudf` wheels: https://github.com/rapidsai/cudf/pull/15483
* consolidating `pip install` calls in CI scripts for `cudf`: https://github.com/rapidsai/cudf/pull/16575
* `cudf` dropping its Arrow library dependency: https://github.com/rapidsai/cudf/pull/16640

### How I tested this

Confirmed in local builds and CI logs that `cudf` is being *found*, not *built*, in `cuspatial` builds.

```text
-- CPM: Using local package cudf@24.10.0
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/10602971716/job/29386288614?pr=1447#step:9:23472))

Built `cuspatial` wheels locally and ran all the unit tests, without issue.

### Checklist


python/cuspatial/pyproject.toml
```diff
@@ -134,6 +136,8 @@ requires = [
     "cmake>=3.26.4,!=3.30.0",
     "cudf==24.10.*,>=0.0.0a0",
     "cython>=3.0.0",
+    "libcudf==24.10.*,>=0.0.0a0",
```
jameslamb, line 139: Putting this somewhere on the diff so it can be threaded.

Just realized I may have missed one detail that wouldn't be caught in CI.

@vyasr do we want something like https://github.com/rapidsai/cudf/blob/925530afe8178b7e788ea1a8d4df4c0eb4d042dc/python/cudf/cudf/__init__.py#L3-L11 near the top of `cuspatial`'s top-level `__init__.py`?

I *think* the answer is yes, so that other `libcudf.so` lying around in the environment aren't loaded first by some other import.

jakirkham: Reading the linked code (and where it lives), it seems just doing `import cudf` would be sufficient

Though please let me know if I'm missing something

bdice (approved):
All seems fine to me. Please double-check `ci/release/update-version.sh`. It may need libcudf and librmm listings.

jameslamb:
> Please double-check `ci/release/update-version.sh`. It may need libcudf and librmm listings.

https://github.com/rapidsai/cuspatial/blob/e323178e7f287bd98771fcafa34d0a9e1462db4e/ci/release/update-version.sh#L50-L53

Yep I don't think any changes are needed there. It already had them because the conda recipes reference those packages.

python/cuspatial/pyproject.toml
vyasr: John is right that the import should be sufficient for now. When you create the libcuspatial wheel in the follow-up, though, you'll end up needing to load libcudf there. [You can see how I did that in the cugraph PR for raft/cugraph-ops](https://github.com/rapidsai/cugraph/pull/4340/files#diff-c2e04ef6cef1505812981008cfadfa1901e4cfb9a537eee3cefaa3e93adaf9a8).

ci/build_wheel.sh
```diff
@@ -5,6 +5,7 @@ set -euo pipefail
 
 package_name=$1
 package_dir=$2
+package_type=$3
```
vyasr, line 8: I guess this is in anticipation of the C++ wheel? I have a very minor preference for sticking that change in the subsequent PR to make the reason more obvious, but now that it's done let's leave it here.

ci/test_wheel_cuspatial.sh
```diff
@@ -5,15 +5,19 @@ set -eou pipefail
 
 mkdir -p ./dist
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
-RAPIDS_PY_WHEEL_NAME="cuspatial_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
 
-# Install additional dependencies
+# install build dependencies for fiona
 apt update
 DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends libgdal-dev
```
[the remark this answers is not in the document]
vyasr: Is the question why we install fiona with `--no-binary`? If so, I do think arm is one of the reasons, but it may not be the only one. @trxcllnt might remember more.

python/cuspatial/CMakeLists.txt
```diff
@@ -44,12 +44,8 @@ if(NOT cuspatial_FOUND)
 
   set(cython_lib_dir cuspatial)
   include(cmake/Modules/WheelHelpers.cmake)
-  # TODO: This install is currently overzealous. We should only install the libraries that are
-  # downloaded by CPM during the build, not libraries that were found on the system.  However, in
-  # practice this would only be a problem if libcudf was not found but some of the
-  # dependencies were, and we have no real use cases where that happens.
   install_aliased_imported_targets(
-    TARGETS cuspatial arrow_shared nvcomp::nvcomp nvcomp::nvcomp_gdeflate nvcomp::nvcomp_bitcomp
+    TARGETS cuspatial
```
vyasr, line 48: Actually, now that we've removed all of these targets we should just remove the WheelHelpers altogether and just do a normal CMake `install`.

vyasr (approved):
Couple of notes, but I don't need to review again.

ci/build_wheel.sh
jameslamb: Technically yes, but I think it'd be desirable even if we weren't adding a C++ wheel in this repo.

In my mind, https://github.com/rapidsai/shared-workflows/pull/209 and https://github.com/rapidsai/gha-tools/pull/105 shipped with `package-type` defaulting to `python` just to avoid having to do updates across all the repos to make progress there.

But that it'd be better to be explicit everywhere (like @bdice mentioned in https://github.com/rapidsai/cudf/pull/15483#discussion_r1725767995), to avoid issues like https://github.com/rapidsai/cudf/pull/16650.

Anyway yeah, we *are* adding C++ wheels here very soon so I'll leave this change in here.

python/cuspatial/CMakeLists.txt
jameslamb: Great point! Done in https://github.com/rapidsai/cuspatial/pull/1447/commits/fe7df5d7c0d55b085be4ae9e2d23de31df2fc169

mroeschke (approved):
Based on the above discussions about importing cudf, Python changes LGTM

ci/test_wheel_cuspatial.sh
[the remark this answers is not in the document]
jameslamb: I don't know the original reason, not having ARM packages seems like a likely candidate.

This PR doesn't change anything about it (it's just showing up in the diff because I'm moving other install-related code around), so I'm going to treat this discussion as non-blocking.

ci/build_wheel.sh
vyasr: Just to clarify: I wasn't meaning the passing of `python` to the GHA tool, just the passing of an argument to our script, i.e. we could have hardcoded `python` in this script and added the `package_type` parameter in a follow-up.

jameslamb: ohhhh ok yeah got it, fair

PR #1448, opened 2024-08-29
cuproj: depend on librmm, not rmm

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/33

`cuproj` does not need the `rmm` Python package... it only needs the RMM headers at build time. This proposes the following changes for `cuproj`:

* dropping the runtime requirement on `rmm` in wheels and conda packages
* switching the build requirement from `rmm` to `librmm` for wheels and conda packages
* removing unnecessary imports in the `test:` environment for conda packages

For more context on these changes, see https://github.com/rapidsai/build-planning/issues/92.

### Benefits of these changes

Faster conda builds (via dropping unnecessary dependencies).

Cheaper (in terms of bandwidth and disk space) installation of wheels and conda packages (via removing an unnecessary runtime dependency).

Reduces a source of network calls (and therefore CI instability) by removing some CPM downloads of RMM.

Before:

```text
-- CPM: Adding package rmm@24.10 (branch-24.10)
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/10618529204/job/29434041322#step:9:16754))

After (this PR):

```text
  -- CPM: Using local package rmm@24.10.0
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/10619138604/job/29436119470?pr=1448#step:9:11256))

### Is this required for `libcuspatial` wheel packaging?

No, it's just a side thing I noticed while working on that. The two are totally independent.

### Checklist


conda/recipes/cuproj/meta.yaml
```diff
@@ -74,17 +74,11 @@ requirements:
...
-    - rmm ={{ minor_version }}
     - cupy>=12.0.0
 
 test:            # [linux64]
   imports:       # [linux64]
     - cuproj     # [linux64]
-  requires:
-    - cupy>=12.0.0
-    - cuspatial ={{ minor_version }}
-    - rmm ={{ minor_version }}
```
jameslamb, line 86: `cuproj`'s tests do use `cuspatial`, but those tests aren't run during conda builds. The only test being run here is `import cuproj`, which shouldn't need any additional test-specific dependencies (if it did, that might indicate a missing `run:` dependency!).

KyleFromNVIDIA (approved):

PR #1450, opened 2024-08-30
Distribute libcuspatial wheels

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/33

Adds `libcuspatial` wheels, and switches `cuspatial` wheels to using them.

### Benefits of these changes

Faster CI runs and smaller total footprint on package repositories (because now `libcuspatial` no longer needs to be compiled once per Python version).

Smaller `cuspatial` wheels.

| whee.  l          | size (before) | size (this PR)    |
|:-------------:|-------------:|----------------:|
| `libcuspatial` |   ---              |      17.0M            |
| `cuspatial`.   |     21.0M        |        4.1M            |
| `cuproj `        |      0.9M        |         0.9M           |
|**TOTAL**      |  **21.9M**    |      **22.0M**     |

*NOTES: size = compressed, "before" = 2024-09-02 nightlies (https://github.com/rapidsai/cuspatial/commit/1544e7bd16f11457436d1fb6a43da0db3aa9af57), CUDA = 12, Python = 3.11*

<details><summary>how I calculated those (click me)</summary>

```shell
docker run \
    --rm \
    -v $(pwd):/opt/work:ro \
    -w /opt/work \
    --network host \
    --env RAPIDS_NIGHTLY_DATE=2024-09-02 \
    --env RAPIDS_NIGHTLY_SHA=1544e7b \
    --env RAPIDS_PR_NUMBER=1450 \
    --env RAPIDS_PY_CUDA_SUFFIX=cu12 \
    --env RAPIDS_REPOSITORY=rapidsai/cuspatial \
    --env WHEEL_DIR_BEFORE=/tmp/wheels-before \
    --env WHEEL_DIR_AFTER=/tmp/wheels-after \
    -it rapidsai/ci-wheel:cuda12.5.1-rockylinux8-py3.11 \
    bash

mkdir -p "${WHEEL_DIR_BEFORE}"
mkdir -p "${WHEEL_DIR_AFTER}"

py_projects=(
    cuspatial
    cuproj
)

for project in "${py_projects[@]}"; do
    # before
    RAPIDS_BUILD_TYPE=nightly \
    RAPIDS_PY_WHEEL_NAME="${project}_${RAPIDS_PY_CUDA_SUFFIX}" \
    RAPIDS_REF_NAME="branch-24.10" \
    RAPIDS_SHA=${RAPIDS_NIGHTLY_SHA} \
[...]

PR #1451, opened 2024-08-30
rearrange dependencies.yaml, fix development instructions

jameslamb:
## Description

Follow-up to #1448. This proposes some `dependencies.yaml` changes that I noticed while working on adding `libcuspatial` wheels in #1450.

I've left inline comments with more details, but in short:

* not including any cuspatial projects in the conda environment files checked into this repo (which are intended to be used to set up a development environment)
* removing an unnecessary conda-only list of dependencies from pyproject.toml-specific config for `cuproj`
* introduction of `depends_on_libcuspatial` and `depends_on_cuproj` lists in `dependencies.yaml`, to reduce duplication

## Checklist


conda/environments/all_cuda-125_arch-x86_64.yaml
```diff
@@ -27,8 +27,6 @@ dependencies:
 - ipython
 - ipywidgets
 - libcudf==24.10.*,>=0.0.0a0
-- libcuspatial-tests==24.10.*,>=0.0.0a0
-- libcuspatial==24.10.*,>=0.0.0a0
```
jameslamb, line 31: See the build instructions in the development docs: https://github.com/rapidsai/cuspatial/blob/1544e7bd16f11457436d1fb6a43da0db3aa9af57/docs/source/developer_guide/build.md#L15-L32

1. they recommend starting by creating a conda environment from one of the files in `conda/environments`
2. they recommend building `libcuspatial` and its tests from source

Those steps are currently broken.

The `libcuspatial-tests` conda package is not distributed anywhere, so that env creation will fail like this:

```text
Solving environment: failed

ResolvePackageNotFound: 
  - libcuspatial-tests[version='24.10.*,>=0.0.0a0']
```

Here, I'm recommending dropping those packages from these files. Similar to how for example you won't find e.g. `libcuml` in these files in the `cuml` repo ([code link](https://github.com/rapidsai/cuml/blob/2f1157047c4db32897bcfe2c477c55c0708a34ec/conda/environments/all_cuda-118_arch-x86_64.yaml)) or `libcuml` in the `cudf` repo ([code link](https://github.com/rapidsai/cudf/blob/2d6758f39592e6296a042eb8e771171c50899013/conda/environments/all_cuda-125_arch-x86_64.yaml)).

dependencies.yaml
```diff
@@ -133,7 +137,6 @@ files:
     includes:
       - test_python_cuproj
       - depends_on_cuspatial
-      - test_cuspatial
```
jameslamb, line 136: This section populates the `[test]` extra for `cuproj` wheels.

The `test_cuspatial` list contained only `conda` dependencies, so it didn't affect wheels in any way. That's why you see this being removed without replacement.

dependencies.yaml
```diff
@@ -391,6 +394,12 @@ dependencies:
         packages:
           - pyproj>=3.6.0,<3.7a0
 
+  depends_on_libcuspatial:
+    common:
+      - output_types: conda
+        packages:
+          - &libcuspatial_unsuffixed libcuspatial==24.10.*,>=0.0.0a0
```
jameslamb, line 401: #1450 will add requirements / pyproject sections to this. They don't need to exist until `libcuspatial` wheels exist.

dependencies.yaml
```diff
@@ -30,24 +29,27 @@ files:
...
   test_python:
     output: none
     includes:
       - cuda_version
+      - depends_on_cuproj
+      - depends_on_cuspatial
       - py_version
       - test_python_cuspatial
       - test_python_cuproj
-      - test_cuspatial
```
jameslamb, line 41: Notice that here, I've taken a list that previously contained `cuproj`, `cuspatial`, and `libcuspatial` with only `cuproj` and `cuspatial`.

That's intentional.... this `test_python` list is used for testing the Python conda packages:

https://github.com/rapidsai/cuspatial/blob/1544e7bd16f11457436d1fb6a43da0db3aa9af57/ci/test_python.sh#L13-L17

Omitting `libcuspatial` here gives us a chance to learn about packaging errors of the form "`cuspatial` is missing a dependency on `libcuspatial" in CI.

harrism:
Are you planning on fixing the instructions, as the title of this PR says?

jameslamb:
> *Are you planning on fixing the instructions, as the title of this PR says*

The changes to `dependencies.yaml` do fix the instructions... now following this step in the docs will succeed:

https://github.com/rapidsai/cuspatial/blob/0e9d36cff2c8e4fcfdd9e88e7ec3fa07a6db6184/docs/source/developer_guide/build.md?plain=1#L21

where previously that would have failed because `libcuspatial-tests` is not distributed anywhere.

Sorry if the wording in the title here was unclear, I can see how it makes it seems like this might *modify* those instructions.

dependencies.yaml
```diff
@@ -573,13 +595,6 @@ dependencies:
       - output_types: conda
         packages:
           - libcuspatial-tests==24.10.*,>=0.0.0a0
```
bdice, line 597: I’m not sure if I like this structure where we have a dependency list that defines libcuspatial-tests. In libcudf tests we list the libcudf-tests package explicitly rather than using a dependency list. https://github.com/rapidsai/cudf/blob/7018a33be752da9363db5431560d8d12bf378920/ci/test_cpp_common.sh#L34

The *-tests conda packages are kind of special in that they are only CI artifacts and thus never available from any “proper” conda channels, only local channels of CI built artifacts. For that reason I am hesitant to use RDFG to define them

bdice: Hmmmm. The alternate proposal I had in mind (removing the test_libcuspatial dependency list and installing libcuspatial-tests manually) would not make sense for this repo, since we’ve moved to the “single step solve” for C++ tests in cuspatial.  https://github.com/rapidsai/cuspatial/blob/0e9d36cff2c8e4fcfdd9e88e7ec3fa07a6db6184/ci/test_cpp.sh#L17

Maybe we do have the right approach and no changes are needed…

bdice (approved):
Approving with a question about best practices… which I partly answered for myself. Happy to hear feedback but I think no changes are required.

dependencies.yaml
jameslamb: yep exactly, the use of that `--prepend-channel` pattern here is why I left this as-is.

PR #1467, opened 2024-10-07
xfail Python tests dependent on invalid Polygons

mroeschke:
## Description
Appears the `gs` fixture in the `conftest.py` contains invalid polygon (the holes are beyond shell), and geos 3.13 now does not consider the same invalid polygon equivalent https://github.com/libgeos/geos/issues/1177.

The `gs` fixture and the `polys` fixture (which is hard-coded) need to both be updated, but I wasn't able to get a quick-working fix (I'm not familiar with this area). This PR just xfails the effected tests.

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/tests/conftest.py
+++ b/python/cuspatial/cuspatial/tests/conftest.py
@@ -29,6 +29,8 @@ def gs():
     g8 = Polygon(
         ((35, 36), (37, 38), (39, 40), (41, 42)),
     )
+    # TODO: g9, g10, g11 are invalid
+    # https://github.com/libgeos/geos/issues/1177
     g9 = MultiPolygon(
         [
             (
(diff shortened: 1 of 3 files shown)


vyasr (approved):
Approving to unblock the 24.10 release. It would be good to get confirmation from a cuspatial developer more familiar with these tests to determine if these tests are important and indicative of some real issue that we need to address, but I don't think it's worth having cuspatial hold up the release so if needed we can release a hotfix.

python/cuspatial/cuspatial/tests/conftest.py
```diff
@@ -29,6 +29,8 @@ def gs():
     g8 = Polygon(
         ((35, 36), (37, 38), (39, 40), (41, 42)),
     )
+    # TODO: g9, g10, g11 are invalid
+    # https://github.com/libgeos/geos/issues/1177
```
harrism, line 33: These are multipolygons, not polygons, so I don't think they are invalid!

PR #1469, opened 2024-10-08
remove unused variables in build_docs script

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/106

This project doesn't need any of the changes that are being made as part of that issue across most other repos... it's already doing all of its `pip` and `conda` installs in CI in single installs (which is strict and safe) 🎉 

This small PR just makes its handling of environment variables in `ci/build_docs.sh` consistent with other RAPIDS repos, in the interest of keeping those scripts looking similar so it's easy to apply changes across all RAPIDS projects. ref: https://github.com/rapidsai/cudf/pull/17013#discussion_r1792163289

## Checklist

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -21,6 +21,8 @@ conda activate docs
 
 rapids-print-env
 
+export RAPIDS_VERSION="$(rapids-version)"
+export RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
 export RAPIDS_DOCS_DIR="$(mktemp -d)"
 
 rapids-logger "Build cuSpatial CPP docs"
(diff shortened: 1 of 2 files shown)


harrism (approved):

jameslamb:
I realized that `RAPIDS_VERSION` and `RAPIDS_VERSION_MAJOR_MINOR` need to be set in the environment by the `build_docs` script, because they're used for Doxygen.

https://github.com/rapidsai/cuspatial/blob/e002dc8adbfcdc6197e7357c1f96f839af64ec3a/cpp/doxygen/Doxyfile#L41

https://github.com/rapidsai/cuspatial/blob/e002dc8adbfcdc6197e7357c1f96f839af64ec3a/cpp/doxygen/Doxyfile#L2174

Just pushed a commit restoring them. I think the small diff left in this PR is still worth it, to replace a hard-coded version number with one that's driven by the `VERSION` file.

msarahan (approved):

PR #1473, opened 2024-10-11
build wheels without build isolation

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/108

Contributes to https://github.com/rapidsai/build-planning/issues/111

Proposes building `libcuspatial` wheels with `--no-build-isolation`, to improve the rate of `sccache` cache hits and therefore reduce CI times.

Also proposes printing `sccache` stats to CI logs after wheel and conda builds.

### Checklist

### How I tested this

Ran a first build with these changes to populate the cache, then ran again... saw a 100% cache hit rate and the actual wheel-building part of a `wheel-build-libcuspatial` job take only 37 seconds.

```text
Successfully built libcuspatial-cu11
Compile requests                     28
Compile requests executed            28
Cache hits                           28
Cache hits (c [gcc])                  2
Cache hits (c++ [gcc])                2
Cache hits (cuda [nvcc])             24
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/11300325066/job/31433017812?pr=1473#step:9:393))


ci/build_wheel.sh
```diff
@@ -33,9 +33,49 @@ rapids-generate-version > ./VERSION
...
+rapids-logger "Building '${package_name}' wheel"
+python -m pip wheel \
+    -w dist \
+    -v \
+    --no-build-isolation \
+    --no-deps \
+    --disable-pip-version-check \
+    .
+
+sccache --show-adv-stats
```
jameslamb, line 78: Anticipating a review comment like "is this left over from debugging?"... I've intentionally left this in. I think it's generally useful to print the `sccache` stats in these CI logs, as they could provide a hint when you experience issues like CI timeouts or out-of-memory.

This only adds 27 lines to the logs, like this:

```text
Compile requests                     28
Compile requests executed            28
Cache hits                           28
Cache hits (c [gcc])                  2
Cache hits (c++ [gcc])                2
Cache hits (cuda [nvcc])             24
Cache misses                          0
Cache timeouts                        0
Cache read errors                     0
Forced recaches                       0
Cache write errors                    0
Compilation failures                  0
Cache errors                          0
Non-cacheable compilations            0
Non-cacheable calls                   0
Non-compilation calls                 0
Unsupported compiler calls            0
Average cache write               0.000 s
Average compiler                  0.000 s
Average cache read hit            0.077 s
Failed distributed compilations       0
Cache location                  s3, name: rapids-sccache-east, prefix: /
Version (client)                0.7.7
```

ci/build_wheel.sh
```diff
@@ -33,9 +33,49 @@ rapids-generate-version > ./VERSION
...
+  # then ensures 'cuspatial' wheel builds always use the 'libcuspatial' just built in the same CI run.
+  RAPIDS_PY_WHEEL_NAME="libcuspatial_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 cpp /tmp/libcuspatial_dist
+  echo "libcuspatial-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo /tmp/libcuspatial_dist/libcuspatial_*.whl)" > /tmp/constraints-build.txt
+fi
+
+rapids-logger "Generating build requirements"
+declare -r matrix_selectors="cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION};cuda_suffixed=true"
+
+rapids-dependency-file-generator \
+  --output requirements \
```
KyleFromNVIDIA, line 49: This approach means we're committing to always specifying `requirements` in `dependencies.yaml` alongside `pyproject`. I've been guilty of neglecting this in the past, and I suspect others are too. (This approach also means people may start neglecting `pyproject` and we wouldn't notice.) Plus, it means we're duplicating all of `pip`'s build dependency installation logic. Is there any way we can keep the duplication to a minimum?

ci/build_wheel.sh
msarahan: Yes, this is good IMO. I'm also going to add this kind of thing into the telemetry.

msarahan (approved):

ci/build_wheel.sh
KyleFromNVIDIA: Ultimately, I think the correct approach is adding a `--build-isolation-dir` argument to `pip wheel` so that we can control the prefix for `sccache` purposes, but that obviously won't work in the short term.

jameslamb: At a minimum, I think we'll have to install the build backend (`rapids-build-backend` + `setuptools` / `scikit-build-core`) separately like this for this to work, because the build backend is needed to determine all the other build-time dependencies. That should simplify the scripts and alleviate most of the "people may start neglecting `pyproject`" worry, I hope.

It should be possible to then have the rest of the build dependencies (`librmm`, `libcudf`, etc.) still be determined by `rapids-build-backend` and no require pre-installation. Like this:

```shell
# before
pip wheel --no-build-isolation --no-deps .

# after
pip wheel --no-build-isolation .
```

Let me try that.

ci/build_wheel.sh
vyasr: I agree, I want us to start printing sccache in more jobs so that we can proactively identify these kinds of issues.

ci/build_wheel.sh
vyasr: Yeah, all the concerns are valid but also not new. These are essentially all of the reasons why I preferred to use build isolation when setting everything up, but I don't see a way around this right now. 

> Ultimately, I think the correct approach is adding a --build-isolation-dir argument to pip wheel so that we can control the prefix for sccache purposes, but that obviously won't work in the short term.

I would like this to happen, but I think it will be an uphill battle. Repacking dynamic libraries in wheels is still something that is pretty uncommon outside of auditwheel-like bundling use cases, and repackaging headers is even rarer. I'm not saying that we shouldn't fight this battle, but I expect that it will take a long time before this changes.

vyasr:
I found something that might be a reasonable workaround described in https://github.com/pypa/build/issues/655. Basically, create a new temp dir for _all_ of Python's temporary file creation and assume that the build backend is going to use that. It may not work if the build backend doesn't do that, but I think it's reasonable to assume that any sane build backend will use `tempfile` APIs for temporary file creation (and certainly we can make sure of that for SKBC). Maybe give that a shot and see if that works for you.

jameslamb:
@vyasr that's a good idea I hadn't considered! Thanks for the links.

I just tried this idea locally, and unfortunately I don't think it'll work for what we're trying to do here.

Setting `TMPDIR` does caused the build environment virtual env to be created inside a specific top-level directory.

```shell
mkdir -p /tmp/wheel-build

TMPDIR=/tmp/wheel-build \
  python -m pip wheel \
      -w dist \
      -v \
      --no-deps \
      --disable-pip-version-check \
      .
```

That causes `pip wheel` to create a build env in a location like `/tmp/wheel-build/pip-build-env-{name}/`.

```text
...
Run Build Command(s): /tmp/wheel-build/pip-build-env-ih3w97sd/normal/lib/python3.12/site-packages/ninja/data/bin/ninja -v -j 16
...
[19/25] sccache /usr/local/cuda/bin/nvcc ... \
    -I/tmp/wheel-build/pip-build-env-ih3w97sd/normal/lib/python3.12/site-packages/libcudf/include/libcudf/rapids ...
```

Unfortunately, that `{name}` part is randomly generated and I don't think it's controllable through environment variables or anything in `pip`'s public API.

Specifically, `pip` is doing this to generate that:

```python
path = os.path.realpath(tempfile.mkdtemp(prefix=f"pip-{kind}-"))
```

([pip code link](https://github.com/pypa/pip/blob/ec5faeac4ef6bae97df0e779566ceb2b0de89d3f/src/pip/_internal/utils/temp_dir.py#L177))

Here's the code path in `tempfile.mkdtemp()` that generates it.

1. `tmpfile.mkdtemp()` calls `tmpfile._get_candidates_names()`

<details><summary>code snippet (click me)</summary>

```python
def mkdtemp(suffix=None, prefix=None, dir=None):
    prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)

    names = _get_candidate_names()

    for seq in range(TMP_MAX):
        name = next(names)
        file = _os.path.join(dir, prefix + name + suffix)
```

([code link](https://github.com/python/cpython/blob/e99650b80ace3893c2a80b3f2a4aca99cb305191/Lib/tempfile.py#L375))

</details>

2. `tempfile._get_candidate_names()` calls `tmpfile._RandomNameSequence()`

<details><summary>code snippet (click me)</summary>

```python
_name_sequence = None

def _get_candidate_names():
    """Common setup sequence for all user-callable interfaces."""

    global _name_sequence
    if _name_sequence is None:
        _once_lock.acquire()
        try:
            if _name_sequence is None:
                _name_sequence = _RandomNameSequence()
        finally:
            _once_lock.release()
    return _name_sequence
```

([code link](https://github.com/python/cpython/blob/e99650b80ace3893c2a80b3f2a4aca99cb305191/Lib/tempfile.py#L227-L240))

</details>

3. `tempfile._RandomNameSquence()` calls `random.Random.choices()` to generate 8-character random strings

<details><summary>code snippet (click me)</summary>

```python
from random import Random as _Random

class _RandomNameSequence:
    characters = "abcdefghijklmnopqrstuvwxyz0123456789_"

    @property
    def rng(self):
        cur_pid = _os.getpid()
        if cur_pid != getattr(self, '_rng_pid', None):
            self._rng = _Random()
            self._rng_pid = cur_pid
        return self._rng

    def __iter__(self):
        return self

    def __next__(self):
        return ''.join(self.rng.choices(self.characters, k=8))
```

([code link](https://github.com/python/cpython/blob/e99650b80ace3893c2a80b3f2a4aca99cb305191/Lib/tempfile.py#L132-L154))

</details>

Unless it's possible to influence the seed for `random.Random()` from outside of the `pip wheel` process (e.g. via environment variables), I don't think we can get `pip wheel` to use an identical build directory on every run.

According to https://docs.python.org/3/library/random.html#random.seed

> `random.seed(a=None, version=2)`
> *Initialize the random number generator.*
>
> *If a is omitted or None, the current system time is used. If randomness sources are provided by the operating system, they are used instead of the system time (see the [`os.urandom()`](https://docs.python.org/3/library/os.html#os.urandom) function for details on availability).*

vyasr:
It's definitely not worth doing anything that requires trying to hack random number generation. At that point just turning off build isolation is the right move IMHO.

jameslamb:
> *It's definitely not worth doing anything that requires trying to hack random number generation. At that point just turning off build isolation is the right move IMHO.*

I agree, I think we should move forward with this.

@KyleFromNVIDIA can you review the recent discussion here and approve if you don't have other suggestions?

KyleFromNVIDIA (approved):

ci/build_wheel.sh
```diff
@@ -33,9 +33,49 @@ rapids-generate-version > ./VERSION
...
+
+if [[ "${package_name}" == "cuspatial" ]]; then
+  # Downloads libcuspatial wheel from this current build,
+  # then ensures 'cuspatial' wheel builds always use the 'libcuspatial' just built in the same CI run.
+  RAPIDS_PY_WHEEL_NAME="libcuspatial_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 cpp /tmp/libcuspatial_dist
+  echo "libcuspatial-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo /tmp/libcuspatial_dist/libcuspatial_*.whl)" > /tmp/constraints-build.txt
+fi
+
+rapids-logger "Generating build requirements"
+declare -r matrix_selectors="cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION};cuda_suffixed=true"
```
vyasr, line 46: Minor nit: Is using `declare` necessary? If we want to start making these kinds of improvements there are probably lots of variables in our CI scripts that should be local and read-only, and perhaps other things too.

ci/build_wheel.sh
```diff
@@ -33,9 +33,49 @@ rapids-generate-version > ./VERSION
...
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 
+touch /tmp/constraints-build.txt
+
+if [[ "${package_name}" == "cuspatial" ]]; then
+  # Downloads libcuspatial wheel from this current build,
+  # then ensures 'cuspatial' wheel builds always use the 'libcuspatial' just built in the same CI run.
+  RAPIDS_PY_WHEEL_NAME="libcuspatial_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 cpp /tmp/libcuspatial_dist
+  echo "libcuspatial-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo /tmp/libcuspatial_dist/libcuspatial_*.whl)" > /tmp/constraints-build.txt
+fi
```
vyasr, line 43: It's not so important for cuspatial, but to establish a pattern that will translate better to all of the other repositories, what about putting the download back into `build_wheel_cuspatial.sh` and passing a constraint file through instead? Then this script would be responsible for creating an empty constraint file.

Alternatively, leave the logic here but have the script accept variables to indicate what packages need to be downloaded from s3 and added to constraints. Either way, we want a pattern that will support longer chains (e.g. libcugraph->pylibcugraph->cugraph).

vyasr (changes requested):

ci/build_wheel.sh
jameslamb: No not strictly necessary, I was just trying to be explicit about this being read-only.

If you think it's too weird here I'll happily remove it... I definitely don't want to bring "more carefully mark the visibility of all variables in the `ci/` shell scripts" into the scope of this.

vyasr: I don't have a strong preference, it just seemed a bit odd to suddenly introduce this when we aren't careful about this in any of our other scripts. I'm fine keeping it. The only real downside is that it may not be familiar to all readers. I have no interest in increasing the scope of this PR in that direction either. The other change request is much more pertinent for getting this merged.

ci/build_wheel.sh
jameslamb: I've moved the download back in https://github.com/rapidsai/cuspatial/pull/1473/commits/29b005edb4d96a826ce8e6b03cabcb6ab5998124. 

That pattern is already supporting "longer chains" like `libcudf -> pylibcudf -> cudf` ([cudf code link](https://github.com/rapidsai/cudf/blob/768fbaa28033446dc899872b3c94213c75bc1a98/ci/build_wheel_cudf.sh#L10-L21)). Keeping that as-is will keep the diffs for these PRs smaller... I should have just done that from the beginning.

----

However ... I don't understand the suggestion "this script would be responsible for creating an empty constraint file" though... could you explain that?

`build_wheel_cuspatial.sh` is populated in a constraints file and populated environment variable `PIP_CONSTRAINT` with the path to it, so every `pip {command}` will respect that.

Other `build_wheel_{project}.sh` scripts are not setting that env variable, and so not using `pip`'s constraints mechanism.

What's the purpose of generating an empty constraints file?

ci/build_wheel.sh
jameslamb: I removed the `declare -r` in https://github.com/rapidsai/cuspatial/pull/1473/commits/29b005edb4d96a826ce8e6b03cabcb6ab5998124, in the interest of consistency with the other scripts and out of a desire to not take on updating all of the references like this.

jameslamb:
Looks like CI here is breaking because of some upstream changes in `cudf`.

> join_quadtree_and_bounding_boxes_test.cpp:(.text._ZN53JoinQuadtreeAndBoundingBoxesErrorTest_test_empty_TestIdE8TestBodyEv[_ZN53JoinQuadtreeAndBoundingBoxesErrorTest_test_empty_TestIdE8TestBodyEv]+0x430): undefined reference to `cudf::test::detail::expect_tables_equal(cudf::table_view, cudf::table_view)

([build link](https://github.com/rapidsai/cuspatial/actions/runs/11335727428/job/31525723610?pr=1473#step:7:1109))

bdice:
I merged in the upstream to get CI fixes from #1475.

ci/build_wheel.sh
vyasr: The version of the code here was already generating an empty constraints file by default. You had the `touch  /tmp/constraints-build.txt`, and then you replaced it in the `package_name == cuspatial` block. I assumed that you did this because in the libcuspatial build there were no constraints but you didn't want to modify the pip commands to include or exclude the constraint argument based on the name, so you chose to use an empty constraints file instead. Was I misunderstanding? I was attempting to give you a way to maintain that structure. Exporting the `PIP_CONSTRAINT` variable is cleaner though since then you always invoke pip without the `--constraint` argument. The other benefit of the env var is that it also works with build environment construction, although that is no longer relevant given the nature of this PR.

The point I was trying to make about longer chains is that it makes less sense to me to have the generic `build_wheel.sh` script have a bunch of `if package_name ==...` logic when you could instead have the `build_wheel_${package_name}.sh` have that logic and pass a standardized input to the `build_wheel.sh` script. Now that you've put it back I think that stays simpler.

vyasr (approved):
I opened https://github.com/rapidsai/dependency-file-generator/issues/114 as a suggestion for simplifying these kinds of scripts, but feel free to proceed in whatever order you prefer. This LGTM now.

vyasr:
Temporarily blocking merge pending offline discussion about whether we want to only turn off build isolation for C++ wheel builds so that Python builds can still test the build isolation code path via RBB and ensure that we get things mostly right.

ci/build_wheel.sh
jameslamb: Ok I understand, thanks. Will leave it as-is (using `PIP_CONSTRAINT` env variable set in `ci/build_wheel_{project}.sh` instead of if-else blocks in `ci/build_wheel.sh`).

jameslamb:
> *I opened https://github.com/rapidsai/dependency-file-generator/issues/114 as a suggestion for simplifying these kinds of scripts, but feel free to proceed in whatever order you prefer.*

Ok, I'll try this right now. If it takes too long, I'll abandon it for now and proceed with 2 `rapids-dependency-file-generator` calls.

ci/build_wheel_libcuspatial.sh
```diff
@@ -3,4 +3,31 @@
...
 set -euo pipefail
 
+git clone \
+  --branch multiple-file-keys \
+  https://github.com/jameslamb/dependency-file-generator.git \
+  /tmp/rapids-dependency-file-generator
+
+pip uninstall --yes rapids-dependency-file-generator
+pip install /tmp/rapids-dependency-file-generator
+
```
jameslamb, line 13:
```suggestion
```

TODO: remove this before merging

ci/build_wheel_libcuspatial.sh
```diff
@@ -3,4 +3,34 @@
...
+rapids-logger "Installing build requirements"
+python -m pip install \
+    -v \
+    --prefer-binary \
+    -r /tmp/requirements-build.txt
+
+# build with '--no-build-isolation', for better sccache hit rate
+# 0 really means "add --no-build-isolation" (ref: https://github.com/pypa/pip/issues/5735)
+export PIP_NO_BUILD_ISOLATION=0
```
jameslamb, line 34: Ok I think this is working.

On the latest build on `branch-24.12`, you can see log lines about installing build dependencies and creating a temporary build directory.

```text
Created temporary directory: /tmp/pip-wheel-ndbc3klu
Created temporary directory: /tmp/pip-ephem-wheel-cache-blfbvohw
Looking in indexes: https://pypi.org/simple, https://pypi.anaconda.org/rapidsai-wheels-nightly/simple, https://pypi.nvidia.com/
Processing /__w/cuspatial/cuspatial/python/libcuspatial
  Added file:///__w/cuspatial/cuspatial/python/libcuspatial to build tracker '/tmp/pip-build-tracker-vn6jabrn'
  Created temporary directory: /tmp/pip-build-env-64dc7d88
  Running command pip subprocess to install build dependencies
  Installing build dependencies: started
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/11378013030/job/31653081303#step:9:19))

On this PR, none of that.

```text
Looking in indexes: https://pypi.org/simple, https://pypi.anaconda.org/rapidsai-wheels-nightly/simple, https://pypi.nvidia.com/
Processing /__w/cuspatial/cuspatial/python/libcuspatial
  Running command Preparing metadata (pyproject.toml)
  Preparing metadata (pyproject.toml): started
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/11390386364/job/31691951451?pr=1473#step:9:133))

jameslamb: I mention that because it looks to me like we're not getting `sccache` hits.

```text
Successfully built libcuspatial-cu11
Compile requests                      28
Compile requests executed             28
Cache hits                             4
Cache hits (c [gcc])                   2
```

There hasn't been a new `rmm` commit merged since yesterday, and the last new `cudf` commit merged was an hour ago.

Gonna try another build to double-check this.

jameslamb: Yeah cool ok, I just did another build and saw all cache hits. Must have just gotten unlucky before with new `cudf` changes being committed and new nightlies appearing.

```text
Compile requests                     28
Compile requests executed            28
Cache hits                           28
Cache hits (c [gcc])                  2
Cache hits (c++ [gcc])                2
Cache hits (cuda [nvcc])             24
```

([build link](https://github.com/rapidsai/cuspatial/actions/runs/11391151277/job/31694320694?pr=1473))

ci/build_wheel_libcuspatial.sh
jameslamb: removed this now that https://github.com/rapidsai/dependency-file-generator/pull/115 has been merged, `rapids-dependency-file-generator` 1.16.0 has been released, and all new CI images have been built (https://github.com/rapidsai/ci-imgs/actions/runs/11461526811)

.pre-commit-config.yaml
```diff
@@ -48,7 +48,7 @@ repos:
             - id: verify-copyright
             - id: verify-alpha-spec
       - repo: https://github.com/rapidsai/dependency-file-generator
-        rev: v1.13.11
+        rev: v1.16.0
```
jameslamb, line 51: Updating the `rapids-dependency-file-generator` pre-commit hook version isn't necessary for this change, but since I just made that release (https://github.com/rapidsai/dependency-file-generator/pull/115), thought it'd be helpful to also do that across these PRs fiddling with build isolation. So that if that new release broke something, we'll find out sooner.

bdice (approved):
Changes look good to me. Just to verify, nothing is needed for `cuproj` since it doesn't have C++ wheels, right?

ci/build_wheel_libcuspatial.sh
vyasr: Awesome thanks for being careful with the checking.

ci/build_wheel.sh
vyasr: I opened https://github.com/rapidsai/build-planning/issues/111.

vyasr (approved):
This looks great! Thanks for iterating a few times James, I think we've come up with a clean end state that we can port everywhere now.

vyasr:
And yes that's right Bradley, no C++ cuproj. It's built as part of the cuspatial package.

ci/build_cpp.sh
```diff
@@ -15,7 +15,11 @@ rapids-print-env
 
 rapids-logger "Begin cpp build"
 
+sccache --zero-stats
+
 RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry mambabuild \
     conda/recipes/libcuspatial
 
+sccache --show-adv-stats
```
jameslamb, line 23: Added `sccache` stats printing for conda builds as well based on https://github.com/rapidsai/build-planning/issues/111 and offline conversation with @vyasr .

### Why not `build.sh`?

Here, I'm proposing that these be put in `ci/build_{cpp,python}.sh`, wrapped around any `conda mambabuild` invocations that will end up compiling anything (so, for example, I included them in `build_python.sh` because `cuspatial` and `cuproj` both have some Cython extensions).

I think this is preferable to `build.sh`, because the `build_{cpp,python}.sh` scripts are unambiguously for CI, and can unconditionally invoke `sccache`. `build.sh` is invoked in CI, but it's also intended for local use, and I didn't want to assume the presence of `sccache` for all local builds are introduce the complexity of having to check for its use.

There are also some projects, like `cuvs`, that have their own handling of printing `sccache` stats and creating build summary reports based on them in `build.sh`, and I didn't want to interfere with that.

example:

https://github.com/rapidsai/cuvs/blob/e7f1085b71c340b9600f5f38f7f0059a5c7aa806/build.sh#L364-L366

https://github.com/rapidsai/cuvs/blob/e7f1085b71c340b9600f5f38f7f0059a5c7aa806/build.sh#L408-L424

PR #1475, opened 2024-10-14
Updated libcudftestutil CMake linking logic for 24.12

lamarrr:
## Description
This merge request fixes the CMake linking logic to cudftestutil for branch-24.12, allowing you to specify which version of GTest/GBench to use as long as the API is source compatible.

Follows up on: https://github.com/rapidsai/cudf/pull/16839

## Checklist


cpp/tests/CMakeLists.txt
```diff
@@ -17,6 +17,25 @@
 ###################################################################################################
 # - compiler function -----------------------------------------------------------------------------
 
+add_library(cuspatial_test_common OBJECT test_common.cpp)
```
harrism, line 20: question: Why do we need a library with an empty source file that doesn't include anything?

harrism (commented):
Just a question.

cpp/tests/CMakeLists.txt
lamarrr: Two-fold:
- cmake doesn't allow adding an empty list of sources for a library.
- cudftestutil has been split into cudftestutil and cudftestutil_impl (cmake's INTERFACE sources), and they need to be linked to the executable. if we link to cudftestutil_impl for each test executable, we'll be paying the compilation cost repeatedly for each test compilation (due to source injection). instead, we create an object library called test_common and link to that, paying the compilation cost once.

harrism: OK, thanks. Can you please document this in a comment?

harrism (approved):
I'm approving so we can get this merged ASAP, but please document the reason for the empty source file / library.

cpp/tests/CMakeLists.txt
lamarrr: yeah, sure!

bdice (approved):
Nice work! This all looks good to me. Let's merge when CI passes, to unblock other PRs.

bdice:
I'm going to queue a merge for this, which will trigger once tests pass.

cpp/tests/CMakeLists.txt
harrism: Thanks. Late question: why can't this empty library object just be created at the cudf level, rather than in cuspatial?

lamarrr: cudftestutil depends on gtest and gmock. cudf doesn't intend to ship a gtest binary artifact any longer and we want users to have the freedom to choose whichever gtest versions they intend to use as long as it's source-compatible.

PR #1483, opened 2024-11-13
prefer wheel-provided libcuspatial.so in load_library(), use RTLD_LOCAL

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/118

Modifies `libcuspatial.load_library()` in the following ways:

* prefer wheel-provided `libcuspatial.so` to system installation
* expose environment variable `RAPIDS_LIBCUSPATIAL_PREFER_SYSTEM_LIBRARY` for switching that preference
* load `libcuspatial.so` with `RTLD_LOCAL`, to prevent adding symbols to the global namespace ([dlopen docs](https://linux.die.net/man/3/dlopen))

### How I tested this

See "How I tested this" in https://github.com/rapidsai/cudf/pull/17316

Also opened this PR originally pulling in built packages from https://github.com/rapidsai/cudf/pull/17316

### Checklist


python/libcuspatial/libcuspatial/load.py
```diff
@@ -16,8 +16,38 @@
 import ctypes
 import os
 
+# RTLD_LOCAL is here for safety... using it loads symbols into the
+# library-specific table maintained by the loader, but not into the
+# global namespace where they may conflict with symbols from other
+# loaded DSOs.
```
vyasr, line 22:
```suggestion
# Loading with RTLD_LOCAL adds the library itself to the loader's
# loaded library cache without loading any symbols into the global
# namespace. This allows libraries that express a dependency on
# this library to be loaded later and successfully satisfy this dependency
# without polluting the global symbol table with symbols from
# libcuspatial that could conflict with symbols from other DSOs.
```

python/libcuspatial/libcuspatial/load.py
```diff
@@ -33,32 +63,30 @@ def load_library():
...
+            libcuspatial_lib = _load_system_installation(soname)
+        except OSError:
+            libcuspatial_lib = _load_wheel_installation(soname)
+    else:
+        # Prefer the libraries bundled in this package. If they aren't found
+        # (which might be the case in builds where the library was prebuilt
+        # before packaging the wheel), look for a system installation.
+        libcuspatial_lib = _load_wheel_installation(soname)
+        if libcuspatial_lib is None:
+            libcuspatial_lib = _load_system_installation(soname)
```
vyasr, line 87: I'm on the fence about whether we should have the fallback in this case, or if we should require the environment variable to be set to get the system library and explicitly raise if the local load fails for some reason. Loading the system library is otherwise somewhat meaningless here because if it is already on the loader's search path then the subsequent DSOs containing a DT_NEEDED entry on it will automatically find it, so the end result is the same. WDYT?

vyasr (commented):
The general logic looks right. Rewrote the explanatory comment to hopefully clarify what we're doing, and left one comment on resolution order.

python/libcuspatial/libcuspatial/load.py
jameslamb: Ok sure, works for me.

python/libcuspatial/libcuspatial/load.py
jameslamb: I think we *should* have the fallback here, because then you can say with higher confidence that after a call to `libcuspatial.load_library()` returns, `libcuspatial.so` has been loaded.

Because that "subsequent DSOs containing a DT_NEEDED entry" loading order might be different in different variants (or change as this library's code changes), having the fallback right here in `libcuspatial.load_library()` should generally reduce the variability of library-loading orders we might see out in real-world use of these libraries, which should reduce the scope for bugs related to the order that things are loaded. 

Maybe this isn't *that* important for `libcuspatial`, but I think it's a better general pattern, and would be more important for things further upstream.

Inspired by this:

https://github.com/rapidsai/cuspatial/blob/724d170a2105441a3533b5eaf9ee82ddcfc49be0/python/libcuspatial/libcuspatial/load.py#L21-L25

And scared by this: https://github.com/microsoft/LightGBM/issues/6595#issuecomment-2273745921

Unless you're thinking that the use of `RTLD_LOCAL` now makes the exact loading order less important?

python/libcuspatial/libcuspatial/load.py
```diff
@@ -16,8 +16,40 @@
...
+    return ctypes.CDLL(soname, PREFERRED_LOAD_FLAG)
+
+
+def _load_wheel_installation(soname: str):
+    """Try to dlopen() the library indicated by ``soname``
+
+    Returns ``None`` if the library cannot be loaded.
+    """
+    out = None
+    for lib_dir in ("lib", "lib64"):
```
vyasr, line 42: Can we double-check whether this loop is actually necessary? I think I originally added it because of a case where some libraries in a package were installed to lib while others were lib64 (specifically, I believe I did this for nvcomp libraries in cudf/kvikio). Our wheels should be consistent in this placement though AFAIK.

python/libcuspatial/libcuspatial/load.py
vyasr: You are correct that there is an order-dependence here, and if some other package also chose to bundle a copy of libcuspatial.so (using it as an example, of course something like OpenMP is far more likely to have this problem) then we'd be opening ourselves up to choosing their copy. Conversely, doing it this way could _break_ such usages. Ultimately if packages are not in agreement about the version of a library they are using, they either each need to vendor a fully disambiguated copy (a la auditwheel) or they accept some risk of symbol collisions.

That being said, I guess my question would be by your logic shouldn't we fail if we didn't load the library successfully?

jameslamb: Sorry, it was maybe distracting and unhelpful to post that LightGBM issue that talked about multiple *copies*, when really I was just trying to focus on the importance of loading *order* even if only 1 copy will ultimately be found.

In my mind, we should choose between these options in this part of the code:

* fall back to loading the system-installed library
* do nothing (just successfully return `None` from `load_library()`, and rely on the loader getting it later when it loads some other DSO that needs it)

I don't think we should raise an exception here.

That would mean these `RAPIDS_{lib}_PREFER_SYSTEM_LIBRARY` environment variables would have to be threaded through devcontainers and DLFW builds (which do a build + system install of the shared libraries and then build wheels that expect to find those).

I don't think that maintenance effort (now + once per new C++ wheel we add in the future) is worth it.

RAPIDS wheel-testing CI involves setting up a fresh container and installing wheels into it, so no risk of accidentally picking up a system-installed `libcuspatial.so` ... and therefore I think we can rely on CI to catch any issues with the wheels finding/loading the library.

python/libcuspatial/libcuspatial/load.py
jameslamb: Sure! Fastest way I could think of to test this was to just use CI, so pushed a commit: https://github.com/rapidsai/cuspatial/pull/1483/commits/b484ffde00580cb6540b6eaefcb3b3dd14612124

This path is only searched within the wheels itself, and we aren't supporting sdists, so I think we can be confident that if CI works, both paths aren't necessary.

python/libcuspatial/libcuspatial/load.py
vyasr: > I don't think we should raise an exception here.
>
> That would mean these RAPIDS_{lib}_PREFER_SYSTEM_LIBRARY environment variables would have to be threaded through devcontainers and DLFW builds (which do a build + system install of the shared libraries and then build wheels that expect to find those).
> 
> I don't think that maintenance effort (now + once per new C++ wheel we add in the future) is worth it.

I guess that is probably OK logic for cudf. I don't know if it'll fly in the more general case. I'm starting to feel like we're jumping through too many hoops to allow for installation of the "empty" libcudf wheel in devcontainers and DLFW and maybe we need to think through a different solution. Between this kind of hackery, all the dependencies.yaml/RBB settings we had to introduce, and some of the potential bugs, I'm starting to think it might be easier to reconsider the premise and see if we should just find a way to not install these Python packages at all in the context of environments where we don't want them to actually ship a library. It seems wrong to support a case where these wheels try to load their own copy of a library and then fail.

But I suppose that is a digression and thinking too far ahead for 24.12. For now, I agree that we should not error. Between the two options above, I am OK with either. In your previous message, you said

> having the fallback right here in libcuspatial.load_library() should generally reduce the variability of library-loading orders we might see out in real-world use of these libraries, which should reduce the scope for bugs related to the order that things are loaded.

I don't think that is really true in any practical sense since the only time that we ever expect these `load_library` calls to happen is right before importing extension modules (or other C++ wheels) that need this library, at which point the exact same search will happen (just managed directly by the loader instead of by our manual dlopen calls). That being said, the explicit call at least gives us an exact line to drill down to when debugging, so I'll leave the decision to your discretion.

jameslamb: > But I suppose that is a digression and thinking too far ahead for 24.12

I agree with everything you said about the amount of effort we're going through to support those other types of builds being too large. But I also agree... it's too late in 24.12 and there is too much other stuff left to do to change that as part of this.

> I don't think that is really true in any practical sense since the only time that we ever expect these load_library calls to happen is right before importing extension modules (or other C++ wheels)

Very fair, I was being (unjustifiably) cautious here.

>  the explicit call at least gives us an exact line to drill down to when debugging, so I'll leave the decision to your discretion.

Let's keep it in.

This point about having an exact line to point to makes sense to me, and having the explicit call also means we're explicitly controlling the visibility (by setting `RTLD_LOCAL`) instead of deferring that to the loader's configuration, which I think is helpful.

python/libcuspatial/libcuspatial/load.py
jameslamb: Just made these changes in `cudf` as well (https://github.com/rapidsai/cudf/pull/17316#issuecomment-2476823209). Now that the builds on that PR have completed, just re-triggered CI here to test one more time that `libcudf` and `libcuspatial` would play nicely together.

python/libcuspatial/libcuspatial/load.py
vyasr: Yeah let's revisit this next release.

vyasr: Note that this

> This point about having an exact line to point to makes sense to me, and having the explicit call also means we're explicitly controlling the visibility (by setting RTLD_LOCAL) instead of deferring that to the loader's configuration, which I think is helpful.

is unfortunately only sort of true. Local loads can be promoted to global loads by future calls, so control remains limited.

python/libcuspatial/libcuspatial/load.py
vyasr: Great thanks for that. Everything looks green now.

vyasr (approved):

python/libcuspatial/libcuspatial/load.py
jameslamb: > Local loads can be promoted to global loads by future calls

Every time I think I'm understanding this, there's another level of complexity 😅 

Thanks for that note, I'll keep that in mind.

PR #1495, opened 2024-12-03
Remove --expt-relaxed-constexpr NVCC flag

### LINKED ISSUE #1494: [FEA]: Remove --expt-relaxed-constexpr from NVCC options

Improvement

Medium

Projects that use cuProj may prefer not to specify `--expt-relaxed-constexpr`. This option is no longer needed by cuProj and we should probably remove it to prevent future breakages.

After testing, ~neither cuSpatial nor~ cuProj doesn't currently need this flag enabled.  

Note that the dependency on `cudftestutil` injects `--expt-relaxed-constexpr` onto the command line for cuSpatial tests. Once we remove that dependency it will be gone completely.

Keep it. We have had at least one user of cuProj who preferred not to have this command line argument be a requirement. It's not likely, but it is possible for this to cause problems. So I think it's better to just remove it.

### Additional context

See also [_No response_](https://github.com/rapidsai/cudf/issues/7795)

harrism:
## Description
Closes #1494

Removes all (unneeded, currently) instances of `expt-relaxed-constexpr` from cuSpatial and cuProj CMake. 

Also adds linkage of `rmm::rmm_logger_impl` needed after changes to RMM and removal of `cudftestutil` from cuproj benchmarks.

## Checklist


trxcllnt (approved):

isVoid (approved):

cpp/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ set(CUSPATIAL_BUILD_TESTS ${BUILD_TESTS})
 set(CUSPATIAL_BUILD_BENCHMARKS ${BUILD_BENCHMARKS})
 
 set(CUSPATIAL_CXX_FLAGS "")
-set(CUSPATIAL_CUDA_FLAGS --expt-extended-lambda --expt-relaxed-constexpr)
```
isVoid, line 71: Do we also want to remove `--expt-extended-lambda` here? I think it's a good idea to remove exception flags, just to make sure this is intended since it's not mentioned in the PR desc.

isVoid (commented):
Additionally, when I searched this flag in `compile_commands.json`, it still shows up in `cuprojshim` compile commands. Is this a flag inherited? Since I didn't see it in cmakelists.

java/src/main/native/CMakeLists.txt
```diff
@@ -58,8 +58,6 @@ endif()
 set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -gencode=arch=compute_60,code=sm_60")
 set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -gencode=arch=compute_70,code=sm_70 -gencode=arch=compute_70,code=compute_70")
 
-set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-extended-lambda --expt-relaxed-constexpr")
-
```
isVoid, line 62: Same question regarding removal of `--expt-extended-lambda`.

cpp/CMakeLists.txt
harrism: I think it's fine to remove here since it's set in `ConfigureCUDA.cmake`. This was redundant.

java/src/main/native/CMakeLists.txt
harrism: I don't think that this file is used. Also it's sloppy in that it uses `CMAKE_CUDA_FLAGS` unstead of somethign like `CUSPATIAL_JNI_CUDA_FLAGS`.

harrism:
> Additionally, when I searched this flag in `compile_commands.json`, it still shows up in `cuprojshim` compile commands. Is this a flag inherited? Since I didn't see it in cmakelists.

Thanks. I just checked and initially I saw the cuprojshim relaxed-constexpr too. But I deleted the python build directory and rebuilt (to get rid of compile_commands.json) and it regenerated that file without --expt-relaxed-constexpr in it.

PR #1509, opened 2024-12-23
remove versioneer and setup.cfg references in docs, update rapids-dependency-file-generator

jameslamb:
## Description

Removes references to `setup.cfg` and `versioneer` in the contributing docs... `cuspatial` no longer uses either of these.

Also updates `rapids-dependency-file-generator` to its latest version (v1.17.0) and removes an unused YAML anchor in `dependencies.yaml`.

## Checklist

### DIFF
--- a/docs/source/developer_guide/contributing_guide.md
+++ b/docs/source/developer_guide/contributing_guide.md
@@ -72,15 +72,11 @@ Specifically, cuSpatial uses the following tools:
 - [`isort`](https://pycqa.github.io/isort/) ensures imports are sorted consistently.
 
 Linter config data is stored in a number of files. cuSpatial generally uses `pyproject.toml` over
-`setup.cfg` and avoids project-specific files (e.g. `setup.cfg` > `python/cudf/setup.cfg`). However,
-differences between tools and the different packages in the repo result in the following caveats:
+`setup.cfg` and prefers root-level files to project-specific files.
+However, differences between tools and the different packages in the repo result in the following caveats:
 
-- `flake8` has no plans to support `pyproject.toml`, so it must live in `setup.cfg`.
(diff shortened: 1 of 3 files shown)


docs/source/developer_guide/contributing_guide.md
```diff
@@ -72,16 +72,12 @@ Specifically, cuSpatial uses the following tools:
 - [`isort`](https://pycqa.github.io/isort/) ensures imports are sorted consistently.
 
 Linter config data is stored in a number of files. cuSpatial generally uses `pyproject.toml` over
-`setup.cfg` and avoids project-specific files (e.g. `setup.cfg` > `python/cudf/setup.cfg`). However,
-differences between tools and the different packages in the repo result in the following caveats:
+`setup.cfg` and prefers root-level files to project-specific files.
```
bdice, line 75: I would be fine with removing the reference to `setup.cfg` altogether.

Also we need to do a pass through RAPIDS repos and get rid of flake8/black/isort in favor of ruff. There is no technical reason why we are still using flake8, we just haven't switched over to ruff in several RAPIDS repos. (The only minor caveat I am aware of is that isort supports Cython files while ruff's isort does not, but we decided not to worry about that.)

bdice (approved):

docs/source/developer_guide/contributing_guide.md
jameslamb: Sure, pushed https://github.com/rapidsai/cuspatial/pull/1509/commits/e50c506cd933439b53e9de9f222408728283a42c removing this last reference to it.

And put up a tracking issue for the `ruff` migration: https://github.com/rapidsai/build-planning/issues/130

PR #1514, opened 2025-01-14
Have Cython methods accept pylibcudf.Column instead of cudf._lib.column.Column

mroeschke:
## Description
Related to https://github.com/rapidsai/cudf/issues/17317, it appears the Cython bindings are not dependent on any methods specifically on `cudf._lib.column.Column`. It appears most routines operate on `column_view` which can be provided by a `pylibcudf.Column`.

This PR essentially makes the routines defined in Cython accept a `pylibcudf.Column` instead and return a `cudf.core.column.Column` object instead (which should help cudf transition away from its `cudf._lib.column.Column`)

## Checklist

### DIFF
--- a/python/cuspatial/cuspatial/_lib/distance.pyx
+++ b/python/cuspatial/cuspatial/_lib/distance.pyx
@@ -1,10 +1,10 @@
-# Copyright (c) 2022-2024, NVIDIA CORPORATION.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION.
 
 from libcpp.memory cimport make_shared, shared_ptr, unique_ptr
 from libcpp.utility cimport move, pair
 
-from cudf._lib.column cimport Column
-from pylibcudf cimport Table as plc_Table
+from cudf.core.column.column import Column
+from pylibcudf cimport Column as plc_Column, Table as plc_Table
 from pylibcudf.libcudf.column.column cimport column
 from pylibcudf.libcudf.column.column_view cimport column_view
 from pylibcudf.libcudf.table.table_view cimport table_view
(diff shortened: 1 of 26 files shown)


python/cuspatial/cuspatial/_lib/distance.pyx
```diff
@@ -26,7 +26,12 @@ from cuspatial._lib.cpp.types cimport collection_type_id, geometry_type_id
 from cuspatial._lib.types cimport collection_type_py_to_c
 
 
-cpdef haversine_distance(Column x1, Column y1, Column x2, Column y2):
+cpdef haversine_distance(
+    plc_Column x1,
```
harrism, line 30: Can these just be imported as Column?

mroeschke: We'll have a namespace shadowing issue because we need `cudf.core.column.Column` (for the output) and `pylibcudf.Column` (to type these inputs). I decided to alias `pylibcudf.Column` to `plc_Column`, but happy to use another alias if you'd like

isVoid: Why are we mixing the use of cudf column and pylibcudf column here?

mroeschke: The primary motivation of this PR is based on removing `cudf._lib` in favor of pylibcudf https://github.com/rapidsai/cudf/issues/17317, so `cudf._lib.column.Column` is going away. The libcudf-like APIs (mainly accessing a `column_view` in cuspatial) will now be available on `pylibcudf.Column`, but to be compatible with cudf Python APIs we still need to return a cudf Python column.

We also have to do this `cudf column -> pylibcudf column -> cudf column` conversion currently in cudf itself before calling a Cythonized function.

isVoid: Yes, in an ideal world I'd like to see if we can implement most of the features with just pylibcudf and not having to do the round trip. I think the strong typing in cython functions without any supports of protocol / duck typing makes that a bit harder to achieve that. However, I wonder if cuspatial has a smaller API surface that can help make directly using pylibcudf a bit easier?

As always, thanks for consistent work on removing cudf internals in cuspatial!

isVoid (approved):

python/cuspatial/cuspatial/_lib/distance.pyx
mroeschke: > However, I wonder if cuspatial has a smaller API surface that can help make directly using pylibcudf a bit easier?

Yes definitely. I could do a follow up PR to move the `pylibcudf column -> cudf column` conversion from the Cython layer to the Python layer. (I could have done that in this PR but there would have been more visual noise)

PR #1515, opened 2025-01-22
Add shellcheck to pre-commit and fix warnings

gforsyth:
## Description

`shellcheck` is a fast, static analysis tool for shell scripts. It's good at
flagging up unused variables, unintentional glob expansions, and other potential
execution and security headaches that arise from the wonders of `bash` (and
other shlangs).                   
                                                                                                              
This PR adds a `pre-commit` hook to run `shellcheck` on all of the `sh-lang`
files in the `ci/` directory, and the changes requested by `shellcheck` to make
the existing files pass the check.                              
                                                                                                              
xref: rapidsai/build-planning#135   

## Checklist

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2023-2024, NVIDIA CORPORATION.
+# Copyright (c) 2023-2025, NVIDIA CORPORATION.
 
 set -euo pipefail
 
@@ -21,9 +21,12 @@ conda activate docs
 
 rapids-print-env
 
-export RAPIDS_VERSION="$(rapids-version)"
-export RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
-export RAPIDS_DOCS_DIR="$(mktemp -d)"
+RAPIDS_VERSION="$(rapids-version)"
+RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
+RAPIDS_DOCS_DIR="$(mktemp -d)"
+export RAPIDS_VERSION
+export RAPIDS_VERSION_MAJOR_MINOR
+export RAPIDS_DOCS_DIR
 
 rapids-logger "Build cuSpatial CPP docs"
 pushd cpp/doxygen
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -22,10 +22,12 @@ conda activate docs
 rapids-print-env
 
 RAPIDS_VERSION="$(rapids-version)"
-RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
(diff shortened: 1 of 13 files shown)


gforsyth:
Hmm, it looks like the taxi zone geojson file has disappeared from the NYC Open Data page, which is causing the NYC notebook to fail.

gforsyth:
Rotted link: https://data.cityofnewyork.us/Transportation/NYC-Taxi-Zones/d3c5-ddgc

mroeschke:
> Hmm, it looks like the taxi zone geojson file has disappeared from the NYC Open Data page, which is causing the NYC notebook to fail.

https://github.com/rapidsai/cuspatial/pull/1517 has been merged so this PR should pass now with an update with the 25.02 branch

ci/build_docs.sh
```diff
@@ -21,9 +21,12 @@ conda activate docs
...
 
-export RAPIDS_VERSION="$(rapids-version)"
-export RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
-export RAPIDS_DOCS_DIR="$(mktemp -d)"
+RAPIDS_VERSION="$(rapids-version)"
+RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
+RAPIDS_DOCS_DIR="$(mktemp -d)"
+export RAPIDS_VERSION
+export RAPIDS_VERSION_MAJOR_MINOR
+export RAPIDS_DOCS_DIR
```
jameslamb, line 29:
```suggestion
RAPIDS_VERSION="$(rapids-version)"
export RAPIDS_VERSION

RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
export RAPIDS_VERSION_MAJOR_MINOR

RAPIDS_DOCS_DIR="$(mktemp -d)"
export RAPIDS_DOCS_DIR
```

Very minor style thing, do what you want with it ... but if we're going to separate assigning and exporting like this, I think I'd prefer to do them one variable at a time, instead of having so much distance between the assignment and export calls for each variable. I find that form a bit easier to read.

ci/build_python.sh
```diff
@@ -13,8 +13,6 @@ export CMAKE_GENERATOR=Ninja
 
 rapids-print-env
 
-package_dir="python"
```
jameslamb, line 16: Yesss glad `shellcheck` is helping us catch stuff like this!

jameslamb (approved):
I'm good with this! Love that this caught a couple of unused variables so we could simplify these scripts a bit.

Left one small cosmetic suggestion, do whatever you want with it. I don't feel that strongly.

ci/build_docs.sh
gforsyth: yeah, I had this originally and then changed it, but I agree that having the assignment and export calls closer together makes sense.

PR #1518, opened 2025-01-28
Build and test with CUDA 12.8.0

bdice:
This PR uses CUDA 12.8.0 to build and test.

xref: https://github.com/rapidsai/build-planning/issues/139


jameslamb (approved):
I don't see anything else that needs to be updated, and CI seems happy. Looks good!

PR #1539, opened 2025-02-11
Change download method for trajectory_clustering.ipynb's data for container testing

taureandyernv:
wget no longer exists in our container.  As per our notebook guidelines, for testing, every possible notebook should be testable.  As we dropped wget from our container, we need to install it here.  This code addition intelligently installs wget if necessary.


bdice (changes requested):
I would recommend one or more of the following alternatives, to avoid mucking up the notebook with `apt` commands (which may not be portable if run on non-Debian-based systems):
- remove this line https://github.com/rapidsai/docker/blob/c0679e4792950f0066ce64c314f47f9c6538278d/Dockerfile#L49 and leave wget installed in our base Docker images
- or list `wget` from conda-forge as a cuspatial notebook dependency, by specifying it here (next to `curl`): https://github.com/rapidsai/cuspatial/blob/2e7f4fdcfd266a48b049977cf181c6f4404668f8/dependencies.yaml#L418
- or just modify the notebook to use `curl`, which should already be installed

taureandyernv:
i'll modify to curl.  Thanks for the suggestion, @bdice.  Thanks for the catch @harrism!

harrism (approved):

bdice (approved):

PR #1545, opened 2025-03-04
Replace `thrust::optional` with `cuda::std::optional`

miscco:
Its deprecated and will be removed in a future CCCL release

### DIFF
--- a/cpp/include/cuspatial/detail/functors.cuh
+++ b/cpp/include/cuspatial/detail/functors.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2025, NVIDIA CORPORATION.
  *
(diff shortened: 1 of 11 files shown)


cpp/include/cuspatial/detail/functors.cuh
```diff
@@ -35,8 +36,8 @@ namespace detail {
  * pair of offsets: (0, 3), (3, 5), (5, 8)
  * number of elements between offsets: 3, 2, 3
  *
- * @tparam OffsetPairIterator Must be iterator type to thrust::pair of indices.
- * @param p Iterator of thrust::pair of indices.
+ * @tparam OffsetPairIterator Must be iterator type to cuda::std::pair of indices.
+ * @param p Iterator of cuda::std::pair of indices.
```
bdice, line 40: This function body is using `thrust::get`. Should it use `cuda::std::get`?

bdice (approved):
Approving, now with more power.

PR #1546, opened 2025-03-04
Replace `thrust::tuple` with `cuda::std::tuple`

miscco:
The former is effectively just an alias and we want to get rid of it


bdice (approved):

bdice (approved):

bdice:
@miscco This appears to be running into compilation issues like:

```
 $PREFIX/include/libcudf/rapids/thrust/iterator/detail/transform_output_iterator.inl(55): error: no operator "=" matches these operands
            operand types are: thrust::detail::tuple_of_iterator_references<double &, double &, double &, double &> = cuda::std::__4::tuple<double, double, double, double>
```

https://github.com/rapidsai/cuspatial/actions/runs/13659872808/job/38188389286?pr=1546#step:9:1957

miscco:
Yeah I was not able to compile this properly locally, this will need to wait until at least 2.8

PR #1550, opened 2025-03-20
Define cuproj before cudf to avoid conflicting dependencies.

bdice:
## Description
@isVoid observed a CMake configuration issue:

```
CMake Error at build/pip/cuda-12.8/release/_deps/proj-src/CMakeLists.txt:403 (add_custom_target):
  add_custom_target cannot create target "uninstall" because another target
  with the same name already exists.  The existing target is a custom target
  created in source directory
  "/home/coder/cuspatial/cpp/build/pip/cuda-12.8/release/_deps/zstd-src/build/cmake/lib".
  See documentation for policy CMP0002 for more details.
```

This is a conflict between the CMake code for `proj` and the CMake code for `zstd` (added as a dependency of `cudf` in https://github.com/rapidsai/cudf/pull/17935).

Both `zstd` here https://github.com/facebook/zstd/blob/eca205fc7849a61ab287492931a04960ac58e031/build/cmake/lib/CMakeLists.txt#L287 and `proj` here https://github.com/OSGeo/PROJ/blob/master/CMakeLists.txt#L492 define the same target named `uninstall`.

The `zstd` dependency comes from `cudf`, and the `proj` dependency comes from `cuproj`. However, `zstd` has `if (NOT TARGET uninstall)` guarding its creation of the target. This PR attempts to work around the problem by defining `cuproj` targets before finding `cudf`.

## Checklist

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -103,6 +103,12 @@ include(cmake/Modules/ConfigureCUDA.cmake)
 
 # add third party dependencies using CPM
 rapids_cpm_init()
+
+# Add cuProj before any dependencies, because it needs to define proj before cudf defines zstd.
+# This avoids a bug where both of those dependencies specify conflicting
+# `uninstall` targets.
+add_subdirectory(cuproj)
+
 # find or add cuDF
 include(cmake/thirdparty/get_cudf.cmake)
 # find or install GoogleTest
@@ -115,9 +121,6 @@ include (cmake/thirdparty/get_ranger.cmake)
 
 ###################################################################################################
 # - library targets -------------------------------------------------------------------------------
-
-# cuProj
-add_subdirectory(cuproj)
 
 add_library(cuspatial
     src/bounding_boxes/linestring_bounding_boxes.cu
--- a/cpp/cuproj/CMakeLists.txt
+++ b/cpp/cuproj/CMakeLists.txt
@@ -91,6 +91,7 @@ include(cmake/modules/ConfigureCUDA.cmake)
 # add third party dependencies using CPM
 rapids_cpm_init()
 
+include(cmake/thirdparty/get_nvtx.cmake)
 include(cmake/thirdparty/get_rmm.cmake)
 
 # find or install GoogleTest and Proj
--- a/cpp/cuproj/cmake/thirdparty/get_nvtx.cmake
+++ b/cpp/cuproj/cmake/thirdparty/get_nvtx.cmake
@@ -1,0 +1,25 @@
+# =============================================================================
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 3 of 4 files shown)


KyleFromNVIDIA (approved):

harrism:
@bdice I wonder if it's also worth filing issues against `proj` and `zstd` asking them to "namespace" their cmake targets? Or perhaps there's a way to make these targets hidden from outside / dependent projects so they don't conflict?

bdice:
That sounds reasonable to me, but @KyleFromNVIDIA might know the best practices here better than I do. Kyle, could you make a proposal for what we could ask upstream libraries to do?

KyleFromNVIDIA:
I've opened https://github.com/rapidsai/rapids-cmake/pull/801 to make it easier for us to patch projects with patches that only affect the build process, which would allow us to patch the upstream libraries without having to contribute anything.

If we do ask them to change anything, I would suggest only creating the `uninstall` target if their top-level `CMakeLists.txt` is equal to `CMAKE_SOURCE_DIR`.

isVoid:
There seems like another issue with cmake:
```
2025-03-20T15:45:28.2843735Z   CMake Error at /home/coder/cuspatial/cpp/build/pip/cuda-12.8/release/_deps/rmm-build/rmm-targets.cmake:52 (set_target_properties):
2025-03-20T15:45:28.2845128Z     The link interface of target "rmm::rmm" contains:
2025-03-20T15:45:28.2845351Z 
2025-03-20T15:45:28.2845642Z 
2025-03-20T15:45:28.2848704Z       nvtx3::nvtx3-cpp
2025-03-20T15:45:28.2848934Z 
2025-03-20T15:45:28.2854262Z     but the target was not found.  Possible reasons include:
2025-03-20T15:45:28.2855817Z 
2025-03-20T15:45:28.2855878Z 
2025-03-20T15:45:28.2856150Z       * There is a typo in the target name.
2025-03-20T15:45:28.2856502Z       * A find_package call is missing for an IMPORTED target.
2025-03-20T15:45:28.2856804Z 
2025-03-20T15:45:28.2856931Z       * An ALIAS target is missing.
2025-03-20T15:45:28.2857103Z 
```

KyleFromNVIDIA:
Where did you see that error? Could you please post the full stack trace (or a link to it)?

isVoid:
@KyleFromNVIDIA Sure, this is from the devcontainer test with pip install:
https://github.com/rapidsai/cuspatial/actions/runs/13971130617/job/39112956381?pr=1550#step:9:8059

KyleFromNVIDIA:
We might need to add our own `get_nvtx.cmake`, the way cudf has: https://github.com/rapidsai/cudf/blob/branch-25.06/cpp/cmake/thirdparty/get_nvtx.cmake

isVoid:
The problem still persists. This error was reported from rmm, do we see any error outside of cuspatial?

KyleFromNVIDIA:
It's an issue with rapids-cmake. I've opened https://github.com/rapidsai/rapids-cmake/pull/804 to address it.

KyleFromNVIDIA:
I've decided to instead deal with this by patching the appropriate projects. I've opened https://github.com/rapidsai/rapids-cmake/pull/805 to backport the build patch functionality and will then patch the projects.

KyleFromNVIDIA:
I'm testing the patch approach at https://github.com/rapidsai/cuspatial/pull/1552. If it works, we can think about backporting it to 25.04.

KyleFromNVIDIA:
Closing this in favor of #1552.

PR #1558, opened 2025-04-08
fix(deps): pin shapely<2.1.0

gforsyth:
## Description

Shapely 2.1.0 is a major release and it broke `geopandas` usage of `shapely`. Pinning to <2.1.0

## Checklist


dependencies.yaml
```diff
@@ -456,6 +456,8 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - &geopandas geopandas>=1.0.0
+          # TODO: remove this transitive dependency pin once the breakages are resolved
```
jameslamb, line 459: Are there any GitHub issues tracking this, either here in `cuspatial` or in `geopandas` / `shapely`?

I think it'd be helpful to link those in this comment, or otherwise in the PR description (which people could find by working backwards from the git blame).

If there aren't any, could you please write one up here in `cuspatial` and link it in this code comment?

python/cuspatial/pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "libcuspatial==25.4.*,>=0.0.0a0",
     "numpy>=1.23,<3.0a0",
     "rmm==25.4.*,>=0.0.0a0",
+    "shapely<2.1.0",
```
jameslamb, line 39: I think the `conda` recipe should pick up a `run_constrained` entry with this same pin, to achieve similar behavior.

Otherwise, users installing the `cuspatial` conda package will get the newer versions of `shapely` and face the issues this PR aims to fix.

You're not seeing that show up in CI because of the `shapely` pin being added in `dependencies.yaml`.

jameslamb (approved):
Good to see you have #1559 up to fix the `pip` devcontainer, I'm glad we don't need to put a ceiling on `CMake` here (as that would affect the unified devcontainers in https://github.com/rapidsai/devcontainers).

I left some other comments. Marking this "approve" so you don't have to wait for a new review from me... I saw this is targeting `branch-25.04` and that release is already in progress, so I'm guessing this is highly time-sensitive. But please do consider those suggestions.

dependencies.yaml
gforsyth: good call, @jameslamb !

Link here: https://github.com/rapidsai/cuspatial/issues/1561
and also inlined in `dependencies.yaml` and the `meta.yaml`

python/cuspatial/pyproject.toml
gforsyth: Good catch, added a `run_constrained` along with a link to #1561

gforsyth:
Thanks @jameslamb -- I'm going to close this out for now, as `geopandas` released version 1.0.1 which fixes the regression with shapely 2.1.0

jameslamb:
Oh hey great! That's an even better solution haha

PR #1567, opened 2025-05-12
CUDA 12.9 use updated compression flags

robertmaynard:
Set more fatbin compression flags to maintain existing binary sizes

### DIFF
--- a/cpp/cmake/Modules/ConfigureCUDA.cmake
+++ b/cpp/cmake/Modules/ConfigureCUDA.cmake
@@ -32,7 +32,10 @@ list(APPEND CUSPATIAL_CUDA_FLAGS -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-
 
 # Produce smallest binary size
 list(APPEND CUSPATIAL_CUDA_FLAGS -Xfatbin=-compress-all)
-if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA" AND CMAKE_CUDA_COMPILER_VERSION VERSION_EQUAL 12.9)
+if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA"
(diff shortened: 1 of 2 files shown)


cpp/cmake/Modules/ConfigureCUDA.cmake
```diff
@@ -32,6 +32,9 @@ list(APPEND CUSPATIAL_CUDA_FLAGS -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-
 
 # Produce smallest binary size
 list(APPEND CUSPATIAL_CUDA_FLAGS -Xfatbin=-compress-all)
+if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA" AND CMAKE_CUDA_COMPILER_VERSION VERSION_EQUAL 12.9)
+  list(APPEND CUSPATIAL_CUDA_FLAGS -Xfatbin=--compress-level=3)
+endif()
```
karthikeyann: Should it be `VERSION_EQUAL` or `VERSION_GREATER_EQUAL`?

jakirkham:
Thanks Rob! 🙏

Given we are dropping cuSpatial in RAPIDS 25.06 ( [RSN 45]( https://docs.rapids.ai/notices/rsn0045/ ) ), think we need not pursue this further

cpp/cmake/Modules/ConfigureCUDA.cmake
robertmaynard: This is only required for 12.9, so that is why we are using `VERSION_EQUAL`

harrism (approved):

harrism:
Until I get cuProj moved out into its new repo, I'm happy to get improvements like this in so I can move them over.

bdice (approved):

harrism:
Why isn't CI running?

PR #1570, opened 2025-05-29
Provide script input where required

KyleFromNVIDIA:
## Description
Due to https://github.com/rapidsai/shared-workflows/pull/362.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1569, opened 2025-05-29
Remove CUDA 11 from dependencies.yaml

KyleFromNVIDIA:
## Description
Issue: https://github.com/rapidsai/build-planning/issues/184

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/cuspatial/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1571, opened 2025-06-25
Remove nvidia and dask channels

vyasr:
Now that we have dropped support for CUDA 11 we no longer require the nvidia channel.
With the changes in https://github.com/rapidsai/rapids-dask-dependency/pull/85, RAPIDS now only uses released versions of dask, so we no longer need the dask channel either.
Contributes to https://github.com/rapidsai/build-planning/issues/184

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,7 @@ repos:
             - id: verify-codeowners
               args: [--fix, --project-prefix=cuspatial]
       - repo: https://github.com/rapidsai/dependency-file-generator
-        rev: v1.17.0
+        rev: v1.19.0
         hooks:
             - id: rapids-dependency-file-generator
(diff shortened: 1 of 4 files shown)


jameslamb (changes requested):
Development on this repo is frozen, I don't think we need this PR.

CI isn't running because this repo no longer has authorization to use our runners. Even if it did, it's unlikely that CI would pass... it's been skipped for many other build/packaging updates recently.

ref:

* https://github.com/rapidsai/cuspatial/pull/1570#issuecomment-2920583684
* https://github.com/rapidsai/cuspatial/pull/1569#issuecomment-2920584538

vyasr:
Sorry, forgot to remove cuspatial in the second edit of my rapids-reviser changes (I started without it but copy-pasta happened again after).

